PR #31, opened 2016-01-19
[threading] Don't use cout - not thread-intelligent

bjmoor:
StopAction should not use cout (in fact, nothing should), as threading with cout is insane at best.

Should help address sst-elements/issue78


nmhamster:
Merged.

PR #33, opened 2016-01-25
Fix mempool reference outside of ifdefs

ericvh:
Currently breaks when you configure with --disable-mem-pools

Signed-off-by: Eric Van Hensbergen <ericvh@gmail.com>


nmhamster:
Eric, are you using the devel branch or master for your builds? This pull request is against master, unfortunately we have to stage the merges into devel where we have our testing configured first. Can you let me know?

ericvh:
It is actually against devel, I'll try to fix -- otherwise I'll close the request and submit another against devel.

PR #34, opened 2016-01-26
Fix build with --disable-mem-pools

ericvh:
This adds an ifdef conditionalize some mempool reservation code when we've disabled mempools with configure.  Without this patch, a build with --disable-mem-pools fails.

Signed-off-by: Eric Van Hensbergen <ericvh@gmail.com>


nmhamster:
@ericvh - thanks for the patch!

PR #36, opened 2016-02-04
fix compilation on gcc 5.2.1

cdunham:
The ConfigComponent constructor is listed under the wrong access specifier:
    ../../sst/core/configGraph.h:162:5: error: ‘SST::ConfigComponent::ConfigComponent()’ is private

### DIFF
--- a/configGraph.h
+++ b/configGraph.h
@@ -145,6 +145,7 @@ public:
     ConfigComponent cloneWithoutLinks() const;
     ConfigComponent cloneWithoutLinksOrParams() const;
     
+    ConfigComponent() {}
     ~ConfigComponent() {}
 private:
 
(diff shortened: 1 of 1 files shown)


bjmoor:
Is this actually a GCC 5.2.1 issue, or is it a Boost version > 1.56 issue?  I just wish to be clear on this.  I have seen the same issue, but it has been around Boost Serialization problems, not compiler versions.

cdunham:
Brandon,
Ahh, Okay.  You are probably right; I used Boost 1.58 in this case.
Apologies.


Curtis

nmhamster:
Do we have a decision on this and what we need to do longer term anyone?

nmhamster:
I'm closing the pull request for now because we want to be addressing Boost correctly into SST v6.0.

nmhamster:
Closed while we review options on Boost.

PR #56, opened 2016-04-12
Param cleanup

feldergast:
Added new code for templated find functions in Params. Removed some of the old map-like functions from Params; will remove the rest once none of the elements reference them.


nmhamster:
FYI - @allevin says that the Jenkins system is undergoing maintenance at the moment and so the build attempts are likely to time out. This does not indicate problems with the code.

PR #57, opened 2016-04-14
Now calls destructors for all components.

feldergast:
When simulation shuts down, it will now delete all the components in the simulation.


allevin:
Closed per Scotts request

PR #82, opened 2016-05-17
Adds Verbose Print out for Library Path Resolution

nmhamster:
Adds Verbose Print out for Library Path Resolution


bjmoor:
In general, we're trying to avoid the use of `cout` and `cerr`, as they are highly non-threadsafe.   I'd recommend switching to either an Output instance, or `fprintf()` as appropriate.

PR #89, opened 2016-05-20
Initial Fix for Environment Loading Issues with SST Core

nmhamster:
This fix introduces an agreed ordering for loading configuration:

(1) Ordered file list provided by the core (can come from command line or Python once wired in, or any other sources.
(2) Environment variable: `SST_CONFIG_FILE_PATH`
(3) `$HOME/.sst/sstsimulator.conf`
(4) `$SST_INSTALL_PREFIX/etc/sst/sstsimulator.conf`

This still needs additional work. The `sat-register` tool now by default attempts to update the global configuration first, if this fails it will then attempt to update user home stored configuration.


nmhamster:
Environment variable `SST_CONFIG_FILE_PATH` is a *list* of configuration files which is separated by default with `:` but this can be overridden with environment variable `SST_CONFIG_FILE_PATH_SEPARATOR`. Default is to tokenize and load each file in turn (in order specified - i.e. right most takes precedence over left).

nmhamster:
Close to check that we have correct merge status.

PR #128, opened 2016-07-07
Remove extra (unguarded) #include of mpi.h

bjmoor:
Don't want/need extra #include of MPI, especially when MPI is disabled.


allevin:
Will recreate this pull request after build issues are resolved

PR #129, opened 2016-07-10
Retry on EEXIST errors in IPC setup

bjmoor:
When IPC startup file clashes, retry with a different filename.


nmhamster:
Looks good :-)

PR #131, opened 2016-07-11
Call component destructors when owning simulation is destroyed

### LINKED ISSUE #126: SST Core does not call Component Destructors
The SST core is not calling component destructors. @feldergast may have added a fix already so we need to check with him before further work on this bug.

bjmoor:
Attempt to fix #126.  Delete components when Simulation is destroyed.  (Thus, calling destructors of components.)


nmhamster:
Addresses bug #126.

Issue #130, opened 2016-07-11
configure --with-zoltan is successful even if zoltan is not found


allevin:
When configure is run against the sst-core, if "--with-zoltan" is used, but Zoltan cannot be found the configure will still succeed.  The report at the end does properly reflect that Zoltan is not enabled.  

This may not be a bug, but it may be confusing to the user who expected Zoltan to be part of the core build and did not pay attention to the report at the end of configure.  

Recommend that configure should fail if Zoltan is not found in this situation.

allevin:
Changes in PR #132 did NOT correct the issue.

nmhamster:
@allevin - additional change set in "Fix issue with Zoltan configure where --with-zoltan supplied by is empty string #137". You found an additional error case.

allevin:
Verified that if zoltan not found, configure will fail.

PR #132, opened 2016-07-11
Fixes Zoltan Configure Issue where ./configure wouldn't fail if not found.

nmhamster:
Fixes Zoltan Configure Issue where ./configure wouldn't fail if not found.


nmhamster:
Addresses issue #130 raised by @allevin.

PR #150, opened 2016-10-03
End time fix

jpvandy:
Resolve the tiny reported time difference between single thread and Multi-Thread or Rank shown on memH sdl9-2


jpvandy:
Should be to devel

PR #155, opened 2016-10-31
Fix printing incorrect directory when a directory cannot be opened

bjmoor:
sst-info was printing out the previous directory tested, rather than the current directory tested, when such directory could not be found.


nmhamster:
Passes tests and has been reviewed. Approved.

PR #208, opened 2017-03-30
Barrier tweaks

bjmoor:
* Break apart more barrier re-use.
* Add spin back-off to barrier during wait
* Add memory fence to barrier release.


bjmoor:
With these changes (Mostly, the backoff), performance (cycles/second) under over-constrained resources, (aka, 2 rank, 2 thread, on a single-core system) is about 10x improved.

allevin:
NOTE: PR autotester or Github failed to set the test and inspection status.  However the comments above indicate that the status should have been set.  I will investigate via the logs and submit an issue with GitHub.  Due to the critical nature of this PR, I used admin privileges to manually merge the PR since it was Inspected by Si and passed all testing.

PR #218, opened 2017-04-10
Add coordinates to Python API

bjmoor:
Allow Python API to specify coordinates for Components.  HDF5 StatOutput will record these.

### DIFF
--- a/src/sst/core/baseComponent.h
+++ b/src/sst/core/baseComponent.h
@@ -263,10 +263,8 @@ public:
     SubComponent* loadNamedSubComponent(std::string name, Params& params);
 
     /** Retrieve the X,Y,Z coordinates of this component */
-    void getCoordinates(double *x, double *y, double *z) const {
-        if ( x != NULL ) *x = my_info->coordinates[0];
-        if ( y != NULL ) *y = my_info->coordinates[1];
-        if ( z != NULL ) *z = my_info->coordinates[2];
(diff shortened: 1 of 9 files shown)


bjmoor:
This code expects a 1-, 2-, or 3-D set of coordinates.   I'm not opposed to a higher-dimensional set of coordinates, however, do visualization engines typically allow for that?   I'm really only familiar with up to 3D visualizations.  I understand that higher-order networks have logically higher numbers of dimensions, but in meat-space, they still are no more than 3 dimensional.

nmhamster:
@bjmoor would it be bad to implement today as an array of coordinates so we have some flexibility on 2D vs. 3D vs. higher dimensionality stuff in the future, but just use 3D for viz right now?

bjmoor:
I'd actually prefer to rename "Coordinates" to "VizCoords" or something like that, rather than to make it an unbounded array.  The rationale for adding this is explicitly for visualization.   Could a concept of generic coordinates be useful elsewhere?  Probably.   However, I feel that there is the potential for significant semantic confusion if we have an un-bounded array "coordinates", and other times, we implicitly declare that `coordinates[0]` is an X-dimension, `coordinates[1]` is a Y-dimension, etc.

bjmoor:
I went ahead and made coordinates a vector, of a minimum length of 3.  If you supply less than 3 elements, it will populate out to the 3rd index with zeros.   @nmhamster , please review again.

Issue #213, opened 2017-04-05
Early dection of mispelled port name does not work with 2x2 mulit- on El Capitan


jpvandy:
A long time ago processing was insert to induce an early seg-fault if a port name was misspelled.   (This was protection against this error being encounter later in a less obvious way.)   Also a test was created to verify the non-matching port was thusly detected.
Currently a bad port name does not induce a seg-fault on 2x2 multi on El Capitan.
It is not clear when a code update removed this failure, because until quite recently 2x2 on all platforms had so test failures that this low priority error wasn't noticed.

bjmoor:
I do not recall ever forcing a segfault in the code.    The ConfigGraph's `checkForStructuralErrors` should be printing a warning about invalid ports.  I would be OK upgrading that warning to a `Output.fatal()` call.

Thoughts?  ( @nmhamster @feldergast )

nmhamster:
@bjmoor definitely would prefer an `Output.fatal()` call because that should handle cleanup correctly with respect to MPI etc.

PR #219, opened 2017-04-11
Make it an error to use an invalid port name.

bjmoor:
Addresses #213


bjmoor:
It appears that there is a BadPort test, which is looking for a SIGSEGV when a bad port is used.  That test needs to be updated to allow this change to pass tests cleanly.  @jpvandy @allevin

PR #235, opened 2017-04-21
Added SubComponentSlots fully to new ELI.

feldergast:


feldergast:
@bjmoor If we don't mark the old ELI deprecated in the 7.0 release, we can't remove it until 9.0, since we guarantee an entire major release before removing deprecated APIs.  That's too long to keep the old ELI around.  We either need to ship with the deprecated warnings in the elements or move all the elements to the new ELI before the release.  Not sure what the plan is at this point.

Issue #145, opened 2016-07-18
SST "--output-directory" option not working?


shavvn:
Earlier I submitted a pull request ( #144 ) regarding the duplicated "--output-directory" option in _config.cc_. After fixing it I found the "--output_directory" still doesn't work as it described: 

> "Controls where SST will place output files debug output and simulation, default is for SST to create a unique directory."  

E.g. the stats csv file would still be dumped where it was after I specified this option to another location.

Then I looked at the code in _simulation.cc_ and found that the variable _output_directory_ in line 157 was never set or referenced.

I know I can still control where the stats file go by setting parameters to _setStatisticOutputOptions_. But it would still be nice if one could specify it through command line.

nmhamster:
Fixed in issue #267

PR #267, opened 2017-06-12
Add Support for Directory Output in SST

nmhamster:
Can put output files into a specific directory. Useful when running large numbers of experiments. See Issue: #145.


nmhamster:
@bjmoor - can you review please?

Issue #182, opened 2017-02-20
sst-config does not set core prefix when SST_CONFIG_FILE_PATH is set


bjmoor:
Copied from sst-element#473

@mjleven Commented on  on Oct 19, 2016
> If SST_CONFIG_FILE_PATH is set before sst-config is run it does not set the core's location prefix.

nmhamster:
Fixed in issue #272

PR #272, opened 2017-06-23
Fixes #270 where Configuration Prefix is set to NONE and Configuration Cannot be Read

### LINKED ISSUE #270: sst-config broken when installed to default `$prefix`
If a user does a clean install of SST with no `--prefix` on the `./configure` line, SST installs into `/usr/local/` as expected.   However, `sst-config` complains that:
```SST: Unableto open configuration file 'NONE/etc/sst/sstsimulator.conf'```

Also, when then attempting to build `sst-elements`, `./configure` succeeds, but at build time, errors are thrown due to an inability to find `sst_config.h`.  The include paths are:
```-I. -I../../../../src -I/include/sst/core -I/include```

Seen on a LOFT install of OSX 10.11.6.

nmhamster:
Reported in bug #270 and #182 if `--prefix` is not set for configuration then `sst-config` is unable to find any root for configuration information and errors. This detects the "NONE" setting (auto-configuration records the parameter is unset) and overrides this to `/usr/local`. Tested on Mac OS X.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,6 +6,7 @@ AC_PREREQ([2.59])
 
 AC_CONFIG_AUX_DIR([config])
 AC_CONFIG_MACRO_DIR([config])
+AC_PREFIX_DEFAULT([/usr/local])
 AM_INIT_AUTOMAKE([1.9.6 foreign dist-bzip2 subdir-objects no-define tar-pax])
 
 # If Automake supports silent rules, enable them.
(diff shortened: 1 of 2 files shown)


nmhamster:
@bjmoor - well, I tried this and I've uploaded the change I made that should have forced it. Even when we explicitly set the prefix we still get NONE returned.

Issue #286, opened 2017-07-26
SST-Info Error while Compiling on POWER8 w/ GCC 4.8.5/RHEL7.3


nmhamster:
```
sstinfo.cc: In member function ‘int SST::SSTInfoConfig::parseCmdLine(int, char**)’:
sstinfo.cc:327:20: warning: comparison is always false due to limited range of data type [-Wtype-limits]
         if ( c == -1 )
```

nmhamster:
Fixed in #287

PR #287, opened 2017-07-26
Fix Runtime Error when using getopt_long

nmhamster:
The calls to getopt_long currently assume return type will be `char`. The return type is actually `int` and GCC 4.8.5 on POWER8 complains that the check to `-1` will always be false, causing the compiler to generate code in which SST will not run. This fixes the behavior on POWER8 w/ GCC 4.8.5. This is also a problem on `sst-info`.


nmhamster:
This PR addresses issue #286.

PR #314, opened 2017-09-28
Change to SharedRegion

feldergast:
Changed the SharedRegionManager to always exchange data on an updateState() call, instead of just when it is being finalized.  This will make the shared region data available to other ranks during init, which is required for some of the elements to function properly.

I believe this fix will cause the change set state to be exchanged multiple times, but this is only an efficiency issue and not a correctness issue.  We can revisit with a more efficient fix at a later time.


feldergast:
Closing for now. Not passing tests.

PR #315, opened 2017-10-03
Shared region fix

feldergast:
SharedRegions need to be updated between every cycle of init(), so that the data can be used during init().  One use case is building up a logical to physical mapping which is needed during init() to route init events.


feldergast:
Removed commented lines that were left in accidentally.

PR #320, opened 2017-10-19
Added an iterative complete() stage following the run() stage.

feldergast:
As part of this, added calls sendUntimedData and recvUntimed data.
There will eventually replace the send/recvInitData calls to keep
the names more generic.  As part of this, changed all variables with
init in the name to use untimed instead (except for those which were
referring specifically to the init() stage.

### DIFF
--- a/src/sst/core/interfaces/simpleNetwork.h
+++ b/src/sst/core/interfaces/simpleNetwork.h
@@ -273,23 +273,33 @@ public:
     virtual Request* recvInitData() = 0;
 
     /**
-     * Sends a network request during the init() or complete() phase
+     * Sends a network request during untimed phases (init() and
+     * complete()).
+     * @see SST::Link::sendUntimedData()
      *
      * For now, simply call sendInitData.  Once that call is
      * deprecated and removed, this will become a pure virtual
-     * function.
+     * function.  This means that when classes implement
+     * SimpleNetwork, they will need to overload both sendUntimedData
+     * and sendInitData with identical functionality (or having the
+     * Init version call the Untimed version) until sendInitData is
+     * removed in SST 9.0.
      */
     virtual void sendUntimedData(Request *req) {
         sendInitData(req);
     }
         
     /**
-     * Receive any data during the init() or complete() phase.
-     * @see SST::Link::recvInitData()
+     * Receive any data during untimed phases (init() and complete()).
+     * @see SST::Link::recvUntimedData()
      *
      * For now, simply call recvInitData.  Once that call is
      * deprecated and removed, this will become a pure virtual
-     * function.
+     * function.  This means that when classes implement
+     * SimpleNetwork, they will need to overload both recvUntimedData
+     * and recvInitData with identical functionality (or having the
+     * Init version call the Untimed version) until recvInitData is
+     * removed in SST 9.0.
      */
     virtual Request* recvUntimedData() {
         return recvInitData();
--- a/src/sst/core/interfaces/simpleNetwork.h
+++ b/src/sst/core/interfaces/simpleNetwork.h
@@ -272,6 +272,29 @@ public:
      */
     virtual Request* recvInitData() = 0;
 
+    /**
+     * Sends a network request during the init() or complete() phase
+     *
+     * For now, simply call sendInitData.  Once that call is
+     * deprecated and removed, this will become a pure virtual
+     * function.
+     */
+    virtual void sendUntimedData(Request *req) {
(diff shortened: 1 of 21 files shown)


src/sst/core/interfaces/simpleNetwork.h
```diff
@@ -272,6 +272,29 @@ class SimpleNetwork : public SubComponent {
...
     virtual Request* recvInitData() = 0;
 
+    /**
+     * Sends a network request during the init() or complete() phase
+     *
+     * For now, simply call sendInitData.  Once that call is
+     * deprecated and removed, this will become a pure virtual
+     * function.
+     */
+    virtual void sendUntimedData(Request *req) {
```
bjmoor: Shouldn't this be the other way around?   If `sendInitData()` will be the deprecated route, shouldn't we be moving functionality there, and making the old versions the wrapper?

src/sst/core/link.cc
```diff
@@ -41,10 +42,13 @@ Link::Link(LinkId_t id) :
     if ( uninitQueue == NULL )
 	uninitQueue = new UninitializedQueue("ERROR: Trying to send or recv from link during initialization.  Send and Recv cannot be called before setup.");
     if ( afterInitQueue == NULL )
```
bjmoor: Checking `afterInitQueue`, but then assigning to `afterRunQueue`.  This doesn't feel right.

src/sst/core/main.cc
```diff
@@ -628,7 +631,7 @@ main(int argc, char *argv[])
             delete your_graph;
         }
 
-        if ( *my_ranks.begin() != myRank.rank) cout << "ERROR" << endl;
```
bjmoor: Should this actually cause an fatal(), rather than just printing "ERROR"... with no indication of what kind of error is going on?

src/sst/core/simulation.cc
```diff
@@ -468,10 +456,41 @@ void Simulation::initialize() {
...
+    untimed_phase = 0;
+    // Walk through all the links and call prepareForComplete()
+    for ( auto &i : compInfoMap ) {
+        i->prepareForComplete();
+    }
+
+    syncManager->prepareForComplete();
+
+    bool done = false;
+    completeBarrier.wait();
```
bjmoor: Do we need a barrier before the compInfoMap's `prepareForComplete` as well?

src/sst/core/interfaces/simpleNetwork.h
feldergast: I don't like the way this turned out either, but making the sendInitData the wrapper won't be backward compatible because it is a pure virtual and anything that implements this SubComponent API will be overriding it.  Ideally I would make the new Untimed called pure virtual, but then old code won't compile.  My goal is to allow components using to SimpleNetwork to use the new calls.  However, this will create a problem with implementations of SimpleNetwork in that they will need to either replicate code (implement both init and untimed) or will have to make one call the other.  If you have any ideas how to make this cleaner, I am glad to change how this is implemented.

src/sst/core/link.cc
feldergast: Fixed.

src/sst/core/main.cc
feldergast: Probably.  Looks like I just changed the cout to printf (wasn't sure if Output was initialized here), but didn't take a good look at what it was doing.  I'm not actually sure what error that's checking for.  I'll look at the code a bit deeper and either add a reasonable error message or remove the check if it's not needed.  It has the feel of a temporary debug printout, but it's been in there for so long, I don't remember for sure.

src/sst/core/simulation.cc
feldergast: No, there is a barrier just before complete is called.  Though it could be argued that the barriers between each of the stages in main.cc should be moved into the various calls for the stages.  Any preference here?  Now is a good time to make a change.  Look at main.cc starting at line 319.

src/sst/core/main.cc
feldergast: This was an old debug statement and has been removed entirely.

src/sst/core/simulation.cc
feldergast: While I don't think the barrier was necessary, I added it because all the other phases had a barrier right up front.  I think we're doing extra barriers, but they are in a non-time critical section, so I'm not too concerned.  This also means, that if the barriers between phases in main.cc are removed, things should still function correctly.

PR #344, opened 2018-04-16
Changing ELI change deprecation warning.  Removal of old ELI will be …

feldergast:
…deferred to version 9.0


nmhamster:
@feldergast - ping.

Issue #374, opened 2019-03-04
Update ELI to use new builder/info interface


jjwilke:
Rather than re-implementing factories/info for each new base class, use generic
````
template <class Base, class... Args>
class Builder {
  virtual Base* create(Args...) = 0;
};
````
For info, use
````
template <class Base, class... Args>
class Info {
};
````
where `Args...` is a type list of the information provided, e.g. for Component
````
class Info<Component,ProvidesParams,ProvidesPorts....> {
};
````
For now, `Info` base class should have two functions to support sst-info:
````
class Info<...>
{
 virtual void outputHumanReadable(std::ostream& os) =0;
 virtual void outputXML(XMLNode* xml) = 0;
}
````

jjwilke:
Key feature is:
````    
for (auto loader : ELI::LoadedLibraries::getLoaders()){
  loader();
}
````
which works around dlopen issues by delaying registration.

PR #376, opened 2019-03-04
Eli policies

jjwilke:
Initial port to new ELI setup allowing arbitrary base class types and arbitrary info types. Issue #374.


jjwilke:
This gets a bit more complicated because of dlopen.

dlopen behavior is undefined when symbols are seen in two different static variables are seen in .so files with the same name, which occurs with template static instantiations.

Loading has to be delayed until after all calls to dlopen. This ensures static lookup tables are unique.

PR #384, opened 2019-03-08
Fix missing functional Header to get SST to compile with GCC 8.0

nmhamster:
Required to get SST to compile with GCC 8.0.

### DIFF
--- a/src/sst/core/elibase.h
+++ b/src/sst/core/elibase.h
@@ -14,6 +14,7 @@
 
 #include <sst/core/sst_types.h>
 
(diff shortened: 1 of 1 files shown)


nmhamster:
Looks like failure could be at the network end? Don't see issues in the build itself.

nmhamster:
Am approving the pull request merge to ensure we have a working code base. This is a critical bug fix which prevents compilation on some test and non-test platforms. We may need to improve test coverage to ensure we correctly identify these problems in the future.

Issue #391, opened 2019-04-08
Cleanup compiler warnings for Intel/GCC


jjwilke:
Big ones are:
* meaningless 'const' on return types
* -1 conversions on unsigned types
* dynamic exception specifications

jjwilke:
Need separate sentinel values for the bitmask and the error code.

allevin:
Will be done for SST 10

Issue #389, opened 2019-04-05
Better error messages for new ELI


jjwilke:
The error messages can be a bit cryptic when elements are not found. This could be improved, particularly for empty strings.

Empty strings are very common since the `find<string>(...)` default behavior is not to crash but to return nothing.

PR #392, opened 2019-04-08
Eli cleanup 389

jjwilke:
Includes both issues #389 and #391. No major code changes. Just eliminating compiler warnings and better error checks/printing.


jjwilke:
I guess I would argue the '-1' is not very descriptive. It's not clear what that parameter means in context. If you switch to an integer, I guess practically speaking everything is two's complement so the bitmasking still just works when you pass in '-1'

jjwilke:
Your call, though. I have made my preference known! I can roll back to just using '-1' if you like that better.

jjwilke:
Woah, cancel pull request I see what I did. This was only supposed to be for the bitmask on output statements and I got carried away.

PR #413, opened 2019-05-29
loopback links

lpsmodsim:
Allow local links for loopback purposes

### DIFF
--- a/src/sst/core/linkPair.h
+++ b/src/sst/core/linkPair.h
@@ -26,7 +26,7 @@ public:
     LinkPair() {}
 
     /** Create a new LinkPair with specified ID */
-    LinkPair(LinkId_t id, bool selfLink = false) :
+    LinkPair(LinkId_t id, bool selfLink) :
         left(new Link(id)),
         right(selfLink ? nullptr : new Link(id)),
         my_id(id)
--- a/src/sst/core/linkPair.h
+++ b/src/sst/core/linkPair.h
@@ -28,15 +28,14 @@ public:
     /** Create a new LinkPair with specified ID */
     LinkPair(LinkId_t id, bool selfLink = false) :
         left(new Link(id)),
-        right(new Link(id)),
-        link(new Link(id))
+        right(selfLink ? nullptr : new Link(id)),
+        my_id(id)
     {
-        my_id = id;
-        self = selfLink;
-        
+        if (selfLink) {
+        	right = left;
+        }
         left->pair_link = right;
         right->pair_link = left;
-        link->pair_link = link;
     }
     
     virtual ~LinkPair() {}
@@ -47,20 +46,14 @@ public:
     }
 
     /** Return the Left Link */
-    inline Link* getLeft() {return self ? link : left;}
+    inline Link* getLeft() {return left;}
     /** Return the Right Link */
-    inline Link* getRight() {return self ? link : right;}
+    inline Link* getRight() {return right;}
 
 private:
-
-    bool self;
-    
     Link* left;
     Link* right;
-    Link* link;
-
     LinkId_t my_id;
-
 };
 
 } //namespace SST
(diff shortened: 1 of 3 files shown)


jjwilke:
I have some questions about this:

1) Why is this necessary? Are self links not okay for the use case?
1) Is connecting a link to the same component on both ends currently not working?
1) At some point, the simulation is going to call `getRight()` - and this will be null. I don't see any null checks in the code and worry about allowing links in the pair to be null.

lpsmodsim:
Re @jjwilke:
1. For the use case self links did not seem to work as I was attempting to mimic the path a link needs to take in the merlin hr_router (VC buffers etc.). Self links also didn't seem to be configurable from the python configuration file, and instead needed to be inserted in the component programatically.
2. At least when I initially attempted this, it did not work using the code unmodified.
3. right is assigned to left in the selfLink case, meaning getRight() points to the same link as getLeft(). In the normal case there are 2 links created so there should never be a null pointer reference.

jjwilke:
Gotcha. I see the `left = right` now. I can't read.

Personally, I'd be more comfortable figuring out why it doesn't work to connect a component to itself. It doesn't seem like we should need a boolean for the special case (but maybe).

Is the problem here that we are going in/out on the same *port*, not just the same *component*? That would be my guess as to why this doesn't just already work.

If this is a same *port* issue, I would suggest fitting it into the SelfLink concept that already exists.

@feldergast Is there any distinction left between ports and self ports? The isSelfPort function doesn't seem to be used anymore.

lpsmodsim:
Yes I forgot but it is specifically connecting one port to itself.

feldergast:
There is no real distinction between a self port and a regular port once configureLink is called.  

You are right that this would work with the existing code if you were connected to two different ports in the same component.  We only need to special case things when you are connected to the same port.  There's really no way to fit this into the configureSelfLink if you want the functionality available in the python.  In the case of a single port, configureLink only gets called once, instead of twice (once on each end), so it needs to be set up slightly differently.  

However, this brings up a good point: we should only special case this if we are using the same port in the same component.  In which case, the way we're doing it may not be best.  I'll give some thought to the best way to support this.  I'll hold of on merging this for now while I think this through.  I will add a request for changes so no one else accidentally merges it either.

lpsmodsim:
Just for my understanding: configureLink gets called on each port? meaning a loopback link on a single port needs to have a single link in the linkPair where everything else should have 2 links in the linkPair because both ports call configureLink? I will look through some of the code to try and get a better understanding of how this all works

allevin:
@feldergast is this PR still valid?  This PR is over 6 months old.

nicmcd:
@feldergast, can we enable making SelfLinks from python? As you mentioned above, once confiureLink is called, the component shouldn't need to know if the link is a regular link or a self link. Having to create SelfLinks in C++ breaks the idea of configuration all in python (not to mention cumbersome). Consider a switch component. The switch should function the same regardless of how its ports get wired up. Having a loopback capability from python seems logical and necessary.

The current python link.connect() function takes two parameters. You could either allow for one or two parameters, or create a separate function called link.loopback().

feldergast:
I'm not quite sure how you physically hook a router port to itself.  Are there cables that do that?  Regardless we can enable the functionality to do loopbacks on a single port to itself.  Hooking together two ports from the same component should already work, though I don't think we actually test this case.  I don't know if it's better to have an explicit call for loopback or just detect it when the two sides that are connected are the same port. I'm leaning toward just detecting when both sides of the link are hooked to the same port.  The link does need to be configured slightly differently in that case.  I'll this to my list of things to look at for the next release.

I am going to close this PR and open a new issue.

nicmcd:
Loopback cable: https://community.fs.com/blog/what-is-loopback-cable-and-how-to-use-it.html

PR #416, opened 2019-06-05
Subcomp update

feldergast:
Disambiguated some of the templated APIs for SubComponent loading and added backward compatibility into the new load*SubComponent calls.  The calls will now try to load using the new API, but will fallback to the old, if needed.  Also added a new call to check if it loaded with the new or old API.

### DIFF
--- a/src/sst/core/baseComponent.h
+++ b/src/sst/core/baseComponent.h
@@ -384,9 +384,15 @@ protected:
         //This shouldn't happen since we just put it in, but just in case
         if ( sub_info == NULL ) return NULL;
 
-        auto ret = Factory::getFactory()->Create<T>(type, params, sub_info->id, params, args...);
-
-        return ret;        
+        // Check to see if this can be loaded with new API or if we have to fallback to old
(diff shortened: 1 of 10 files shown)


src/sst/core/baseComponent.h
```diff
@@ -605,22 +687,122 @@ class SubComponentSlotInfo {
...
+
+       For ease in backward compatibility to old API, this call will
+       try to load using new API and will fallback to old if
+       unsuccessful.
+    */
+    template <typename T, class... ARGS>
+    void createAllSparse(std::vector<T*>& vec, uint64_t share_flags, ARGS... args) const {
+        for ( int i = 0; i <= getMaxPopulatedSlotNumber(); ++i ) {
+            T* sub = create<T>(i, share_flags, args...);
+            vec.push_back(sub);
```
gvoskuilen: Should this check for sub == nullptr before push_back?

feldergast: Yes.  Cut too much from the previous code.

Issue #443, opened 2019-10-17
sst-info segfault


gvoskuilen:
sst-info segfaults at exit on Mac only (Linux runs fine).

jjwilke:
Well, that’s what I get for using `unique_ptr` and and trying to make core sanitizer friendly.

Mac does some awesome private memory/segmentation things during `dlopen`. If I call operator new during `dlopen`, these allocations stay local to the .so file. When I `dlclose`, any memory allocated during `dlopen` ceases to be valid. You would think something you allocated with `operator new` would stay permanent – but no. `unique_ptr` deletes the object, the pointer is no longer valid following `dlclose`, and you get the segfault.

Honestly, we’ve been getting lucky prior to 9.0 since we just leaked memory and didn’t try to clean up. For now, I can get devel working by just leaking memory. It will take some non-trivial finessing to clean up memory during `dlclose`, not just `cxa_finalize`.

jjwilke:
Updated: the destructor of a class defined in `libX.so` is getting called, which then gets `dlclose`d. The destructor code is no longer valid, leading to the segfault. This may not have been happening prior to 9.0.

gvoskuilen:
Segfault is fixed.

PR #444, opened 2019-10-18
address segfault after dlclose: Issue #443

jjwilke:


jjwilke:
I would consider this a critical fix. There were a couple issues here.

I actually don't understand yet why the simulator is not also segfaulting. I may have also discovered the issues we were having with Intel `dlopen`.

jjwilke:
This is Step #1 in fixing the issue. For now, we just leak memory and don't clean up. More significant changes will be required to clean things up during dlclose, not just the end of simulation.

PR #448, opened 2019-11-01
Fix SST_EXPORT_CFLAGS and SST_EXPORT_CXXFLAGS regex

stkaplan:
The \s character class does not work in all versions of sed (even
with -E). Use [:space:] instead.


jjwilke:
This initially looks okay to me. You need to change the base to `devel`, though, instead of `master`.

stkaplan:
> This initially looks okay to me. You need to change the base to `devel`, though, instead of `master`.

Whoops. Fixed.

jjwilke:
This is fine with me. I'm not a `sed` master, but I don't see why this would break any system where it already works.

PR #469, opened 2019-12-10
Added preview build mode.

feldergast:
Removed --enable-deprecated from configure and add --enable-preview-build.  Used the new flag to remove all the code that is targeted to be removed in SST 10.0.


config/sst_check_deprecated.m4
```diff
@@ -1,21 +0,0 @@
-AC_DEFUN([SST_CHECK_DEPRECATED],
-[
-  sst_check_hdf5_happy="yes"
-
-  AC_ARG_ENABLE([deprecated],
-    [AS_HELP_STRING([--(dis|en)able-deprecated],
-      [Enable or disable deprecated code (Default: enabled)])],
-    [
```
jjwilke: Did you forget to add a file? Or am I just misreading? I don't see where the `CHECK_PREVIEW_BUILD` function is defined.

feldergast: Yes.  Good catch.  File added.  I also pushed some other changes to get rid of the Component* version SubComponent constructor as a valid API.

src/sst/core/baseComponent.cc
```diff
@@ -312,13 +319,15 @@ BaseComponent::configureLink(const std::string& name, TimeConverter* time_base,
                     my_info->link_map = myLinks;
                 }
                 myLinks->insertLink(name,tmp);
+#ifndef SST_ENABLE_PREVIEW_BUILD
                 // Need to set the link's defaultTimeBase to nullptr,
                 // except in the case of this being an Anonymously
                 // loadeed SubComponent, then for backward
                 // compatibility, we leave it as is.
                 if ( !my_info->isLegacySubComponent() ) {
                     tmp->setDefaultTimeBase(nullptr);
```
gvoskuilen: Should defaulttimebase be set to nullptr for all subcomponents if SST_ENABLE_PREVIEW_BUILD is enabled? So the "if" should be ifdef'd but not the call in it? Or is there code elsewhere that takes care of it?

src/sst/core/componentInfo.cc
```diff
@@ -143,12 +143,14 @@ ComponentInfo::ComponentInfo(ComponentInfo &&o) :
 ComponentInfo::~ComponentInfo() {
     if ( link_map ) delete link_map;
     if ( component ) {
+#ifndef SST_ENABLE_PREVIEW_BUILD        
         // For backward compatibility, don't delete component defined
```
gvoskuilen: Similar to the other case, should the 'if' statement be ifdefed but not the contents? So all component defined subcomponents get deleted?

src/sst/core/baseComponent.cc
feldergast: Fixed and pushed.

src/sst/core/componentInfo.cc
feldergast: Fixed and pushed.

PR #476, opened 2020-01-21
Python3

gvoskuilen:
Python3 support for sst-core
- Exports a version flag to sstsimulator.conf
- Preferentially searches PATH for Python2, once we are testing with 3 we should switch this.


src/sst/core/model/element_python.h
```diff
@@ -171,7 +171,12 @@ class SSTElementPythonModule {
     /// \param filename filname used when reporting errors
     ///
     ///     
+#ifdef SST_CONFIG_HAVE_PYTHON3
```
feldergast: Let's just require the filename for both.  I think the only instance of this being used in elements includes a filename (and debugging without the filename can be difficult).

PR #493, opened 2020-02-18
DotConfigOutput

lpsmodsim:
Removed newlines from dot nodes as they error out now


feldergast:
Looks like branch started too far behind devel.  Putting WIP flag on this PR until this is resolved.

PR #495, opened 2020-02-19
Improve Python2 search, more Python info in sst-config

jjwilke:
On some platforms, `python2-config` is fully versioned as `python2.7-config`, etc.

This also makes it possible for elements to query directly for `PYTHON_CONFIG` and `PYTHON` executables, if useful (which it is for sst-macro)


jjwilke:
I don't think we need the check. This is mostly a homebrew/macports issue.  The `python-config`s that are in the path all have full version names because they are installed in a common location like `/opt/local/bin`. The `python-config --prefix` always gives you a location with a `python2` or a `python3` (although we could always attempt to use the more specific name).

gvoskuilen:
Lets leave it as is then.

PR #497, opened 2020-02-25
Split IPC tunnel into defintion and management to enable Pin3

gvoskuilen:


src/sst/core/interprocess/mmapchild_pin3.h
```diff
@@ -0,0 +1,114 @@
...
+            exit(1);
+        }
+
+        shmPtr = NULL;
+        retval = OS_MapFileToMemory(NATIVE_PID_CURRENT, OS_PAGE_PROTECTION_TYPE_READ | OS_PAGE_PROTECTION_TYPE_WRITE, 
+                sizeof(InternalSharedData), OS_MEMORY_FLAGS_SHARED, fd, 0, &shmPtr);
+
+        if (!OS_RETURN_CODE_IS_SUCCESS(retval)) {
+            // Not using Output because IPC means Output might not be available
+            fprintf(stderr, "mmap failed\n");
```
jjwilke: Do the return codes mean anything? If so, we could be be better about printing `strerror`

src/sst/core/interprocess/mmapchild_pin3.h
```diff
@@ -0,0 +1,114 @@
...
+        shmSize = tunnel->getTunnelSize();
+
+        OS_FreeMemory(NATIVE_PID_CURRENT, shmPtr, sizeof(InternalSharedData));
+
+        // Remap with correct size
+        retval = OS_MapFileToMemory(NATIVE_PID_CURRENT, OS_PAGE_PROTECTION_TYPE_READ | OS_PAGE_PROTECTION_TYPE_WRITE,
+                shmSize, OS_MEMORY_FLAGS_SHARED, fd, 0, &shmPtr);
+
+        if (!OS_RETURN_CODE_IS_SUCCESS(retval)) {
+            // Not using Output because IPC means Output might not be available
```
jjwilke: Same question on more error details

src/sst/core/interprocess/mmapchild_pin3.h
```diff
@@ -0,0 +1,114 @@
...
+#define SST_CORE_INTERPROCESS_TUNNEL_MMAP_CHILD_PIN3_H 1
+
+#include "pin.H"
+#include <iostream>
+#include <fstream>
+#include <sys/shm.h>
+#include <sys/stat.h>
+#include "atomic.hpp"
+
+#include "sst/core/interprocess/tunneldef.h"
```
jjwilke: Why does this get included? The `TunnelDef` class doesn't seem to be used.

src/sst/core/interprocess/tunneldef.h
```diff
@@ -0,0 +1,262 @@
...
+    size_t shmSize;
+
+    // Local data
+    size_t numBuffs;
+    size_t buffSize;
+    uint32_t children;
+
+    // Shared objects
+    InternalSharedData *isd;
+    std::vector<CircBuff_t* > circBuffs;
```
jjwilke: I'm not sure we have ABI compatibility between C++11 and C++98. Does this get built with C++11 in SST and then not C++11 in PIN land?

src/sst/core/interprocess/mmapparent.cc
```diff
@@ -0,0 +1,15 @@
...
+// Government retains certain rights in this software.
+//
+// Copyright (c) 2009-2019, NTESS
+// All rights reserved.
+//
+// This file is part of the SST software package. For license
+// information, see the LICENSE file in the top level directory of the
+// distribution.
+
+#include "sst_config.h"
```
jjwilke: Why do we need an implementation file that just includes the header?

src/sst/core/interprocess/mmapchild_pin3.h
gvoskuilen: I'm not sure how detailed the return codes are, the documentation for this function says it can return OS_RETURN_CODE_NO_ERROR or OS_RETURN_CODE_MEMORY_MAP_FAILED but I don't know if it provides more detail than that.

src/sst/core/interprocess/mmapchild_pin3.h
gvoskuilen: Oops, I'll fix it.

src/sst/core/interprocess/tunneldef.h
gvoskuilen: Yes. I also just pushed the ariel-side branch in sst-elements "ariel_pin3" which has all of the Pin stuff.

src/sst/core/interprocess/mmapchild_pin3.h
jjwilke: Ha, okay, so this really is a binary pass/no pass

src/sst/core/interprocess/mmapparent.cc
gvoskuilen: The only reason is to check for syntax errors and such during the core compile, otherwise those aren't caught until sst-elements tries to use it. I don't see a use outside of development, so we could remove them too.

src/sst/core/interprocess/tunneldef.h
jjwilke: We might never ever have a problem.... but.... if we could make this STL-free it would guarantee we don't have an impossible to find bug later. I'm basically 100% sure that pair would have no ABI issues. But, since the first type is also size_t, we could replace it with a `queue_entry<T>`.  I again doubt that `vector` would ever be an issue... but this seems far more likely to have ABI issues between 98 and 11 if we can just use a raw pointer array.

src/sst/core/interprocess/mmapchild_pin3.h
gvoskuilen: I remember now, it needs the size of InternalSharedData from the tunneldef class

src/sst/core/interprocess/mmapparent.cc
jjwilke: Okay. No, I like the strategy actually for debugging in core. We could also do an explicit template instantiation in this file so we also "debug" a template instantiation.

src/sst/core/interprocess/tunneldef.h
gvoskuilen: That's a good point, I'll rework it.

gvoskuilen: Looked at this some more since (it's copied from the original IPCTunnel) and I actually think it is OK. The vector itself lives in this class, not in the shared region. The vector is filled with pointers constructed using offsets stored in a plain array the shared region. The region contains the ISD struct, the shared data struct, and the circular buffers, none of which use anything besides plain data types. Given that, do you still think there might be a problem?

jjwilke: That should be fine, then. My understanding is that 98 and 11 are ABI compatible for exactly the same code (same struct definitions), but there might no be a guarantee that you get the same STL (i.e. 98 and 11 might have different implementations of vector). So if you build with 98 and 11 on POD types, you should be fine.

src/sst/core/interprocess/mmapparent.cc
feldergast: If this file is strictly for catching errors in the header file at compile time, I would add a comment to that effect right after the copyright info.  I would also be nice, though not strictly necessary, to have the file compile using a makefile target that won't link the resulting object file with the executable.

gvoskuilen: There doesn't look to be an easy way to keep the object out of the executable. I will add a comment in these files that they are only for testing and do an explicit instantiation as well.

src/sst/core/interprocess/mmapchild_pin3.h
gvoskuilen: The return value structure also captures errno, so I added better error printing.

PR #500, opened 2020-03-11
Add a newline before each section when registering data into sstsimulator.conf

allevin:
Add a newline before each section when registering data into sstsimulator.conf configuration file.  This makes reading the .conf file a bit easier to the puny human.


jjwilke:
Macro devel is using the new stats stuff. I think you need to rebase onto the most recent devel before this will pass.

Issue #435, opened 2019-08-27
Include paths bug


calewis:
When configuring sst-core, it appears that the top level source directory gets added as an include path for certain targets.  The top level source directory includes the file VERSION.  Newer versions of cstddef include a file <version>.  

OS X is not case sensitive so when building sst-core/VERSION is included into the code, understandably this leads to a compilation failure.  

As far as I can tell there is no reason that the top level source directory should be added as an include path, so the best solution is probably to remove that include from getting added to the includes.

allevin:
@gvoskuilen to investigate

gvoskuilen:
This is fixed by #505

PR #504, opened 2020-03-25
remove top-level source include: Issue #435

jjwilke:
AFAICT, this was included for ltdl stuff. But it doesn't seem to be required for LTDL.


jjwilke:
@gvoskuilen Weird. Not sure how that ended up in there.

jjwilke:
Let's try this again.

PR #505, opened 2020-03-26
remove top-level source include: Issue #435

jjwilke:


jjwilke:
Somehow my devel branch ended up with a non-existent commit. Here is a *clean* PR.

Issue #407, opened 2019-05-23
Compilation Issue for Enabling Event Tracking


amroawad2:
In current devel version in SST core, enabling event tracking by --enable-event-tracking causes a compilation error:

../../../../sst-core/src/sst/core/config.cc: In constructor ‘SST::Config::Config(SST::RankInfo)’:
../../../../sst-core/src/sst/core/config.cc:82:5: error: ‘event_dump_file’ was not declared in this scope
     event_dump_file = "";
     ^~~~~~~~~~~~~~~

amroawad2:
To fix it, you just need to add: 

std::string     event_dump_file;

before line 82 in sst-core/src/sst/core/config.cc

allevin:
Fixed by PR #508

PR #508, opened 2020-03-31
Fix for issue #407

allevin:
This fix is for issue #407, the problem only occurs if both --enable-event-tracking is on and --disable-mem-pools are set.  

In config.h, event_dump_file is only defined if USE_MEMPOOLS is defined (which it is normally).  However, in config.cc if __SST_DEBUG_EVENT_TRACKING__ is defined, then it uses event_dump_file.  

The problem only occurs if both __SST_DEBUG_EVENT_TRACKING__ is defined and USE_MEMPOOLS is not defined -->  --enable-event-tracking is on and --disable-mem-pools configuration settings.

The change is done in configure.ac and checks for either if event tracking is turned on AND mem-pools are turned off, it will then fail the configure with an appropriate error message.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -91,6 +91,10 @@ SST_CHECK_MEM_POOL()
 
 SST_ENABLE_DEBUG_OUTPUT()
 SST_ENABLE_DEBUG_EVENT_TRACKING()
+
+AS_IF([test "$use_mempool" = "no" -a "$enable_debug_event_tracking" = "yes"],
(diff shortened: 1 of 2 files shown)


feldergast:
This probably isn't the right fix.  SST_DEBUG_EVENT_TRACKING will only be useful when MEM_POOLS are enabled.  With mempools, there is no way to track events.  We probably need to make sure in configure that this particular combination isn't allowed and only define the file if both are turned on.

allevin:
@feldergast I undid the change to the configu.h file, and put a check in the configure.ac to detect if we have this specific condition.

PR #513, opened 2020-04-09
Subcomp stat python

feldergast:
Updates the python stat enable API to work with both SubComponents and Components.  Also added the ability to apply changes to children.

### DIFF
--- a/src/sst/core/configGraph.h
+++ b/src/sst/core/configGraph.h
@@ -193,7 +193,7 @@ typedef SparseVectorMap<LinkId_t,ConfigLink> ConfigLinkMap_t;
 class ConfigComponent : public SST::Core::Serialization::serializable {
 public:
     ComponentId_t                 id;                /*!< Unique ID of this component */
-    ConfigComponent*              ulitmate_parent;   /*!< Ultimate parent of this component */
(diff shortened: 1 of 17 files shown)


src/sst/core/configGraph.h
```diff
@@ -283,6 +295,8 @@ class ConfigComponent : public SST::Core::Serialization::serializable {
 // typedef SparseVectorMap<std::string,ConfigLink> ConfigLinkMap_t;
 /** Map IDs to Components */
 typedef SparseVectorMap<ComponentId_t,ConfigComponent> ConfigComponentMap_t;
+/** Map names to Components */
+// typedef std::map<std::string,ConfigComponent*> ConfigComponentNameMap_t;
```
gvoskuilen, line 299: Is this supposed to be commented out? Is it intended for future use?

feldergast: Yes, it's for future use.  Today, the name map is contained in the cpython code.  I'm planning to switch this map over to live in the ConfigGraph, which is really where it belongs.  When I started this, I realized there were more things that should move into the ConfigGraph, like keeping track of the next component id.  I decided to put the code cleanup off and just get the current functionality into level.

PR #523, opened 2020-04-22
Check for missing python library path and manually fix

gvoskuilen:


gvoskuilen:
@feldergast yes both get added. I don't think it matters too much - the old check was doing the same thing. But it would be a little cleaner to remove spurious paths.

PR #526, opened 2020-04-28
Depreceated functions in Simulation that were not intended to be used…

feldergast:
… outside of the core.

The functions will move to an internal implementation class in SST 11.
Also did some minor cleanup of header file includes.

### DIFF
--- a/src/sst/core/baseComponent.cc
+++ b/src/sst/core/baseComponent.cc
@@ -710,9 +710,9 @@ BaseComponent::registerStatisticCore(SST::Params& params, const std::string& sta
 }
 
 void
-BaseComponent::performStatisticOutput(StatisticBase* stat, bool endOfSimFlag)
+BaseComponent::performStatisticOutput(StatisticBase* stat)
 {
-    Simulation::getSimulation()->getStatisticsProcessingEngine()->performStatisticOutput(stat,endOfSimFlag);
+    Simulation::getSimulation()->getStatisticsProcessingEngine()->performStatisticOutput(stat);
 }
 
 void
--- a/src/sst/core/baseComponent.cc
+++ b/src/sst/core/baseComponent.cc
@@ -184,16 +184,35 @@ TimeConverter* BaseComponent::registerTimeBase( const std::string& base, bool re
 }
 
 TimeConverter*
-BaseComponent::getTimeConverter( const std::string& base )
+BaseComponent::getTimeConverter( const std::string& base ) const
 {
     return getSimulation()->getTimeLord()->getTimeConverter(base);
 }
 
 TimeConverter*
-BaseComponent::getTimeConverter( const UnitAlgebra& base )
+BaseComponent::getTimeConverter( const UnitAlgebra& base ) const
 {
     return getSimulation()->getTimeLord()->getTimeConverter(base);
 }
+
+TimeConverter*
+BaseComponent::getTimeConverterNano() const
+{
+    return Simulation::getSimulation()->getTimeLord()->getNano();
+}
+
+TimeConverter*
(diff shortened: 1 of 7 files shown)


src/sst/core/baseComponent.cc
```diff
@@ -648,6 +709,19 @@ BaseComponent::registerStatisticCore(SST::Params& params, const std::string& sta
...
 }
 
+void
+BaseComponent::performStatisticOutput(StatisticBase* stat, bool endOfSimFlag)
+{
+    Simulation::getSimulation()->getStatisticsProcessingEngine()->performStatisticOutput(stat,endOfSimFlag);
+}
+
+void
+BaseComponent::performGlobalStatisticOutput(bool endOfSimFlag)
```
gvoskuilen, line 719: Should components be able to call these with endOfSimFlag=true or is that reserved for the final stat dump done by the core? I would think it should be part of the internal core API and the component API shouldn't have it as an option.

src/sst/core/baseComponent.h
```diff
@@ -329,6 +349,18 @@ class BaseComponent {
...
+      * @param EndOfSimFlag - Indicates that the output is occurring at the end of simulation.
+      */
+    void performStatisticOutput(StatisticBase* stat, bool endOfSimFlag = false);
+
+    /** Performs a global statistic Output.
+     * This routine will force ALL Components and Subcomponents to output their statistic information.
+     * This may lead to unexpected results if the statistic counts or data is reset on output.
+     * @param endOfSimFlag - Indicates that the output is occurring at the end of simulation.
+     */
+    void performGlobalStatisticOutput(bool endOfSimFlag = false);
```
gvoskuilen, line 363: Unless it's been fixed, this function is still not truly global - does not work across threads/ranks. I do not think it's pressing that it be fixed immediately, but it needs to get put in an issue and/or noted in the comment here.

feldergast: Okay, we should open an issue.  This just calls the same function in the stats engine.  I just didn't like having to give a handle to the engine to a component.  I don't think we want to try to fix it before the 10.0 release since getting a truly global stats dump will challenging.  This is only used in one place in the elements repository, do we want to just remove the capability from elements?

src/sst/core/baseComponent.cc
feldergast: No, components should not call it with endOfSimFlag=true.  We can either remove this functionality for elements (it's only used in ember), or I can make the BaseComponent version not take an argument and pass false to the underlying stat engine function.

src/sst/core/baseComponent.h
gvoskuilen: No, Ariel and Ember need the capability. I think we just need to open an issue since this exposes a partially broken function more directly to baseComponents and fix it as soon as possible after 10.0.

src/sst/core/baseComponent.cc
gvoskuilen: The function is needed, Ariel uses it too. But we should remove the endOfSimFlag from this version and pass false to the stat engine.

src/sst/core/baseComponent.cc
```diff
@@ -648,6 +709,19 @@ BaseComponent::registerStatisticCore(SST::Params& params, const std::string& sta
     return statistic;
 }
 
+void
+BaseComponent::performStatisticOutput(StatisticBase* stat, bool endOfSimFlag)
```
gvoskuilen, line 713: Can you remove endOfSimFlag here too?

feldergast: Done

PR #528, opened 2020-05-12
Export a CMake configuration for downstream projects

jjwilke:
This exports a CMake config file compatible with `find_package` for CMake build systems.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -2,6 +2,7 @@
 ACLOCAL_AMFLAGS = -I config
 
 SUBDIRS = \
+	share \
 	src \
 	doc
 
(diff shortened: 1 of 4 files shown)


fryeguy52:
@jjwilke I pushed a commit that added a couple missing options in the cmake calls.  I was not able to get it to work without them in my simple test case.  Is find_package(SST) working for you without those changes?

jjwilke:
Good catch.  Through an amazing set of coincidences, I was pointing to an old install that DID work on my tests.

jjwilke:
I squashed and changed to `INTERFACE`.

Issue #530, opened 2020-05-30
Multiple links on the same port not an error


jjwilke:
If you create multiple links on the same ports this does not produce an error:

````
a = sst.Component(...)
b = sst.Component(...)
link = sst.Link("link1")
link.connect((a,"port0","10ns"),(b,"port0","10ns"))

link = sst.Link("link2")
link.connect((a,"port0","10ns"),(b,"port0","10ns"))
````

This code runs and produces an answer. 
This should at least be a warning - probably an error?

This doubling-up of links occurs in:
https://github.com/sstsimulator/sst-elements/blob/master/src/sst/elements/ariel/frontend/simple/examples/stream/ariel_snb_mlm.py

Related to sst-elements issue: https://github.com/sstsimulator/sst-elements/issues/1489

feldergast:
I'll put in a check for this.  For now it will generate a warning.  In SST 11 it will generate an error.

feldergast:
This now produces an error since PR #614

PR #532, opened 2020-06-09
Minor cleanup of graph generation.

feldergast:
- Added check for multiple use of ports (fixes issue #530)
- Moved component id assignment into ConfigGraph from the python model
- Moved check for duplicate Component names into graph generation from
  CponfigGraph::checkForStructuralErrors().  This was done because of
  the extra memory requirement to do the check after the fact when
  dealing with large graphs.

### DIFF
--- a/src/sst/core/configGraph.cc
+++ b/src/sst/core/configGraph.cc
@@ -460,12 +460,11 @@ ConfigGraph::postCreationCleanup()
 }
 
 
+// Checks for errors that can't be easily detected during the build
+// process
 bool
 ConfigGraph::checkForStructuralErrors()
 {
-    // Output object for error messages
-    Output &output = Output::getDefaultObject();
-
     // Check to make sure there are no dangling links.  A dangling
     // link is found by looking though the links in the graph and
     // making sure there are components on both sides of the link.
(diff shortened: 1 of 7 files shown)


src/sst/core/configGraph.cc
```diff
@@ -715,8 +716,7 @@ ConfigGraph::addLink(ComponentId_t comp_id, const std::string& link_name, const
     }
     ConfigLink &link = links[link_names[link_name]];
     if ( link.current_ref >= 2 ) {
-        cout << "ERROR: Parsing SDL file: Link " << link_name << " referenced more than two times" << endl;
-        exit(1);
+        output.fatal(CALL_INFO,1,"ERROR: Parsing SDL file: Link %s referenced more than two times\n",link_name.c_str());
```
gvoskuilen, line 719: Nitpicky - but this error should probably say "referenced multiple times" or "referenced more than once".

feldergast: A link needs to be referenced two times, one on each “end”.  This catches the case where it tries to connect to a third (or higher, though it would fail before getting to four) port, thus more than two times,

gvoskuilen: Oops, misread that. Marking this resolved.

PR #546, opened 2020-08-04
Added UnitAlgebra support in the python input files.

feldergast:


src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,389 @@
...
+
+// Numerical functions
+static PyObject* unitAlgebraAddIP(PyObject *self, PyObject *other)
+{
+    UnitAlgebraPy_t *self_ua = (UnitAlgebraPy_t*)self;
+    UnitAlgebraPy_t *other_ua = (UnitAlgebraPy_t*)other;
+    *(self_ua->obj) += *(other_ua->obj);
+
+    PyObject *argList = Py_BuildValue("(O)", self);
+    PyObject* ret = PyObject_CallObject((PyObject*)&PyModel_UnitAlgebraType, argList);
```
jjwilke, line 108: This was pretty confusing to me at first about why we have to return a new reference. I would probably add a comment here about why this is being done.  This also gets repeated in several places.  Probably better to just make a function `returnNewInPlaceReference` or something.

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,389 @@
...
+    nullptr,                // unaryfunc nb_positive
+    nullptr,                // unaryfunc nb_absolute
+    nullptr,                // inquiry nb_nonzero (py2) nb_bool (py3)       /* Used by PyObject_IsTrue */
+    nullptr,                // unaryfunc nb_invert
+    nullptr,                // binaryfunc nb_lshift
+    nullptr,                // binaryfunc nb_rshift
+    nullptr,                // binaryfunc nb_and
+    nullptr,                // binaryfunc nb_xor
+    nullptr,                // binaryfunc nb_or
+    SST_NB_COERSE                // coercion nb_coerce Python 2 only
```
jjwilke, line 242: only necessary to appease my spelling tyranny, but COERSE -> COERCE

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,389 @@
...
+    PyObject* ret = PyObject_CallObject((PyObject*)&PyModel_UnitAlgebraType, argList);    
+    Py_DECREF(argList);
+    
+    return ret;
+}
+
+static PyObject* unitAlgebraMul (PyObject *self, PyObject *other)
+{
+    PyObject *argList = Py_BuildValue("(O)", self);
+    PyObject* ret = PyObject_CallObject((PyObject*)&PyModel_UnitAlgebraType, argList);
```
jjwilke, line 168: Is this is a shallow copy?

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,389 @@
...
+
+static PyObject* unitAlgebraMul (PyObject *self, PyObject *other)
+{
+    PyObject *argList = Py_BuildValue("(O)", self);
+    PyObject* ret = PyObject_CallObject((PyObject*)&PyModel_UnitAlgebraType, argList);    
+    Py_DECREF(argList);
+
+    UnitAlgebraPy_t *ret_ua = (UnitAlgebraPy_t*)ret;
+    UnitAlgebraPy_t *other_ua = (UnitAlgebraPy_t*)other;
+    *(ret_ua->obj) *= *(other_ua->obj);
```
jjwilke, line 173: If it is a shallow copy, would this be incorrect behavior? Shouldn't the original unit algebra values remain unmodified?

Even if a deep copy of the Python object, since we are using UnitAlgebra pointers, we are modifying the original UnitAlgebra.

```
a = b + c
``` 
should leave b and c unmodified, but this code would modify b

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,389 @@
...
+    return ret;
+}
+
+static PyObject* unitAlgebraNegate(PyObject *self)
+{
+    PyObject *argList = Py_BuildValue("(O)", self);
+    PyObject* ret = PyObject_CallObject((PyObject*)&PyModel_UnitAlgebraType, argList);    
+    Py_DECREF(argList);
+    UnitAlgebraPy_t *ret_ua = (UnitAlgebraPy_t*)ret;
+    std::cout << ret_ua->obj->toString() << std::endl;
```
jjwilke, line 216: cout should be removed

jjwilke:
We should also sensibly squash the commits (like the 'missing file' one).

jjwilke:
Even if it's not formally part of the test-suite yet, we should add a Python file that unit tests all the algebra functions.  For now we can just run it by-hand and make it part of the formal test suite later.

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: It just uses the underlying = operator on UnitAlgebra, which does create a completely independent copy.

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: Fixed.

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: Resolve with change to direct object rather than pointer.

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: Gone

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: In turns out that with the semantics of python assignment and in-place operators, the in-place and regular math operations ended up being the same code.  I added a comment in the Number methods declaration to explain that.

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,338 @@
...
+        unitAlgebraGetRoundedValue, METH_NOARGS,
+        "Rounds value of UnitAlgebra to nearest whole number and returns a long"},
+    {   "hasUnits",
+        unitAlgebraHasUnits, METH_VARARGS,
+        "Checks to see if the UnitAlgebra has the specified units"},
+    {   "invert",
+        unitAlgebraInvert, METH_NOARGS,
+        "Inverts the UnitAlgebra value and units"},
+    // {   "str",
+    //     unitAlgebraStr, METH_NOARGS,
```
jjwilke, line 263: What is the plan for this? Is there a reason it's commented?

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,338 @@
...
+    }
+    return Py_NotImplemented;
+}
+
+
+// Numerical functions
+static PyObject* unitAlgebraAdd (PyObject *self, PyObject *other)
+{
+    PyObject *argList = Py_BuildValue("(O)", self);
+    PyObject* ret = PyObject_CallObject((PyObject*)&PyModel_UnitAlgebraType, argList);
```
jjwilke, line 103: We execute this block of code in many different places. The code would probably be a lot more maintainable if we create the new references in a utility function.

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,338 @@
...
+
+// Numerical functions
+static PyObject* unitAlgebraAdd (PyObject *self, PyObject *other)
+{
+    PyObject *argList = Py_BuildValue("(O)", self);
+    PyObject* ret = PyObject_CallObject((PyObject*)&PyModel_UnitAlgebraType, argList);    
+    Py_DECREF(argList);
+    
+    UnitAlgebraPy_t *ret_ua = (UnitAlgebraPy_t*)ret;
+    UnitAlgebraPy_t *other_ua = (UnitAlgebraPy_t*)other;
```
jjwilke, line 107: I don't know how to check or fix this - but if someone does math with a type other than UnitAlgebra we get a seg-fault.  Ideally you would rather have a value exception that neatly said "Can only use unit algebra with other unit algebra"

src/sst/core/model/python/pymodel_unitalgebra.cc
```diff
@@ -0,0 +1,338 @@
...
+    nullptr,                // unaryfunc nb_index
+    SST_NB_MATRIX_MULTIPLY     // py3 only
+    SST_NB_INPLACE_MATRIX_MULTIPLY // py3 only
+}; 
+    
+// Other methods
+
+static PyObject* unitAlgebraGetRoundedValue(PyObject *self, PyObject *UNUSED(args))
+{
+    return unitAlgebraToLong(self);
```
jjwilke, line 227: I have been testing this with "10ns" and I keep getting back 0.

jjwilke, line 227: Does it make sense to also allow casting to float?

jjwilke: GHz does work

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: No, that's just a remnant from when I didn't know how to actually do the str function.  I'll make another pass through looking for commented out blocks to remove.

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: That's a good idea.  I will add one.

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: That is a true statement, though we could actually make it work with some other types like integer and strings that properly initialize a UnitAlgebra.  The c++ version of the class does that for convenience.  Is it worth doing here as well?

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: That is the correct behavior.  The value Is 1e-8, so it rounds down to zero when cast to an int.

feldergast: Yes, I considered that but UnitAlgebra didn't support that cast.  However, looking at the code, the underlying data structure does have a toDouble, so we could make this work.

src/sst/core/model/python/pymodel_unitalgebra.cc
feldergast: I added error checking for invalid types to all UnitAlgebra functions and formatted them to match other similar errors in Python 3 (the default python 2 errors are slightly different, but I didn't try to change the error messages for UnitAlgebra based on version).

PR #548, opened 2020-08-25
Params fix

feldergast:
Improved the parsing of of array string for Params::find_array(). Also added repr() function to the UnitAlgebra python object so it worked correctly when used in lists and added as a param.

Also, improved the API in BaseComponent to check to see if a SubComponent was loadable with a specific API.


src/sst/core/baseComponent.h
```diff
@@ -404,6 +405,39 @@ class BaseComponent {
...
+       @return True if loadable as the API specified as the template parameter
+     */
+    template<class T>
+    bool isUserSubComponentLoadableUsingAPI(const std::string& slot_name) {
+        // Get list of ComponentInfo objects and make sure that there is
+        // only one SubComponent put into this slot
+        // const std::vector<ComponentInfo>& subcomps = my_info->getSubComponents();
+        const std::map<ComponentId_t,ComponentInfo>& subcomps = my_info->getSubComponents();
+        int sub_count = 0;
+        int index = -1;
```
gvoskuilen, line 424: Why is index needed? It doesn't look like it's used.

src/sst/core/params.cc
```diff
@@ -242,6 +242,140 @@ Params::getKey(const std::string& str)
...
+ * you may need to create an array string manually.  Also, any string
+ * that starts with a quote character, must end with the same quote
+ * character.
+
+ * The string passed into this function should have the [] already
+ * removed from the string.  Tokens are generated by splitting the
+ * string on commas that are not within quotes (double or single).
+ * All whitespace at the beginning and end of a token is ignored
+ * (unless inside quotes).  Once the tokens are generated, any quoted
+ * string will have the front and back quotes removed.  The '/' for
```
gvoskuilen, line 275: Should the '/' be '\' ?

feldergast: Fixed here and in params.h.  Also added some examples of using double and single quotes in strings and how they show up after parsing.

src/sst/core/baseComponent.h
feldergast: The index is needed, and the 0 in the return statement has been changed to use index.  This matches the semantics of loadUserSubComponent(), which requires only one SubComponent to be loaded into the slot, but it does not require that the one SubComponent be loaded into index 0.

PR #552, opened 2020-09-11
 Implemented Statistic Object creation - refactored

mperrinel:
This PR aims to provide a statistic object that can be manipulated from the python file.

It has been done on the top of the **Implemented Statistic Object creation #537 one**
You will find the previous commits and the two new ones that aims to make the statistic shareable.


jjwilke (approved):
Okay to test

jjwilke (approved):
Okay to test

jjwilke (approved):
okay to test

jjwilke (approved):
okay to test

jjwilke (dismissed):
Please run tests?

jjwilke (approved):
Okay to test

jjwilke (approved):
Okay to test

allevin:
@jjwilke @mperrinel Our build/test machine is having some issues and preventing PR's from being tested... Hope to resolve soon.

Also, is this a repeat of PR #537?  If it is, can we close out #537?

jjwilke:
@allevin Thanks for looking into it. It's not an exact repeat. We are trying to complete a more major cleanup, but aren't sure we can get everything working in a timely manner. If we decide the changes are too much for now, we will switch this to WIP and just merge the other one.

jjwilke (dismissed):
Okay to test

jjwilke (dismissed):
Okay to test

allevin:
@mperrinel @jjwilke Github is reporting merge conflicts that must be resolved.  I labeled this as WIP till the conflict is resolved to avoid Autotester Spam.  When fixed remove the Label.

allevin:
@jjwilke Does this PR need to go into the 10.1 release?

jjwilke:
This does not need to go into 10.1

mperrinel:
@allevin @jjwilke 
I did the necessary fo fix the Merge conflict.
The WIP can be removed.

jjwilke (approved):
Okay to test

jjwilke (approved):
Okay to test

jjwilke (approved):
Okay to test

src/sst/core/baseComponent.h
```diff
@@ -272,6 +270,31 @@ class BaseComponent {
...
 
 
+    template <typename T>
+    Statistics::Statistic<T>* createStatistic(SST::Params& params, StatisticId_t id,
+                                              const std::string& name, const std::string& statSubId){
+      std::string type = configureStatParams(id, params);
+      auto* engine = Statistics::StatisticProcessingEngine::getInstance();
+      Statistics::StatisticBase* existing = findRegisteredStatistic(name, statSubId, StatisticFieldType<T>::id());
+      if (existing){
+        return dynamic_cast<Statistics::Statistic<T>*>(existing);
```
jjwilke, line 280: This dynamic cast should be checked for a null error

src/sst/core/componentInfo.h
```diff
@@ -26,6 +26,13 @@ class LinkMap;
 class BaseComponent;
 
 class ConfigComponent;
+
+namespace Experimental {
```
jjwilke, line 30: I think we are okay to remove experimental namespace

src/sst/core/model/python/pymodel_comp.cc
```diff
@@ -271,6 +268,63 @@ static PyObject* compSetSubComponent(PyObject *self, PyObject *args)
...
+    }
+
+    char errMsg[1024] = {0};
+    snprintf(errMsg, sizeof(errMsg)-1, "Failed to create statistic %s on %s. Already attached a statistic with that name ?\n", name, c->name.c_str());
+    PyErr_SetString(PyExc_RuntimeError, errMsg);
+    return nullptr;
+}
+*/
+
+static PyObject* compSetStatistic(PyObject *self, PyObject *args)
```
jjwilke, line 304: We need the assign statistic function to re-use an existing stat object

src/sst/core/configGraph.cc
```diff
@@ -257,52 +276,47 @@ void ConfigComponent::addParameter(const std::string& key, const std::string& va
...
-            // Add a STATALLFLAG to end of the vector
-            enabledStatistics.emplace_back(STATALLFLAG);
-        }
+    
+    StatisticId_t stat_id;
+    if (statisticName == STATALLFLAG){
+      // special sentinel id for enable all
+      enabledAllStats = true;
+      stat_id = STATALL_ID;
+      enabledStatNames[statisticName] = stat_id;
```
jjwilke, line 295: Add some explanation here about local versus global IDs

src/sst/core/statapi/statengine.cc
```diff
@@ -86,12 +86,12 @@ StatisticProcessingEngine::~StatisticProcessingEngine()
...
 }
 
-bool StatisticProcessingEngine::registerStatisticCore(StatisticBase* stat, uint8_t comp_load_level)
+bool StatisticProcessingEngine::registerStatisticCore(StatisticBase* stat)
 {
     if ( stat->isNullStatistic() )
         return true;
 
-    uint8_t stat_load_level = comp_load_level == STATISTICLOADLEVELUNINITIALIZED ? m_statLoadLevel : comp_load_level;
+    uint8_t stat_load_level = m_statLoadLevel;
```
jjwilke, line 94: We can have a component load level, which needs to be preserved. I think we ended up changing this logic.

jjwilke (changes requested):

jjwilke (approved):
Okay to test

mperrinel:
I rebased on the latest devel branch.

jjwilke (approved):
Okay to test

jjwilke:
@feldergast @gvoskuilen Ready for your review

src/sst/core/baseComponent.cc
```diff
@@ -535,188 +536,121 @@ uint8_t BaseComponent::getComponentInfoStatisticEnableLevel(const std::string& s
...
+  //make sure we have a valid collection rate
+  // Check that the Collection Rate is a valid unit type that we can use
+  if (collectionRate.hasUnits("s") || collectionRate.hasUnits("hz")) {
+      // Rate is Periodic Based
+      statCollectionMode = StatisticBase::STAT_MODE_PERIODIC;
+  } else if (collectionRate.hasUnits("event")) {
+      // Rate is Count Based
+      statCollectionMode = StatisticBase::STAT_MODE_COUNT;
+  } else if (collectionRate.getValue() == 0) {
+      // Collection rate is zero and has no units
```
feldergast, line 556: This isn't technically true.  The getValue() check is only looking for a value of zero, regardless of the units.  All we know is that it doesn't have one of the three units we are looking for.  Probably not a big deal since it's only a comment, but the comment is wrong.

src/sst/core/baseComponent.cc
```diff
@@ -535,188 +536,121 @@ uint8_t BaseComponent::getComponentInfoStatisticEnableLevel(const std::string& s
...
-        // Tell the Statistic what collection mode it is in
-        statistic->setRegisteredCollectionMode(statCollectionMode);
-    }
+Statistics::StatisticBase*
+BaseComponent::createExplicitlyEnabledStatistic(Params &params, StatisticId_t id,
+                                                const std::string &name, const std::string &statSubId,
+                                                StatCreateFunction fxn)
+{
+  if (my_info->parent_info){
+    fatal(__LINE__, __FILE__, "createExplicitlyEnabledStatistic", 1,
```
feldergast, line 619: Why not just use the CALL_INFO or CALL_INFO long macro here instead of explicitly doing __LINE__, etc.?

src/sst/core/baseComponent.cc
jjwilke: Yeah, we can definitely fix the comment. I think that comment might pre-date the changes, actually.

src/sst/core/baseComponent.cc
jjwilke: No reason in particular. Just didn't think to use it. We can change that to match other calls.

src/sst/core/baseComponent.h
```diff
@@ -272,6 +279,51 @@ class BaseComponent {
...
+             name.c_str());
+        return nullptr; //avoid compiler warnings
+      }
+    }
+
+
+    template <typename T>
+    Statistics::Statistic<T>* createNullStatistic(SST::Params& params, const std::string& name, const std::string& statSubId = ""){
+      auto* engine = Statistics::StatisticProcessingEngine::getInstance();
+      return engine->createStatistic<T>(my_info->component, "sst.NullStatistic", name, statSubId, params);
```
feldergast, line 324: I know we’ve always done it this way, but do we need to create a new NullStatistic for each stat that isn’t enabled, or can we just have a single NullStatistic object that we give to everyone.  There wouldn’t be any data associated with it, just an addData() function that immediately returns.  Not sure this is high priority, because it probably wouldn’t save much memory.

feldergast (commented):
Overall, things are looking good.  One general thing, is that you should ensure that all indentation is set to 4 to conform with the coding standard.  It looks like some of the new code only has an indent of 2.  There are also some other specific comments in various files.

Also, I can't quite envision the full API for reusing Stat objects.  I think it would be good (for this reason and just in general) to add a test to the new core test suite that would then include a sample file.  We can get you in touch with the right person for help with this.

src/sst/core/configGraph.cc
```diff
@@ -421,6 +481,43 @@ ConfigComponent* ConfigComponent::findSubComponentByName(const std::string& name
...
+    auto* parent = getParent();
+    if (parent){
+      return parent->findStatistic(sid);
+    } else {
+      auto iter = statistics.find(sid);
+      if (iter == statistics.end()){
+        return nullptr;
+      } else {
+        //I hate that I have to do this
+        return const_cast<ConfigStatistic*>(&iter->second);
```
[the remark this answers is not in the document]
mperrinel: The method is const, so we cannot use it to return a pointer that allows the modification of a member variable of the class. &iter->second refers to the statistic member of the class, which is here const because the method is const.
That's why we need to const_cast it.

src/sst/core/configGraph.cc
```diff
@@ -257,52 +273,95 @@ void ConfigComponent::addParameter(const std::string& key, const std::string& va
...
 }
 
-void ConfigComponent::enableStatistic(const std::string& statisticName, bool recursively)
+ConfigStatistic* ConfigComponent::createStatistic()
+{
+    StatisticId_t stat_id = getNextStatisticID();;
+    auto* parent = getParent();
+    ConfigStatistic* cs = nullptr;
+    if (parent){
+      cs = parent->insertStatistic(stat_id);
```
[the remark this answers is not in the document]
mperrinel: :+1:

src/sst/core/configGraph.cc
```diff
@@ -176,7 +186,7 @@ ConfigComponent::cloneWithoutLinksOrParams() const
     ret.weight = weight;
     ret.rank = rank;
     ret.statLoadLevel = statLoadLevel;
-    ret.enabledStatistics = enabledStatistics;
+    ret.statistics = statistics;
```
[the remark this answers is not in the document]
mperrinel: :+1:

src/sst/core/baseComponent.cc
mperrinel: :+1:

src/sst/core/model/python/pymodel_comp.cc
```diff
@@ -271,6 +268,49 @@ static PyObject* compSetSubComponent(PyObject *self, PyObject *args)
     return nullptr;
 }
 
+static PyObject* compReuseStatistic(PyObject *self, PyObject *args)
```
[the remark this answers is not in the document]
mperrinel: It is bound to the subComponentMethods : 
```
{   "reuseStatistic",
        compReuseStatistic, METH_VARARGS,
        "Reuse a statistic for the binding"},
```

This method is a new method used from the python parameter file to reuse on a subcomponent  a statistic created from a parent component.

src/sst/core/baseComponent.h
jjwilke: We could potentially modify NullStatistic to avoid some of the overheads in the `addData`function, but I think it would be too complicated for the benefit to change to a single `NullStatistic`

src/sst/core/configGraph.cc
```diff
@@ -257,52 +273,95 @@ void ConfigComponent::addParameter(const std::string& key, const std::string& va
...
+    }
+    
+    ConfigStatistic& cs = statistics[stat_id];
+    cs.id = stat_id;
+    return &cs;
+}
+
+void ConfigComponent::reuseStatistic(const std::string& statisticName, StatisticId_t sid) {
+
+    if(statisticName == STATALLFLAG){
```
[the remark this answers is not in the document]
jjwilke: Yes, that will work. A reused statistic is explicitly enabled. Stats first look for explicitly enabled configurations - and if not there - will look for an enabled all configuration. So it's totally fine to reuse some statistics and enable all for the rest.

mperrinel:
I rebased on the latest devel

src/sst/core/configGraph.cc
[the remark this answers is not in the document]
feldergast: In that case, it seems like the function shouldn't be marked const.  I assume that this function is called some place on a ConfigComponent that is marked const??

src/sst/core/model/python/pymodel_comp.cc
```diff
@@ -409,6 +456,51 @@ static PyObject* compEnableStatistics(PyObject *self, PyObject *args)
...
+    //mandatory name
+    //optional subid
+    //optional parameters
+    PyObject* name = nullptr;
+    PyObject* subid = nullptr;
+
+    if ( !PyArg_ParseTuple(args, "O|O", &name, &subid) )
+        return nullptr;
+
+    if ( nullptr == comp ) return nullptr;
```
feldergast, line 477: This check is redundant.  Already checked on line 463.

src/sst/core/model/python/pymodel_comp.cc
feldergast: setStatistic seems to repeat most, if not all, of the code in createStatistic.  The only difference seems to be that setStatistic also enables the statistic.  Does it make sense to call the create code in the set function so that the creation code isn't redundant and will be easier to maintain should we chance some of the initialization in the future?

tests/test_Component_createStats.py
```diff
@@ -0,0 +1,47 @@
+# Automatically generated SST Python input
+import sst
+
+# Define SST core options
+sst.setProgramOption("timebase", "1 ps")
```
feldergast, line 5: Remove this line. In general, we are trying to discourage people from setting timebase.  We made the mistake of adding it into early examples.

tests/test_Component_enableAllStats.py
```diff
@@ -0,0 +1,43 @@
+# Automatically generated SST Python input
+import sst
+
+# Define SST core options
+sst.setProgramOption("timebase", "1 ps")
```
feldergast, line 5: Remove this line.

tests/test_Component_setStats.py
```diff
@@ -0,0 +1,47 @@
+# Automatically generated SST Python input
+import sst
+
+# Define SST core options
+sst.setProgramOption("timebase", "1 ps")
```
feldergast, line 5: Remove this line.

src/sst/core/testElements/coreTest_Component.h
```diff
@@ -84,6 +84,10 @@ class coreTestComponent : public SST::Component
     SST::Link* S;
     SST::Link* E;
     SST::Link* W;
+    SST::Statistics::Statistic<int>* countN;
```
feldergast, line 87: Did these get added to the ELI information for the class?

feldergast (changes requested):

jjwilke:
Closed in favor of PR #586

PR #553, opened 2020-10-08
Add extra hooks for registering components with intel

jjwilke:
Intel does not generate symbols for functions or variables that are only found in header files. It is the only compiler to do so.  This requires us to force instantiation of certain types in the source files. 

This fixes Intel compilers via two steps:
1. A macro is provided to force registration in source fields
1. An undefined reference is generated automatically if the instantiate macro is missing

### DIFF
--- a/src/sst/core/eli/elementinfo.h
+++ b/src/sst/core/eli/elementinfo.h
@@ -230,7 +230,6 @@ struct InfoLoader : public LibraryLoader {
 
 template <class Base> void InfoLibrary<Base>::addLoader(const std::string& elemlib, const std::string& elem,
                                                         BaseInfo* info){
-
    auto loader = new InfoLoader<Base,BaseInfo>(elemlib, elem, info);
    LoadedLibraries::addLoader(elemlib, elem, loader);
 }
@@ -255,6 +254,14 @@ struct InfoDatabase {
     static InfoLibrary<T>* getLibrary(const std::string& name){
         return InfoLibraryDatabase<T>::getLibrary(name);
     }
+};
+
+void force_instantiate_bool(bool b, const char* name);
+
+template <class T> struct ForceExport {
+  static bool ELI_isLoaded() {
+    return T::ELI_isLoaded();
+  }
 };
 
 } //namespace ELI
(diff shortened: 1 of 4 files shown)


src/sst/core/eli/elementinfo.h
```diff
@@ -328,11 +327,35 @@ constexpr unsigned SST_ELI_getTertiaryNumberFromVersion(SST_ELI_element_version_
   bool ELI_isLoaded(); \
   SST_ELI_DEFAULT_INFO(lib,name,ELI_FORWARD_AS_ONE(version),desc)
 
+// The Intel compilers do not correctly instantiate symbols
```
feldergast, line 330: Does this mean that we also need to add SST_ELI_INSTANTIATE for all the elements in sst/elements in order to make them work with Intel.

jjwilke: Yes... there seems to be no way to force Intel to generate symbols (even for functions!!!) in the header file if they aren't actually used in a source file.

The best I could do was generate a linker error or dlopen error telling you that you forget the macro.

feldergast: Okay, I think by SST 11 we probably want to remove the guards that only do the checking for the Intel compiler.  Otherwise it will be consistently broken for Intel because people will add new elements and forget to add the instantiate.  Do you want to add SST_ENABLE_PREVIEW_BUILD guards so that when preview build is turned on, it doesn't check for compiler version?

nmhamster: Can we not do this automatically when a user registers a component using our existing defines .. please? I definitely don't want to add special code for Intel in every component.

jjwilke: I have one more thing I can try... but the Intel compiler is pretty stubborn in not generating symbols.

jjwilke: Okay, I basically have this working.  If you declare your class in a header file, I got this working. You don't need any extra macros and elements should work as-is.

Where this won't work with Intel is if you implement everything entirely in a source file.  You have to declare your class in a header file or Intel simply refuses to generate symbols.  We can still support that use case, but you'll need to use an `SST_ELI_EXPORT` macro.

feldergast: This seems reasonable to me.  Most of the elements are defined in header files, so it shouldn't be too much of an issue to require an extra call when you don't do that.  I approve the changes, but will wait to see if Si has any thoughts before officially approving.

nmhamster: let's try it!

nmhamster (approved):

nmhamster:
@feldergast / @jjwilke - I think we should get this merged, the testing looks robust (many repeats!).

PR #555, opened 2020-10-15
Fixes a Bug in UniformDistribution where bin-0 may not get entries

nmhamster:
Fixes a bug in SSTUniformDistribution where the 0th-bin may not get entries generated by the underlying RNG.


nmhamster:
@feldergast - I think this may break a Merlin test because it is now generating uniforms correctly. Just FYI.

gvoskuilen (approved):

PR #557, opened 2020-10-16
Graph fix

feldergast:


src/sst/core/main.cc
```diff
@@ -426,6 +426,9 @@ main(int argc, char *argv[])
     double start_graph_gen = sst_get_cpu_time();
     graph = new ConfigGraph();
 
+    // Need to initialize TimeLord before we use UnitAlgebra
+    Simulation::getTimeLord()->init(cfg.timeBase);
```
jjwilke, line 430: This breaks macro here:
````python
  sst.setProgramOption("timebase", "100as")
````
Because you're now setting the timebase before the config graph has been processed, elements can no longer set program options. Later on, we create links with "100as" timebase, which now crashes because the timebase is always "1ps" no matter what options you set in your python file.

File: src/sst/core/main.cc [hunk 2]
```diff
@@ -426,6 +426,9 @@ main(int argc, char *argv[])
...
+
```
jjwilke, line 431: The `setProgramOption` function needs to propagate to TimeLord.

````cpp
Config::setTimebase(const std::string& arg){
  timebase = arg;
  Simulation::getTimeLord()->init(cfg.timeBase);
}

Making this change fixes the issue and all tests now pass.  This is, however, not a great way to fix things. There's a vague comment that we need to initialize TimeLord before using UnitAlgebra. It's not clear to me why these depend on each other.

**Review by jjwilke — CHANGES_REQUESTED** [review]

File: src/sst/core/main.cc (hunk 2 above)
Comment (feldergast) [reply src/sst/core/main.cc:431]: Looking at the code, the dependency between TimeLord and UnitAlgebra is gone.  I think when it was first written it stored things based on the timebase, but that hasn't been true for a while.  We should be able to safely reinitialize the TimeLord during the python execution.  Nothing is converted into TimeConverters until after the python is complete.

Comment (feldergast) [reply src/sst/core/main.cc]: For now, I just moved the initialization until after the python completes.  Probably the better way to give access to the timebase in the python is to look directly at the config object, which you can already do.  I may add a specific function to get the timebase later if it is needed to make it easier.

File: src/sst/core/configGraph.cc [hunk 3]
```diff
@@ -904,45 +922,42 @@ ConfigGraph::getCollapsedPartitionGraph()
     PartitionComponentMap_t& pcomps = graph->getComponentMap();
     PartitionLinkMap_t& plinks = graph->getLinkMap();
 
+
+    // Mark all Components as not visited
+    for ( ConfigComponentMap_t::iterator it = comps.begin(); it != comps.end(); ++it ) it->visited = false;
```
gvoskuilen, line 927: Is this needed? It looks like you initialize visited to false in the constructor.

src/sst/core/impl/partitioners/singlepart.h
```diff
@@ -45,13 +45,12 @@ class SSTSinglePartition : public SST::Partition::SSTPartitioner {
        Performs a partition of an SST simulation configuration
        \param graph The simulation configuration to partition
     */
-    void performPartition(PartitionGraph* graph) override;
-
-    void performPartition(ConfigGraph* graph) override {
+    void performPartition(PartitionGraph* graph) override {
         SST::Partition::SSTPartitioner::performPartition(graph);
```
jjwilke, line 49: Is this code ever intended to be called? My understanding is the because `requiresConfigGraph` returns true this always routes to config graph function, not the partition graph function. This wouldn't actually produce a valid partition, would it? Maybe we don't need to fix for this PR, but would it be more appropriate for `performPartition(PartitionGraph* graph)` to crash and exit?

jjwilke:
Does this link to an issue or can we add a brief comment about what this fixes?

gvoskuilen:
@jjwilke we should probably just open a new issue and point to that. I don't think one exists already...

src/sst/core/configGraph.cc
```diff
@@ -29,11 +29,29 @@ using namespace std;
...
+// static bool zero_latency_warning = false;
 
 void ConfigLink::updateLatencies(TimeLord *timeLord)
 {
+    // Need to clean up some elements before we can test for zero latency
     latency[0] = timeLord->getSimCycles(latency_str[0], __FUNCTION__);
+    // if ( latency[0] == 0 ) {
+    //     latency[0] = 1;
+    //     if ( !zero_latency_warning ) {
+    //         Output::getDefaultObject().output("WARNING: Found zero latency link.  Setting all zero latency links to a latency of %s\n",
```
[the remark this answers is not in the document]
feldergast: Yes, in the past we disallowed zero latency links.  The check got removed somewhere along the way and the scheduler element library seems to use them.  Scheduler is being removed for SST 11, so this code will be enabled for that release.

src/sst/core/configGraph.cc
feldergast: Not strictly and I debated on whether to put it in or not.  I thought it safest to put it in since there is technically nothing stopping it from being modified after initialization.

src/sst/core/configGraph.cc
```diff
@@ -904,45 +922,42 @@ ConfigGraph::getCollapsedPartitionGraph()
...
-                    const ConfigLink& link = links[id];
-
-                    if ( !group.contains(COMPONENT_ID_MASK(link.component[0])) || !group.contains(COMPONENT_ID_MASK(link.component[1]) ) ) {
-                        pcomp.links.push_back(link.id);
-                    }
-                    else {
-                        deleted_links.insert(link.id);
-                    }
+        ComponentId_t id = pcomps.size();
+        pcomps.insert(PartitionComponent(id));
```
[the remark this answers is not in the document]
feldergast: We used to use the built-in map types, but they use a lot of extra memory, which for large graphs were sometimes causing us to run out of memory on rank 0.  Changing to a dense data structure fixed that problem.  It may not be an issue with the memory sizes of today, but it was in the past.  Making the change to return a reference would be trivial.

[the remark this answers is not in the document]
nmhamster: @feldergast - the issue came up a lot running on Haswell (Mutrino) with large network simulations. I would still like to be able to use this platform for our larger runs if we can - i.e. I think your approach with the dense structures is still needed right now and I don't see that changing too much for another year or two.

src/sst/core/impl/partitioners/singlepart.h
feldergast: Only one of the versions will ever get called.  The off function is calling the parent function, which does just fatal.  Not sure why it just doesn't declare the one that isn't used because it would do the same thing but the parent function would get called directly.  I suspect the other partitioners are doing the same thing.  It functions the way you suggested, even through it's done in a weird way.  We should probably look at cleaning this up across all partitioners in a different PR.

jjwilke (approved):
Looks good to me. All comments have been addressed and there are enough comments now to document the changes.

feldergast:
Submitted issues #561 and #562 to address comments about partitioners and SparseVectorMap.

PR #559, opened 2020-10-20
Moving MPI check to only be required if Ranks > 1

allevin:
Moving test to check for existence of OpenMPI so that it does not generate errors on rank=1 tests in Jenkins.   Also changed how we compute the numa_param.  Lastly added some support code to allow tests to get the diff file on compares.

### DIFF
--- a/src/sst/core/testingframework/sst_unittest.py
+++ b/src/sst/core/testingframework/sst_unittest.py
@@ -307,13 +307,14 @@ class SSTTestCase(unittest.TestCase):
         # Update the os launch command if we are running multi-rank
         num_cores = host_os_get_num_cores_on_system()
 
-        # Check to see if mpirun is available
+        # Perform any multi-rank checks/setup
         mpi_avail = False
-        rtn = os.system("which mpirun > /dev/null")
-        if rtn == 0:
-            mpi_avail = True
-
         if num_ranks > 1:
+            # Check to see if mpirun is available
+            rtn = os.system("which mpirun > /dev/null")
+            if rtn == 0:
+                mpi_avail = True
+
             numa_param = ""
             if 2 <= num_cores <= 4:
                 numa_param = "-map-by numa:pe=2 -oversubscribe"
--- a/src/sst/core/testingframework/sst_unittest.py
+++ b/src/sst/core/testingframework/sst_unittest.py
@@ -309,17 +309,14 @@ class SSTTestCase(unittest.TestCase):
 
         # Perform any multi-rank checks/setup
         mpi_avail = False
+        numa_param = ""
         if num_ranks > 1:
             # Check to see if mpirun is available
-            rtn = os.system("which mpirun > /dev/null")
+            rtn = os.system("which mpirun > /dev/null 2>&1")
             if rtn == 0:
                 mpi_avail = True
 
-            numa_param = ""
-            if 2 <= num_cores <= 4:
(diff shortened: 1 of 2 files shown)


nmhamster (changes requested):
@allevin - I am wondering if we get the user to supply ranks and threads and use this to set `numa:PE=<num threads>` and we get rid of the oversubscribe. We can then rework our Jenkins jobs so that we don't get invalid/non-useful combinations (that require `-oversubscribe`. In general over-subscription is just going to give us very slow execution. This isn't in this specific change, but I am wondering about the code just below (where `-oversubscribe` is set).

allevin:
> @allevin - I am wondering if we get the user to supply ranks and threads and use this to set `numa:PE=<num threads>` and we get rid of the oversubscribe. We can then rework our Jenkins jobs so that we don't get invalid/non-useful combinations (that require `-oversubscribe`. In general over-subscription is just going to give us very slow execution. This isn't in this specific change, but I am wondering about the code just below (where `-oversubscribe` is set).

@nmhamster The code that you see below is pulled (and converted from bash to python) from what our bamboo script is doing right now:
```
        # Update the os launch command if we are running multi-rank
        num_cores = host_os_get_num_cores_on_system()

        # Perform any multi-rank checks/setup
        mpi_avail = False
        if num_ranks > 1:
            # Check to see if mpirun is available
            rtn = os.system("which mpirun > /dev/null")
            if rtn == 0:
                mpi_avail = True

            numa_param = ""
            if 2 <= num_cores <= 4:
                numa_param = "-map-by numa:pe=2 -oversubscribe"
            elif num_cores >= 4:
                numa_param = "-map-by numa:pe=2"

            oscmd = "mpirun -np {0} {1} -output-filename {2} {3}".format(num_ranks,
                                                                         numa_param,
                                                                         mpiout_filename,
                                                                         oscmd)

Note: oscmd = "sst -n X ..." if num_threads is > 1

```
We have 3 main parameters to consider...
- ```num_cores``` = Number of cores on the system (from the OS)
- ```num_ranks``` = User requested (via command line argument) number of ranks to test with
- ```num_threads``` = User requested (via command line argument) number of threads to test with

I do not fully understand what the -map-by numa:pe=XX does?   Also, I am unsure how to formulate the numa_param differently based upon the number of cores...  If you show me what you want, I will roll it into this PR.

nmhamster:
@allevin - I don't think a direct conversation is really what we want to do anymore (i.e. I think we want flexible rank/thread combinations, not fixed). I realize this is a change to where we were at before. The `numa:PE=XX` means give me `XX` cores for each rank, so if we ask for 4 ranks with `PE=2` it means give me 4 MPI ranks, each rank must have 2 cores, so we use 8 cores total. If MPI doesn't see 8 cores its an error. We don't need 8 cores in *every* single test, it would be good to read this from the command line and run out tests against whatever the user provides. This is partially to make our virtual machine environment flexible but also because users are going to be running on laptops (with less than 8 cores) right the way up to servers with loads of cores (much more than 8).

allevin:
I made the change.   However, some of the tests are failing when I run -r 2 -t 2 that were passing before.    I need to audit the tests to see what is falling down.  I will be poking you for more help soon.

--
Aaron

allevin:
@nmhamster I fixed the issues with the failing tests when -r 2 and -t2 is used.  I think this PR is ready to go.

nmhamster (approved):

PR #569, opened 2020-12-11
Improvements for New Test Frameworks

allevin:
Bug Fixes, improvements & features to the new test frameworks.
- All changes are specific to the new frameworks and will have no impact on sst-core operations.


src/sst/core/testingframework/sst-test-elements
```diff
@@ -18,6 +18,22 @@
 import sys
 import os.path
 
+# Validate Python Versions
```
gvoskuilen, line 21: Probably not a big deal since there's only two places that I see this, but it might be good to have a python version check function that's shared so we only have to change it later in one place.

allevin: Agreed, but I would like to leave it for now.  Those 2 files are the very light wrappers that launch the test engine.  I don’t want to make a “common” module that they would both load just to do that check, because it would be another file that we would have to copy over.

gvoskuilen (approved):

PR #583, opened 2021-02-16
2021 copyright update

allevin:
Updated all 2020 copyright to 2021


feldergast (approved):
Not sure if it was intentional or not, but I like the fact that all the trailing whitespace was deleted.  I think we should make that part of the copyright update each year.

allevin:
My editor normally automatically does this.  I meant to turn that feature off, but now I'm glad I forgot!

feldergast:
We should, however, probably make sure there that no one has any large outstanding changes they're working on.  Removing trailing whitespace can make it hard to merge code sometimes.  Though, I'm not too worried about it being a problem if we keep up with it regularly.

PR #595, opened 2021-03-18
Dot config graph

lpsmodsim:
Updated the --output-dot graph to include a --dot-verbosity flag which determines the amount of detail to output in the dot graph


feldergast (changes requested):
This pull request needs to be based off of a more recent branch from devel.  There are 109 commits from the point where this branch diverged.

PR #605, opened 2021-04-01
Fix issue with wrong reported simulated time in some corner cases

feldergast:
Fixed an issue where the wrong simulated time was reported in multirank runs when there were no links cut across ranks.

### DIFF
--- a/src/sst/core/exit.cc
+++ b/src/sst/core/exit.cc
@@ -144,6 +144,22 @@ Exit::execute()
 
 }
 
+SimTime_t Exit::computeEndTime()
+{
+#ifdef SST_CONFIG_HAVE_MPI
+    // Do an all_reduce to get the end_time
+    SimTime_t end_value;
+    if ( !single_rank ) {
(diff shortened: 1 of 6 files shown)


src/sst/core/exit.cc
```diff
@@ -164,17 +180,7 @@ void Exit::check()
...
-        SimTime_t end_value;
-        if ( !single_rank ) {
-            MPI_Allreduce( &end_time, &end_value, 1, MPI_UINT64_T, MPI_MAX, MPI_COMM_WORLD );
-            end_time = end_value;
-        }
-#endif
-        if (single_rank) {
-            endSimulation(end_time);
-        }
+        computeEndTime();
```
gvoskuilen, line 183: computeEndTime() returns a time, is the return value not actually needed?

feldergast: Not here, but it is used in simulation.cc for the case of multirank runs with no links crossing rank boundaries.

gvoskuilen (approved):

PR #613, opened 2021-04-15
Bug fixes and memory/exection time optimizations for large graphs.

feldergast:
- Fixed a big in objectComms where the serialization size was stored in a 32-bit int.
- Changed ConfigGraph so it stored pointers to ConfigComponents instead of the actual objects.
  - Made it so that SparseVectorMap could store pointers
- Added global parameters.  You can now specify a named global set of parameters.  When adding to Param objects, only a pointer to the set will be kept.

### DIFF
--- a/src/sst/core/params.h
+++ b/src/sst/core/params.h
@@ -191,6 +191,8 @@ public:
      * Enable or disable parameter verification on an instance
      * of Params.  Useful when generating a new set of Params to
      * pass off to a module.
+     *
+     * @return returns the previous state of the flay
      */
     bool enableVerify(bool enable) { bool old = verify_enabled; verify_enabled = enable; return old; }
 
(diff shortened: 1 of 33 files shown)


src/sst/core/params.h
```diff
@@ -191,6 +191,8 @@ class Params : public SST::Core::Serialization::serializable {
      * Enable or disable parameter verification on an instance
      * of Params.  Useful when generating a new set of Params to
      * pass off to a module.
+     *
+     * @return returns the previous state of the flay
```
gvoskuilen, line 195: flay -> flag

nmhamster:
@feldergast - does this fix the serialization issues we have seen recently?

feldergast:
Yes, it does fix the serialization issue.

nmhamster:
> Yes, it does fix the serialization issue.

Great!

PR #630, opened 2021-05-04
Added the ability to loopback a link to the same port

feldergast:
Also, added a set of tests for links in the core tests.  Added versions of link configuration functions that take UnitAlgebra as the timebase.


src/sst/core/baseComponent.h
```diff
@@ -168,6 +168,13 @@ class BaseComponent {
      * @return A pointer to the configured link, or nullptr if an error occured.
      */
     Link* configureLink( const std::string& name, const std::string& time_base, Event::HandlerBase* handler = nullptr);
+    /** Configure a Link
+     * @param name - Port Name on which the link to configure is attached.
+     * @param time_base - Time Base of the link as a string
```
gvoskuilen, line 173: "as a string" -> "as a UnitAlgebra"

gvoskuilen (approved):

PR #631, opened 2021-05-10
Removes Zoltan from SST Core

nmhamster:
Removes Zoltan from SST Core. This removes:

- Configure/detect of Zoltan during `./configure` runs
- Removes Zoltan use in the partitionner code
- Removes Zoltan print out in configure eplilogue print outs


feldergast:
Superseded by #815

PR #639, opened 2021-05-13
Added a log_testing_note feature and added colorized result outputs

allevin:
Added a log_testing_note() feature that allows the developer to add information found during testing that may be of intrest/concern to the user.  These notes have no effect on Pass/Fail results.  All notes are displayed at the end of the run, and will also be displayed inline if debug mode is enabled.  

Also added colorization for the "TESTING PASSED" and "TESTING FAILED" indicators and the Notes section at the end of tests report to make them stand out a bit more.


src/sst/core/testingframework/sst_unittest_support.py
```diff
@@ -778,6 +778,27 @@ def log_fatal(errstr):
     log_forced(finalstr)
     sys.exit(2)
 
+###
+
+def log_testing_note(note_str):
+    """ Log a 'FATAL:' message.
```
feldergast, line 784: This 'FATAL:' contradicts the "NOTE : " below.  Cut and paste leftover?

allevin: Yup, copy/paste error.  Fixed.

gvoskuilen (approved):

Issue #625, opened 2021-04-28
SSTRandom should be named Random


nicmcd:
https://github.com/sstsimulator/sst-core/blob/9ee1db28c6c45129f685179e086739b9d3df83ea/src/sst/core/rng/sstrng.h#L27

SST::RNG::SSTRandom is a redundant name. Including "SST" in the class name is out of place and only done for this class. SST::RNG::Random would be more proper.

PR #644, opened 2021-05-18
Change over SST::RNG::SSTRandom to SST::RNG::Random

nmhamster:
Requested in Issue #625. Does not change the SST API as `SST::RNG::SSTRandom` is still available and will allow team to decide on deprecation in the future.


src/sst/core/rng/rng.cc
```diff
@@ -9,7 +9,6 @@
 // information, see the LICENSE file in the top level directory of the
 // distribution.
 
-
 #include "sst_config.h"
+#include "rng.h"
 #include "sstrng.h"
```
feldergast, line 14: Do we still need to include sstrng.h in rng.cc?

src/sst/core/rng/rng.h
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+        Generates the next random number as a signed 64-bit integer.
+    */
+    virtual int64_t     generateNextInt64() = 0;
+
+    /**
+        Generates the next random number as a signed 32-bit integer
+    */
+        virtual int32_t  generateNextInt32() = 0;
```
feldergast, line 53: Bad indentation

feldergast (changes requested):

src/sst/core/rng/rng.cc
nmhamster: Yes. We are moving to `rng.h` but current API will refer to `sstrng.h` so we are providing both to be compatible. We can decide how to move forward with class deprecation over time using this approach.

src/sst/core/rng/rng.h
nmhamster: I understand you will be working on formatting for code moving forward. Have you determined which indentation is actually correct now?

feldergast: Indentation will stay at 4 spaces

src/sst/core/rng/rng.cc
feldergast: Actually, a closer look shows that this file (rng.cc) isn't strictly needed at all because the rng.h header file is included in some other .cc files and there are no additional implementation details here (confirmed by a quick build without the file).  However, we can make a decision on whether to keep it or not later.

nmhamster: The point it to include both so we can remove `sstrng.h` once deprecation is decided.

src/sst/core/rng/marsaglia.h
```diff
@@ -16,7 +16,7 @@
 #include <stdint.h>
 #include <sys/time.h>
 
-#include "sstrng.h"
+#include "rng.h"
```
feldergast, line 19: This change (and the corresponding change in mersenne.h) could break any code that doesn't explicitly #include sstrng.h.  If, for example, a user only included mersenne.h, but refers to SSTRandom, the code won't compile because it won't find the class SSTRandom (which is defined in sstrng.h).  I don't believe this affects anything in the elements repository, but could affect third party element libraries.  Though, many people would argue that not including sstrng.h when referring to SSTRandom, which is defined there, is bad coding practice.

nmhamster: If we are to handle bad coding practices in our pull requests this is going to become quite limiting.

feldergast: Agreed.  I'm happy to leave this the way it is.  Just wanted it noted here in case it comes up later.

nmhamster: I am moving to include `sstrng.h` in the random generators to address bad coding practices.

feldergast (approved):

PR #646, opened 2021-05-18
Migrates SST Random Distributions to new names without the SST Prefix

nmhamster:
Renames distributions in the SST RNG core, removing SST prefixes from names. This has backwards compatibility until deprecation decisions are made. Addresses concerns identified in Issue #625.

### DIFF
--- a/src/sst/core/rng/marsaglia.cc
+++ b/src/sst/core/rng/marsaglia.cc
@@ -11,6 +11,7 @@
 
 #include "sst_config.h"
 
+#include "rng.h"
 #include "marsaglia.h"
 
 #include <cstdlib>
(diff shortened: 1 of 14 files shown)


feldergast (approved):
That was fun with all the formatting changes thrown in, but looks good.

src/sst/core/rng/marsaglia.cc
```diff
@@ -11,6 +11,7 @@
 
 #include "sst_config.h"
 
+#include "rng.h"
```
gvoskuilen, line 14: This is already included in marsaglia.cc, did you want it in both?

nmhamster: We are moving to an "include what you use" model, so starting to migrate over to including files where we use the classes to meet this future requirement. The header guards should ensure we do not duplicate etc.

PR #654, opened 2021-06-24
Update build to work with autoconf 2.71

gvoskuilen:


feldergast (commented):
Changes look reasonable, though I'm taking most of it on faith.  I assume this breaks some older versions of autoconf?  What versions will now be supported?

gvoskuilen:
@feldergast It *shouldn't* break prior versions as I added a check for the autoconf version for things that aren't backwards compatible with older autoconf versions. I'd prefer not to have version checking, but I didn't see a way around it for this. Nominally we support autoconf 2.59 and newer, in practice we're running 2.69 and now 2.71. 2.59 was released in 2003 so for SST 12 we should revisit whether it makes sense to keep compatibility that far back.

feldergast:
Even 2.69 was released way back in 2012.  I agree we should look at supported versions for SST 12.

feldergast (approved):

Issue #629, opened 2021-04-30
Define bounds of SSTRandom nextUniform


nicmcd:
https://github.com/sstsimulator/sst-core/blob/9ee1db28c6c45129f685179e086739b9d3df83ea/src/sst/core/rng/sstrng.h#L31

Please define whether or not 0 and 1 are included in the bounds of SST::SSTRandrom::nextUniform(). STL uses [0,1) by default which is very convent for selecting and random index in an array/vector:
```
size_t index = rng.nextUniform() * values.size();
```
Is this the case in SST? It is checked in the unit tests of the RNG implementations?

nmhamster:
See proposed changes in PR #657

nmhamster:
@nicmcd - confirmed, SST provided RNG operate in the range `[0,1)` as requested. Please reopen this issue if you experienced issues. Documentation is updated.

PR #657, opened 2021-07-06
Check that RNG never produce a value that equals 1.0

nmhamster:
This enforces that RNGs never produce a value that equals 1.0 during generation of a double. The range of values is [0,1) otherwise the code will loop and generate an alternative. This is requested in Issue #629 and is an extremely unlikely event.

### DIFF
--- a/src/sst/core/rng/marsaglia.cc
+++ b/src/sst/core/rng/marsaglia.cc
@@ -10,6 +10,7 @@
 // distribution.
 
 #include "sst_config.h"
+#include "sst/core/sst_types.h"
 
 #include "rng.h"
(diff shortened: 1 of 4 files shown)


gvoskuilen:
Looks good to me. We should also update the comment on nextUniform() in sstrng.h to say "[0, 1)" so it's clear from the API too. Do you want to add that to this PR?

nmhamster:
> Looks good to me. We should also update the comment on nextUniform() in sstrng.h to say "[0, 1)" so it's clear from the API too. Do you want to add that to this PR?

Updated the `rng.h` file as requested.

gvoskuilen (approved):

PR #658, opened 2021-07-09
Add debug and improved event handling support to StandardMem interface

gvoskuilen:


nmhamster (approved):

feldergast:
Can we hold off on merging this?  Just before vacation I made some changes to event that added a toString() function for better default debugging information for events.  The new default will know the name of any event that uses the serialization macros, instead of just using “generic event” as the default.  I don’t think this will conflict with that, but we may want to consolidate the naming conventions and make sure that it will all work together.  I’ll look back through the code today while I’m getting caught back up from vacation.

PR #662, opened 2021-07-20
Added inheritance for ELI information.

feldergast:
Also removed class ElementInfoPort2 because it was only there for backward compatibility with the old ELI format.

### DIFF
--- a/src/sst/core/component.h
+++ b/src/sst/core/component.h
@@ -97,8 +97,26 @@ protected:
 
 } //namespace SST
 
-#define SST_ELI_REGISTER_COMPONENT(cls,lib,name,version,desc,cat)   \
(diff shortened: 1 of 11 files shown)


gvoskuilen (commented):
Have you tried the "nullptr" to remove something inherited from the parent ELI functionality? It doesn't look like you've got a test of that piece. But the code looks right.

Otherwise, I'll approve.

feldergast:
Yes, several of the classes call SST_ELI_DELETE_* macros in coreTest_SubComponent.h.  A user could conceivably just add an entry with nullptr in the description field without the macro, which is not in any of the tests.  The documentation I'm working on will say not to do that, however.

gvoskuilen (approved):

PR #665, opened 2021-07-22
Core clang format

feldergast:


src/sst/core/cfgoutput/pythonConfigOutput.h
```diff
@@ -10,46 +10,47 @@
...
+    virtual void generate(const Config* cfg, ConfigGraph* graph) override;
 
 protected:
     ConfigGraph* getGraph() { return graph; }
-    void generateParams( const Params &params );
-    void generateCommonComponent( const char* objName, const ConfigComponent* comp);
-    void generateSubComponent( const char* owner, const ConfigComponent* comp );
-    void generateComponent( const ConfigComponent* comp );
-    void generateStatGroup(const ConfigGraph* graph, const ConfigStatGroup &grp);
+    void         generateParams(const Params& params);
```
nmhamster, line 33: This looks kind of wierd, especially since the variables then don't line up. Is this an actual standard we are going to stick to?

src/sst/core/bootshared.cc
```diff
@@ -121,23 +124,30 @@ void boot_sst_executable(const char* binary, const int verbose, char* argv[], co
             break;
 
         case ENAMETOOLONG:
-            fprintf(stderr, "Unable to launch SST, the path to the executable exceeds the operating system maximum. Try installing to a shorter path.\n");
+            fprintf(
+                stderr, "Unable to launch SST, the path to the executable exceeds the operating system maximum. Try "
```
nmhamster, line 128: Why are there new lines here between the function and `stderr`?

src/sst/core/cfgoutput/pythonConfigOutput.h
feldergast: It's set to line up lines that are not separated by spaces after their types.  So, since one of the functions returns a ConfigGraph*, all the ones that return void are aligned.  It's not my favorite, but it looks okay in the source file.  The flag is really turned on for variables, but the same flag seems to cover functions as well.

src/sst/core/bootshared.cc
feldergast: The flag is set to put in a newline after the paren if the function call won't all fit on one line.  This lines up all the parameters better for long function names (i.e. all the variables aren't shoved way to the right).  There are other ways to set these flags, but this produced the most readable result.

feldergast:
Just a general comment.  When configuring clang-format, I leaned toward readability over compact (or in some cases aesthetically pleasing) code.  There are other ways to balance this, but flexibility is limited and I had to set things up in ways I may have preferred not to in order to get things I cared about more.  I'm afraid if we use clang-format to test compliance, we're stuck with what it can produce.  We can still tweak things, but I played around with a lot of options and kept the ones that gave me the closest to what we talked about as coding standards.

src/sst/core/activity.h
```diff
@@ -238,34 +258,33 @@ class Activity : public SST::Core::Serialization::serializable {
          * 2) Alloc item from pool
          * 3) Append PoolID to item, increment pointer
          */
-        Core::MemPool *pool = nullptr;
-        size_t nPools = memPools.size();
-        std::thread::id tid = std::this_thread::get_id();
-        for ( size_t i = 0 ; i < nPools ; i++ ) {
-            PoolInfo_t &p = memPools[i];
+        Core::MemPool*  pool   = nullptr;
+        size_t          nPools = memPools.size();
```
calewis, line 262: Aligning `=`? Up you you, but it seems pretty weird IMO.

calewis (dismissed):
Assuming the tests all pass LGTM.

calewis:
@feldergast it looks like you need a `#include <cstring>` in mempool.h

feldergast:
@calewis Thanks.  I was just about to sit down and see if I could track down the issues.  The reordering of headers caused problems in several places on Mac builds, so it doesn't surprise me that linux had a few more.

calewis (dismissed):
LGTM, word of warning GCC 11 is probably going to cause more of this.  See https://www.gnu.org/software/gcc/gcc-11/porting_to.html#header-dep-change

nmhamster (approved):

PR #667, opened 2021-07-27
Changed the max number of empty lines to 2 in .clang-format.

feldergast:
Changing max number of empty lines to 2.  Also, this is the first PR to test the new clang-format test as part of the autotester.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -107,7 +107,7 @@ JavaScriptWrapImports: true
 KeepEmptyLinesAtTheStartOfBlocks: true
 MacroBlockBegin: ''
 MacroBlockEnd:   ''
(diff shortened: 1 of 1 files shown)


calewis (commented):
Shouldn't this go back in time and be applied to the code before the first round of clang-format? Otherwise this is only going to apply going forward and all previous double empty lines will never again appear.

feldergast:
@calewis It would probably be best to go back, but I had to clean up so much by hand after running clang-format the first time that I don't plan to do that.  I missed that option the first time around, so this change will just allow us an extra line to separate code regions in the future.

calewis (approved):

feldergast:
This change got added to another pull request.

Issue #664, opened 2021-07-22
memory ownership in objectComms.h


calewis:
In objectComms.h there are several `char* buffer = new char[size];` calls.  It is unclear if these are ever deleted? If they are deleted then `deserialize(std::vector<char> &buffer)` in the same file probably has a double delete.  I'd suggest that we switch away from using new if at all possible, but my main goal is just to figure out what is going on.

feldergast:
I'll take a look.  There clearly isn't a double delete in any path we use today or we would have seen it by now, but I'll track down what's happening.

calewis:
@feldergast any updates? I can look into this if needed.

PR #669, opened 2021-07-28
Communication leaks

calewis:
Attempt to address #664 I did not exhaustively prove that the serializer doesn't clean up these pointers, hopefully our tests would fail with the double free if that is the case.


feldergast (approved):
Looks good.

PR #670, opened 2021-07-28
Readme update

calewis:
Added minimal installation info to the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,49 @@ The Structural Simulation Toolkit (SST) was developed to explore innovations in 
 
 ---
 
+
+### Getting Started
+
+#### Building From Source
+
+##### Centos/RHEL 7
+
+```sh
+sudo yum install gcc gcc-c++ python3 python3-devel make automake git libtool libtool-ltdl-devel openmpi openmpi-devel
+mkdir sst-core && cd sst-core
+git clone https://github.com/sstsimulator/sst-core.git sst-core-src
+(cd sst-core-src && ./autogen.sh)
+mkdir build && cd build
+../sst-core-src/configure \
+  MPICC=/usr/lib64/openmpi/bin/mpicc \
+  MPICXX=/usr/lib64/openmpi/bin/mpic++ \
+  --prefix=$PWD/../sst-core-install
+make install 
+```
+
+##### Ubuntu 20.04
+
+```sh
+DEBIAN_FRONTEND=noninteractive sudo apt install openmpi-bin openmpi-common libtool libtool-bin autoconf python3 python3-dev automake build-essential git 
+mkdir sst-core && cd sst-core
+git clone https://github.com/sstsimulator/sst-core.git sst-core-src
+(cd sst-core-src && ./autogen.sh)
+mkdir build && cd build
+../sst-core-src/configure --prefix=$PWD/../sst-core-install
+make install 
+```
+
+#### Testing Your Install
+
+``` sh
+/path/to/sst-core/install/bin/sst-test-core
+```
+
+
+---
+
 Visit [sst-simulator.org](http://sst-simulator.org) to learn more about SST.
 
 See [Contributing](https://github.com/sstsimulator/sst-core/blob/devel/CONTRIBUTING.md) to learn how to contribute to SST.
 
-##### [LICENSE](https://github.com/sstsimulator/sst-core/blob/devel/LICENSE)
+See [LICENSE](https://github.com/sstsimulator/sst-core/blob/devel/LICENSE) for our license


README.md
```diff
@@ -1,17 +1,40 @@
-![SST](http://sst-simulator.org/img/sst-logo-small.png)
+![SST](http:g/sst-simulator.org/img/sst-logo-small.png)
```
nmhamster, line 1: What is the `:g/` here - shouldn't this be `//`

README.md
```diff
@@ -1,17 +1,40 @@
...
+### Copyright (c) 2009-2021, National Technology and Engineering Solutions of Sandia, LLC (NTESS)
 
 ---
 
 The Structural Simulation Toolkit (SST) was developed to explore innovations in highly concurrent systems where the ISA, microarchitecture, and memory interact with the programming model and communications system. The package provides two novel capabilities. The first is a fully modular design that enables extensive exploration of an individual system parameter without the need for intrusive changes to the simulator. The second is a parallel simulation environment based on MPI. This provides a high level of performance and the ability to look at large systems. The framework has been successfully used to model concepts ranging from processing in memory to conventional processors connected by conventional network interfaces and running MPI.
 
 ---
 
-Visit [sst-simulator.org](http://sst-simulator.org) to learn more about SST.
+Visit [sst-simulator.org](http:g/sst-simulator.org) to learn more about SST.
```
nmhamster, line 13: Same here `:g/`

README.md
```diff
@@ -1,17 +1,40 @@
...
 
-See [Contributing](https://github.com/sstsimulator/sst-core/blob/devel/CONTRIBUTING.md) to learn how to contribute to SST.
+See [Contributing](https:g/github.com/sstsimulator/sst-core/blob/devel/CONTRIBUTING.md) to learn how to contribute to SST.
```
nmhamster, line 15: Same here `:g/`

README.md
```diff
@@ -1,17 +1,40 @@
...
+#### Building SST-Core from the repo 
+
+```sh
+git clone https:g/github.com/sstsimulator/sst-core.git
+cd sst-core 
+.gautogen.sh
+mkdir build
+cd build
+# Don't disable MPI if you have one you want to use
+..gconfigure --disable-mpi --prefix=$PWD/../install
```
nmhamster, line 36: Same here `..gconfigure`

README.md
```diff
@@ -1,17 +1,40 @@
...
+```sh
+yum install gcc gcc-c++ python python-devel make automake git libtool libtool-ltdl-devel
+```
+
+#### Building SST-Core from the repo 
+
+```sh
+git clone https:g/github.com/sstsimulator/sst-core.git
+cd sst-core 
+.gautogen.sh
```
nmhamster, line 32: What is `.gautogen.sh`?

nmhamster (changes requested):
Seems to be some issues with replaced characters?

README.md
calewis: That apparently is me mistyping a find replace command 🤦‍♂️

README.md
calewis: Apparently at some point I replaced `/` with `g` oops 🤦‍♂️

README.md
```diff
@@ -14,4 +14,27 @@ Visit [sst-simulator.org](http://sst-simulator.org) to learn more about SST.
...
+#### Building SST-Core from the repo 
+
+```sh
+git clone https://github.com/sstsimulator/sst-core.git
+cd sst-core 
+./autogen.sh
+mkdir build
+cd build
+# Don't disable MPI if you have one you want to use
+../configure --disable-mpi --prefix=$PWD/../install
```
nmhamster, line 36: Why is disabling MPI in the default instructions - we should mostly want people to build with MPI right? This is the fastest current method for making the PDES parallel/performant?

nmhamster (changes requested):

README.md
calewis: Centos7 lets you yum install (I only tested this in the docker image) one of the following 
- openmpi 1.10.7
- openmpi 3.1.3
- mvapich 2.3
- mvapich 2.0a
- mpich 3.0
- mpich 3.2

Which one do you suggest that we tell our users to install?

feldergast: We only test against openmpi 4.0.5 currently

calewis: Exactly :), so if we want a minimal getting started install that only depends on what comes in the default repos it needs to not use MPI.  

@nmhamster, I don't think every example we show users how to build (this PR, or a Dockerfile) absolutely must come with the kitchen sink and  batteries included.  We want people to test the code out and you don't need MPI to do that.

calewis: In my opinion we want at least some instructions that provide the fastest easiest time to get to hello world where by hello world I mean 
```
$ /path/to/sst/install/dir/bin/sst --help
```

nmhamster: @calewis - agreed on a number of the TPLs for elements etc, but I'd say MPI is pretty important for SST. As above, its the main mechanism by which we deliver a significant value add (parallelism in the DES). I'm not convinced it is "kitchen sink and batteries".

nmhamster: One possibly is to have two sub-sections - a build with MPI and a build without MPI (in that order because with MPI is I'm going argue preferred).

calewis: We should trust our users enough to realize that `../configure --disable-mpi --prefix=$PWD/../install` is not the only way to build sst or the most performant or the one that necessarily suits the needs of experts.  I even put a comment that if they have an mpi they want to use they should remove that flag.  

But our hey here is how you get to hello world instructions can't include building openmpi 4.0.5 from source.

calewis: Kokkos has this as the first thing they tell you to do to build Kokkos:
```
cmake $srcdir \
  -DCMAKE_CXX_COMPILER=$path_to_compiler \
  -DCMAKE_INSTALL_PREFIX=$path_to_install \
  -DKokkos_ENABLE_OPENMP=On \
  -DKokkos_ARCH_HSW=On \
  -DKokkos_ENABLE_HWLOC=On \
  -DKokkos_HWLOC_DIR=$path_to_hwloc
  ```
  Is OPENMP the best choice 🤷‍♀️, it is the one most likely to work if you just want to get started.

nmhamster: I would not claim this is the most reduced version of the equivalent even for Kokkos. OpenMP is the most sensible parallel programming model to get started, but enable Haswell and HWLOC are not required for a basic setup. The issue is users building successfully but having a poor performance experience - Kokkos has unfortunately hit this a few times and its a hassle because then users (who don't know what they are doing) write this in slides, papers and talks.

calewis: We can pick one of those already available MPIs it just won't be tested.

feldergast: FWIW, for most simulations on-node MPI and threads provide almost identical performance.  However, I do agree that we want to push people toward MPI.  You could just change it the have the example leave out --disable-mpi and add a comment that if there is no MPI in your path add --disable-mpi.  It's a subtle difference, but does at least suggest that building with MPI is preferred.

calewis: Okay but I want this to work if you copy paste it.

calewis: @feldergast @nmhamster just pinging on how you would like to proceed.  We can either install an older mpi from yum or keep the `--disable-mpi` flag.  I want this code to be directly runnable so it's one of those options.

nmhamster:
> Before we remove the WIP label, could you squash all the commits to one and repush. We don't need the commit history for all these tiny changes.

We perform squash merges once the PR is done. Is this needed by the submitter?

feldergast:
Not necessarily, but Jeremy talked me into doing them by hand so that you could better control the final commit message.  I've always preferred to clean things up on my own rather than rely on GitHub doing it automatically.  And, I don't generally use squash merges for my PRs because I've cleaned things up to the state I want them merged in.  We can just squash merge this if you want, but I'd prefer to get rid of all the incremental comments that are there.

calewis:
But this makes it harder for me to achieve my goal of using up all the 7 character SHAs.

README.md
calewis: Marking resolved since we can pass the tests with the system MPI

nmhamster (approved):

Issue #634, opened 2021-05-12
Fix ill-formed JSON output


nicmcd:
In reference to this line of code: https://github.com/sstsimulator/sst-core/blob/4216b05a431c862c2d9382a28d2998b6cabb3cd8/src/sst/core/statapi/statoutputjson.cc#L93

It produces ill-formed JSON standard output:
```
{
   "rank" : 1

   "components" : [
      {
   <truncated here>
```
As shown, there is no comma after `"rank" : 1` however the JSON standard demands this. Since this code always prints the components, it should be safe to simply always add the comma.

nicmcd:
I've used both [nlohmann::json](https://github.com/nlohmann/json) quite extensively. I highly suggest using this instead of writing your own JSON output as they've spent a lot of time time making it easy to use (works like an STL container) and is JSON compliant (RFC 8259) which means that it will work with any standardized JSON library.

nmhamster:
@nicmcd - should have the "," added in #635. Let us know if that fixes this specific issue for you.

nicmcd:
Without testing it, I'm confident that this will fix this specific issue. Can you add a unit test for this? I suggest using a script that reads the output JSON using python, which is fairly pedantic in its formatting requirements:

python3 -c "import json; fd = open('<filename>'); json.load(fd)"

nmhamster:
@allevin - how would we add a test like this into the new test framework (see comment above)?

allevin:
> @allevin - how would we add a test like this into the new test framework (see comment above)?

@nmhamster A test for JSON would be added to the core tests.  We need a core test element that will can be loaded via an an SDL input file that when runs generates the JSON you want to check.  We would then just run @nicmcd recommendation to load the JSON formatted statistic file to verify its well formatted, and I would also recommend that we test it against a golden copy of the statistic file to see if it ever changes...  I just need someone to give me the core test element and SDL file.  I can do the rest.

feldergast:
There's already a StatisticsComponent element in the testElements directory and a test file in the tests directory.  We have not yet enable the test in a test suite, but we could modify the associated python file to take the type of output as an argument.  We could then create all the different output types with the same python file.  Looks like the StatisticsComponent code needs a little work, but should be easy to get ready.  At some point, we probably want to extend the test to use stats in SubComponents as well.

We could actually use the json parser as part of the test, but it would be just as reasonable to check the reference file using the json parser and then just compare to the reference file.

allevin:
[...]

PR #672, opened 2021-07-28
Start using Json Library 

calewis:
I have included the single header from https://github.com/nlohmann/json

It is MIT licensed so I think that is fine, but @nmhamster and others should confirm that. 

This is partly in response to issue #634.  @nicmcd I assume this is approximately what you had in mind? 

 I would mark WIP, but I want to run the test suite against it to ensure that header compiles on all the test setups.  

There need to be some tests and I'm sure y'all will find other things that need fixing, but I think this will be useful step. 

Plus who doesn't like negative lines of code (if you don't count dropping in the header).


src/sst/core/Makefile.am
```diff
@@ -116,6 +116,7 @@ nobase_dist_sst_HEADERS = \
 	interprocess/circularBuffer.h \
 	interprocess/sstmutex.h \
 	interprocess/ipctunnel.h \
+	nlohmann/json.hpp \
```
nmhamster, line 119: I think we should have a ```contrib/nlohmann` directory - a place where we can put any contributed source code (seeing as we may have others we add over time)?

nmhamster (changes requested):

src/sst/core/Makefile.am
feldergast: Agreed.  Or "tpl" or some such.  We should also move tinyxml into that directory.

calewis: @feldergast @nmhamster I just followed along with how the tinyxml directory was done.  Some projects do this in a top-level `/external` directory for things that aren't part of the project. This would also make it easy to not run the format tester on the files.

nmhamster: Understood regarding `tinyxml`, we just need to get both of these things out into a `contrib` somewhere but I'm not wedded strongly to specifically where that goes as long as we do it.

calewis:
I'm looking at `statoutputjson.cc` and I don't think it will be hard to port to a similar thing as the json output.  But, I assume that the current stat output is constantly writing results as it goes.  If we switch to storing those results and then only writing the file at the end of the simulation this will change the behavior.

nmhamster:
> I'm looking at `statoutputjson.cc` and I don't think it will be hard to port to a similar thing as the json output. But, I assume that the current stat output is constantly writing results as it goes. If we switch to storing those results and then only writing the file at the end of the simulation this will change the behavior.

SST allows periodic dumps of stats during a run - i.e. we have to write out at specific points.

feldergast:
> > I'm looking at `statoutputjson.cc` and I don't think it will be hard to port to a similar thing as the json output. But, I assume that the current stat output is constantly writing results as it goes. If we switch to storing those results and then only writing the file at the end of the simulation this will change the behavior.
> 
> SST allows periodic dumps of stats during a run - i.e. we have to write out at specific points.

And we definitely don't want to hold all that state in memory until the end of the run.  We need to be able to write out the stats and "start fresh" until the next output point.  Oh, and we still need to dump it to the same file.

calewis:
> SST allows periodic dumps of stats during a run - i.e. we have to write out at specific points.

Seems tricky, I suppose we could dump the current state of the stat at each point, completely overwriting the file.  I don't know that there is a nice way to periodically append.  We can look into it. 

What does the way this thing gets called look like? Is it:
```
startOfSimulation();
// Some time passes
while(!end){
   if(heartbeat % period == 0){
     // write a bunch of stuff
   } 
 }
 endOfSimulation();
 }
 ```
 We can probably make that work somehow if that is the expected calling sequence (obviously I'm going for intent here, not how it literally gets called).

**Comment by nmhamster** [comment]
> > SST allows periodic dumps of stats during a run - i.e. we have to write out at specific points.
> 
> Seems tricky, I suppose we could dump the current state of the stat at each point, completely overwriting the file. I don't know that there is a nice way to periodically append. We can look into it.
> 
> What does the way this thing gets called look like? Is it:
> 
> ```
> startOfSimulation();
> // Some time passes
> while(!end){
>    if(heartbeat % period == 0){
>      // write a bunch of stuff
>    } 
>  }
>  endOfSimulation();
>  }
> ```
> 
> We can probably make that work somehow if that is the expected calling sequence (obviously I'm going for intent here, not how it literally gets called).

I think you should just write out a sequence of files e.g. mystats-0.json, mystats-1.json

**Comment by feldergast** [comment]
@calewis That's essentially correct.  In reality, an Event gets put in the TimeVortex at the dump interval.  Also, different stats may be on different intervals, so we may not even be writing all stats at each interval.

**Comment by calewis** [comment]
Do the numbers correspond to output events.

**Comment by nmhamster** [comment]
> Do the numbers correspond to output events.

No monotonically increasing dump number.

**Comment by calewis** [comment]
@feldergast does `implStartOutputEntries` and `implStopOutputEntires` get called every time the stat outputs or only once?

**Comment by feldergast** [comment]
> @feldergast do `implStartOutputEntries` and `implStopOutputEntires` get called every time the stat outputs or only once?

Not sure, I didn't write that code and really haven't done much in that part of the codebase, so will have to dig into it a bit more.

**Comment by nmhamster** [comment]
> > @feldergast do `implStartOutputEntries` and `implStopOutputEntires` get called every time the stat outputs or only once?
> 
> Not sure, I didn't write that code and really haven't done much in that part of the codebase, so will have to dig into it a bit more.

I think they are supposed to get called immediately prior to the dump being generated, each time we generate a dump. This was to allow us to do things like start a new set of statistics with an attached simulated time in outputs like HDF5 that allow for time-series based data.

**Comment by nmhamster** [comment]
@calewis / @feldergast - putting this into WIP for now. Major thing we need to sort out is where the `contrib` code is going to go. I think we had discussed adding a test as well. Main reason for WIP is to prevent it clogging up the autotester.

**Comment by calewis** [comment]
Yeah, I think the proper way to do that is we start with a new PR.  In that PR we should make 
/contrib and then we should put tinyxml in there.  Then iff autogen installs libltdl we should put that in /contrib as well IMO.  

Once we have that PR working we can rebase this pr on top of it and put the json.hpp file in there.

**Comment by feldergast** [comment]
> Yeah, I think the proper way to do that is we start with a new PR. In that PR we should make
> /contrib and then we should put tinyxml in there. Then iff autogen installs libltdl we should put that in /contrib as well IMO.

I agree, this is probably the right way to go.  However, I wouldn't use "contrib" as the name as this seems to imply only a loose connection, whereas these file are relatively integral to the code.  I would use something like "external" or "tpl" or some such.

**Comment by nicmcd** [comment]
I know I'm late to the party but wouldn't it be more proper to have a "--with-nlohmann-json" autotools flag and link to a separately installed library? This would also be very easy with spack since nlohmann-json is already known to spack. In general, multiple repository projects are the future! Importing all third party code into your repository is just a recipe for disaster.

**Comment by calewis** [comment]
@nicmcd We've made a goal over time to not require more external dependencies than we have to.  Your suggestion is reasonable, but we may have people using this who aren't allowed to use spack or brew.  nlohmann-json appears to be designed with the goal of letting you just copy the single-include header to your project.  Since we don't expect users to use this header for now (not planning on installing it) and we won't include it in any of our headers, I think we just copy it to /external until other issues change. 

Potentially we will switch to not packaging it in the future so don't count that out.

**Comment by feldergast** [comment]
@nicmcd In general, I agree that we don't want to import all third party software into the repository.  However for the case of a single header file, pulling it in directly is very easy and provides a couple advantages:  First, it's one less dependency people have to go off and find (not everyone will be using spack or similar for dependancies); in general we try to keep the dependencies on external software in core a low as possible.  JSON support is pretty integral to some future features, so it won't really by "optional" for a lot of users.  Second, it locks the version so we don't have to continuously track and test different versions.  We can move to new versions as we see fit.

**Comment by nmhamster** [comment]
@calewis - are you refactoring this so that the JSON headers are located in external etc?

**Comment by calewis** [comment]
As soon as the other PR hits devel yes.

**Comment by calewis** [comment]
Retesting this, @nmhamster, @feldergast  the json headers are now in external.  They should get packaged with dist, but not installed. I think I fixed #655.  And we are going to wait to test until the json input format has been decided since we need to make sure that anything we output must also be reingestible. 

Aside from trying to fix #655 this is very close to a 1:1 port of the old config outputter where I just turned the printf statements into inserting into the json type, since we don't currently have tests I don't know if this breaks anything or not.

**Review by nmhamster — APPROVED** [review]

File: src/sst/core/cfgoutput/jsonConfigOutput.cc [hunk 2]
```diff
@@ -18,128 +18,113 @@
...
-                       + "\t\t\t\"right\" : \"" + compMap[link.component[1]]->name + "\",\n"
-                       + "\t\t\t\"rightPort\" : \"" + link.port[1] + "\",\n" + "\t\t\t\"latency\" : \""
-                       + link.latency_str[minLatIdx] + "\"\n";
-    fprintf(outputFile, "%s", temp.c_str());
+    auto const& link  = pair.link;
+    auto const* graph = pair.graph;
+
+    // These accesses into compMap are not checked
+    j              = json::json { { "name", link.name } };
+    j["left"]      = graph->findComponent(link.component[0])->name;
```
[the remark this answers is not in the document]
calewis: This was fixed in 78a0a7a

Issue #671, opened 2021-07-28
Clang format script isn't working


calewis:
Clang format script isn't working 

RHEL7

Last commit from upstream core is 98803c0026ae8c7d925dfd0e085ef102d8107077

I'm in the top level directory and I try to run 
```
$ scripts/clang-format-test.sh
Failed to execute process 'scripts/clang-format-test.sh'. Reason:
exec: Exec format error
The file 'scripts/clang-format-test.sh' is marked as an executable but could not be run by the operating system.
```

feldergast:
Looks like it's missing #!/bin/bash or #!/usr/bin/env bash

PR #673, opened 2021-07-28
Added a #!/bin/bash to the top of the clang format script.

calewis:
Attempted fix for #671 just FYI running the script generates failures on on current devel.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,7 +12,7 @@ sstconf_DATA = src/sst/sstsimulator.conf
 EXTRA_DIST = \
 	autogen.sh \
 	CONTRIBUTING.md \
-	scrips/clang-format-test.sh \
+	scripts/clang-format-test.sh \
 	INSTALL \
 	ISSUES \
 	LICENSE \
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,6 +12,7 @@ sstconf_DATA = src/sst/sstsimulator.conf
 EXTRA_DIST = \
 	autogen.sh \
 	CONTRIBUTING.md \
+	scrips/clang-format-test.sh \
 	INSTALL \
 	ISSUES \
 	LICENSE \
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 .dirstamp
 Makefile.in
 libltdl
+clang_format_results*
 Makefile
 src/sst/core/sst.x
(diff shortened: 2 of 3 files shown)


feldergast:
What platform are you running clang-format on?  And which version?  Devel runs clean using clang-format 12 on my Mac and on sst-devel (which is Linux, but I'm not sure what version).

calewis:
Okay this is going to be awkward I'm using clang-format13. I suspect that is the issue.

feldergast:
Yea, we were warned to pick a version and stick with it because they make non-backward compatible changes between versions.  We ended up going with version 12.

allevin (dismissed):
Looks good to me

feldergast:
Did clang-format-test.sh get added to the Makefile for make dist?  If not, we should probably add it.

allevin:
> Did clang-format-test.sh get added to the Makefile for make dist? If not, we should probably add it.

Good catch!   No I did not add it... @calewis can you add this to your fork?

calewis:
Sure just tell me where I need to add it to make it go.

feldergast:
It goes in the EXTRA_DIST section near the top of Makefile.am.

nmhamster:
@feldergast / @calewis - setting this to WIP to prevent autotester clogging. I think we are waiting on the addition of the script in `Makefile.am` right?

calewis:
@nmhamster yeah I just added it to the first `Makefile.am` above the dir it was in.  If it needs to go somewhere else just let me know.

nmhamster:
@calewis - I removed the WIP, let's see what the build says.

Makefile.am
```diff
@@ -12,6 +12,7 @@ sstconf_DATA = src/sst/sstsimulator.conf
 EXTRA_DIST = \
 	autogen.sh \
 	CONTRIBUTING.md \
+	scrips/clang-format-test.sh \
```
nmhamster, line 15: Should this be `scripts` not `scrips` ?

calewis: 🤷‍♀️ yeah probably

nmhamster: Have you developed love or hatred for the auto-tester yet ? :)

calewis: I just try to ignore it :)

nmhamster (approved):

PR #676, opened 2021-07-29
Refactor external code

calewis:
This PR just moves tinyxml to a top level external directory and then does minimal modification of the makefiles to try and make it compile.  I don't know anything about how autotools works so please make sure my makefile modifications make sense. 

This pr is needed by #672 so we don't put the json code deep in our source.


calewis:
WHY 

The .Po files don't get expanded on linux (not sure why this works on Mac) 😭
```
$ find . -name "*.Po" | rg --fixed-strings "\$(top_srcdir)"
./src/sst/core/$(top_srcdir)/external/tinyxml/.deps/tinyxml.Po
./src/sst/core/$(top_srcdir)/external/tinyxml/.deps/tinystr.Po
./src/sst/core/$(top_srcdir)/external/tinyxml/.deps/tinyxmlparser.Po
./src/sst/core/$(top_srcdir)/external/tinyxml/.deps/tinyxmlerror.Po
```
I'm working on a fix now.

calewis:
If I try to run the failing test from before on my Mac I just get:

```
 ./bin/sst  --model-options="--topo=torus --shape=16x16x16  --cmdLine=\"Init\" --cmdLine=\"AllPingPong iterations=10 messageSize=20000 \" --cmdLine=\"Fini\" " sst-elements/src/sst/elements/ember/test/emberLoad.py
  File "/sst-elements/src/sst/elements/ember/test/emberLoad.py", line 8, in <module>
    from loadInfo import *
ModuleNotFoundError: No module named 'loadInfo'
FATAL: SSTPythonModel: Execution of model construction function failed
SST Fatal Backtrace Information:
    0 : 0   sstsim.x                            0x0000000102944cdf _ZNK3SST6Output5fatalEjPKcS2_iS2_z + 1231
    1 : 1   sstsim.x                            0x0000000102c3d2ce _ZN3SST4Core24SSTPythonModelDefinition17createConfigGraphEv + 398
    2 : 2   sstsim.x                            0x000000010271c909 main + 1833
    3 : 3   libdyld.dylib                       0x00007fff6c415cc9 start + 1
--------------------------------------------------------------------------
```
Any idea where to start?

src/sst/core/Makefile.am
```diff
@@ -298,7 +299,7 @@ sstinfo_x_LDFLAGS = \
 	-export-dynamic \
 	$(SST_LTLIBS_ELEMLIBS)
 
-include tinyxml/Makefile.inc
+include ../../../external/tinyxml/Makefile.inc
```
nmhamster, line 302: I think you want to use $(top_srcdir) here, not relative in case someone does an out of source build.

external/tinyxml/Makefile.inc
```diff
@@ -0,0 +1,47 @@
...
+	../../../external/tinyxml/tinyxmlSTL.dsp \
+	../../../external/tinyxml/tinyxmlSTL.vcproj \
+	../../../external/tinyxml/tinyxml_lib.dsp \
+	../../../external/tinyxml/tinyxml_lib.vcproj \
+	../../../external/tinyxml/utf8test.gif \
+	../../../external/tinyxml/utf8test.xml \
+	../../../external/tinyxml/utf8testout.xml \
+	../../../external/tinyxml/utf8testverify.xml \
+	../../../external/tinyxml/xmltest.cpp \
+	../../../external/tinyxml/docs
```
nmhamster, line 42: I think these should use `$(top_srcdir)` or an equivalent so these are not relative. This will not work in out of source builds I think.

nmhamster (changes requested):

external/tinyxml/Makefile.inc
calewis: Check the errors I was having on linux.  The problem was the `.Po` were like
```
./src/sst/core/$(top_srcdir)/external/tinyxml/.deps/tinyxml.Po
```
which was breaking the build so I changed it to relative, which hacky fixed the build.  I agree with you except that it doesn't compile on every linux that I've tested (Centos 7 and RHEL 7) when doing that.

calewis:
@nmhamster based on our conversation.  I have left the `../../../external` relative paths in the makefiles.  And I removed `sstinfo.h` from the install since it was the only header that included `tinyxml.h`.  Hopefully testing will find any issues I missed.

nmhamster (approved):

src/sst/core/Makefile.am
```diff
@@ -15,7 +15,7 @@ AM_CPPFLAGS =  \
 
 DIST_SUBDIRS = libltdl .
 SUBDIRS = libltdl .
-EXTRA_DIST = mainpage.dox eli/README.md
+EXTRA_DIST = sstinfo.h mainpage.dox eli/README.md
```
feldergast, line 18: It may seem couterintuitive, but if you want a header file to be part of make dist, but not to be installed, you should add it to the "sst_core_sources" section of Makefile.am.

feldergast (changes requested):

nmhamster:
@calewis - it looks like merging is blocked right now because there are "change requested" pending (@feldergast).

feldergast (approved):

PR #677, opened 2021-07-29
Util functions

feldergast:
A couple bug fixes plus some extra util functions for better debug prints and getting timing information.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -107,7 +107,7 @@ JavaScriptWrapImports: true
 KeepEmptyLinesAtTheStartOfBlocks: true
 MacroBlockBegin: ''
 MacroBlockEnd:   ''
-MaxEmptyLinesToKeep: 1
+MaxEmptyLinesToKeep: 2
 NamespaceIndentation: None
 ObjCBinPackProtocolList: Auto
 ObjCBlockIndentWidth: 2
--- a/src/sst/core/simulation.cc
+++ b/src/sst/core/simulation.cc
@@ -188,7 +188,13 @@ Simulation_impl::Simulation_impl(Config* cfg, RankInfo my_rank, RankInfo num_ran
     endSimCycle(0),
     currentPriority(0),
     my_rank(my_rank),
-    num_ranks(num_ranks)
+    num_ranks(num_ranks),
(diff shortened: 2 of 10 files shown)


src/sst/core/simulation.cc
```diff
@@ -739,6 +754,36 @@ Simulation_impl::printStatus(bool fullStatus)
     }
 }
 
+double
+Simulation_impl::getRunPhaseElapsedRealTime() const
+{
+    if ( run_phase_start_time == 0.0 ) return 0.0; // Not in run phase yet
+    if ( run_phase_total_time == 0.0 ) { return sst_get_cpu_time() - run_phase_start_time; }
+    else {
```
calewis, line 762: Don't need this else you can just do:
```c++
Simulation_impl::getRunPhaseElapsedRealTime() const
 {
     if ( run_phase_start_time == 0.0 ) return 0.0; // Not in run phase yet
     if ( run_phase_total_time == 0.0 ) { return sst_get_cpu_time() - run_phase_start_time; }
     return run_phase_total_time; 
 }
 ```

Comment (calewis) [reply src/sst/core/simulation.cc:762]: The elses below aren't needed either.

**Review by nmhamster — APPROVED** [review]

File: .clang-format [hunk 2]
```diff
@@ -107,7 +107,7 @@ JavaScriptWrapImports: true
 KeepEmptyLinesAtTheStartOfBlocks: true
 MacroBlockBegin: ''
 MacroBlockEnd:   ''
-MaxEmptyLinesToKeep: 1
+MaxEmptyLinesToKeep: 2
```
nmhamster, line 110: Is this changing the clang-format as well in this PR?

calewis: This was in a different PR and it got closed. I don't think it actually changes any lines since we previously removed all gaps of 2 or more lines.  This just allows them to come back if you want to.

feldergast: Correct.  I intended that to be in the clang-format original, but somehow it got changed back and it would have been too painful to go back to the original code a reformat since there was a lot of changes I had to do by hand to make it compile after changing headers around.  This at least allows two spaces in the future to add a bit of separation between groups of functions, variables, etc.

PR #682, opened 2021-08-02
Check for execinfo.h

calewis:
`backtrace` doesn't work when using musl libc.  This may not be the most robust way to detect this, but on alpine linux at least if you don't install execinfo.h this will solve the problem.  If someone else wants to tell me the proper way to detect the backtrace function I can do that.


calewis:
I need to remove the commit about docker.  That shouldn't be there.

nmhamster (approved):

PR #683, opened 2021-08-02
Utilize Python Detection to Populate SST Core/Element Test Loaders

nmhamster:
Uses SST's Python detection scripts to populate `sst-test-core` and `sst-test-elements` scripts (which are installed in `bin`).

### DIFF
--- a/src/sst/core/testingframework/Makefile.inc
+++ b/src/sst/core/testingframework/Makefile.inc
@@ -40,6 +40,8 @@ endif
 EXTRA_DIST += \
 	testingframework/__init__.py \
 	testingframework/readme.md \
+	testingframework/sst-test-core.in \
+   testingframework/sst-test-elements.in \
 	testingframework/sst_test_engine_loader.py \
 	testingframework/sst_unittest.py \
 	testingframework/sst_unittest_support.py \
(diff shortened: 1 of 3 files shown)


calewis:
Why don't we just move the files that run the tests to the tests directory? Why do we put them down in the core src?

nmhamster:
I'm not sure we need to handle that in this specific PR do we, this effectively recreates current process and functionality.

src/sst/core/testingframework/Makefile.inc
```diff
@@ -40,6 +40,8 @@ endif
 EXTRA_DIST += \
 	testingframework/__init__.py \
 	testingframework/readme.md \
+	testingframework/sst-test-core.in \
+   testingframework/sst-test-elements.in \
```
calewis, line 44: Formating?

nmhamster: Addressed in SHA:35b705d

calewis (approved):

allevin:
@nmhamster @calewis; I didn't get a chance to deeply review this, but there is a problem with this change.  There is a core option called `--enable-testframework-dev` which is used only for test framework development.  Normally, when installed, the test frameworks files are copied to the bin and libexec directories.  However if developing features for the test frameworks, we dont want to copy the files over in the bin and libexec directories for every possible change/test.  So the `--enable-testframework-dev` makes symbolic links to the sst-test-core and sst-test-elements files (which no longer exist).  I did a clean build, and the symbolic links did not exist.

PR #690, opened 2021-08-06
Utilize basic dlopen functionality to work around needing libtldl

nmhamster:
Eliminates use of `lt_` functions for dynamic loading. This overhauls some basic capability and works with SST loading capabilities (at the cost of some generality).

### DIFF
--- a/src/sst/core/elemLoader.cc
+++ b/src/sst/core/elemLoader.cc
@@ -19,9 +19,9 @@
 #include "sst/core/sstpart.h"
 #include "sst/core/subcomponent.h"
 
+#include <climits>
 #include <cstdio>
 #include <cstring>
-#include <climits>
 #include <dirent.h>
 #include <vector>
 
@@ -76,12 +76,10 @@ splitPath(const std::string& searchPaths)
 ElemLoader::ElemLoader(const std::string& searchPaths) : searchPaths(searchPaths)
 {
     preload_symbols();
-	verbose = false;
+    verbose = false;
 
-	const char* verbose_env = getenv("SST_CORE_DL_VERBOSE");
-	if(nullptr != verbose_env) {
-		verbose = atoi(verbose_env) > 0;
-	}
+    const char* verbose_env = getenv("SST_CORE_DL_VERBOSE");
+    if ( nullptr != verbose_env ) { verbose = atoi(verbose_env) > 0; }
 }
 
 ElemLoader::~ElemLoader() {}
@@ -96,9 +94,7 @@ ElemLoader::loadLibrary(const std::string& elemlib, std::ostream& err_os)
     bool  found_element = false;
 
     for ( std::string& next_path : paths ) {
-		  if(verbose) {
-				printf("SST-DL: Searching: %s\n", next_path.c_str());
-		  }
+        if ( verbose ) { printf("SST-DL: Searching: %s\n", next_path.c_str()); }
 
         DIR* current_dir = opendir(next_path.c_str());
 
@@ -106,24 +102,18 @@ ElemLoader::loadLibrary(const std::string& elemlib, std::ostream& err_os)
             struct dirent* dir_file;
 
             while ( (dir_file = readdir(current_dir)) != nullptr ) {
-					  if(verbose) {
-							printf("SST-DL: Checking file types: %s\n", dir_file->d_name);
-		  			  }
+                if ( verbose ) { printf("SST-DL: Checking file types: %s\n", dir_file->d_name); }
 
                 if ( (dir_file->d_type | DT_REG) || (dir_file->d_type | DT_LNK) ) {
                     if ( !strncmp("lib", dir_file->d_name, 3) ) {
                         sprintf(full_path, "%s/%s", next_path.c_str(), dir_file->d_name);
-								if(verbose) {
-									printf("SST-DL: Attempting dynamic load of %s\n", full_path);
-								}
+                        if ( verbose ) { printf("SST-DL: Attempting dynamic load of %s\n", full_path); }
                         void* handle = dlopen(full_path, RTLD_NOW | RTLD_GLOBAL);
 
                         if ( nullptr != handle ) {
                             found_element = true;
 
-									 if(verbose) {
-										printf("SST-DL: Load of %s was successful.\n", full_path);
-									 }
+                            if ( verbose ) { printf("SST-DL: Load of %s was successful.\n", full_path); }
 
                             // loading a library can "wipe" previously loaded libraries depending
                             // on how weak symbol resolution works in dlopen
@@ -137,11 +127,10 @@ ElemLoader::loadLibrary(const std::string& elemlib, std::ostream& err_os)
                                     }
                                 }
                             }
-                        } else {
-									if(verbose) {
-										printf("SST-DL: Load of %s failed, %s\n", full_path, dlerror());
-									}
-								}
+                        }
+                        else {
+                            if ( verbose ) { printf("SST-DL: Load of %s failed, %s\n", full_path, dlerror()); }
+                        }
                     }
                 }
             }
--- a/src/sst/core/elemLoader.cc
+++ b/src/sst/core/elemLoader.cc
@@ -21,8 +21,8 @@
 
 #include <cstdio>
 #include <cstring>
+#include <climits>
 #include <dirent.h>
-#include <limits.h>
 #include <vector>
 
 #ifdef HAVE_DLFCN_H
@@ -76,6 +76,12 @@ splitPath(const std::string& searchPaths)
 ElemLoader::ElemLoader(const std::string& searchPaths) : searchPaths(searchPaths)
 {
     preload_symbols();
+	verbose = false;
+
+	const char* verbose_env = getenv("SST_CORE_DL_VERBOSE");
+	if(nullptr != verbose_env) {
+		verbose = atoi(verbose_env) > 0;
+	}
 }
 
 ElemLoader::~ElemLoader() {}
@@ -90,37 +96,52 @@ ElemLoader::loadLibrary(const std::string& elemlib, std::ostream& err_os)
     bool  found_element = false;
 
     for ( std::string& next_path : paths ) {
+		  if(verbose) {
+				printf("SST-DL: Searching: %s\n", next_path.c_str());
+		  }
+
         DIR* current_dir = opendir(next_path.c_str());
 
         if ( current_dir ) {
             struct dirent* dir_file;
 
             while ( (dir_file = readdir(current_dir)) != nullptr ) {
+					  if(verbose) {
+							printf("SST-DL: Checking file types: %s\n", dir_file->d_name);
+		  			  }
+
                 if ( (dir_file->d_type | DT_REG) || (dir_file->d_type | DT_LNK) ) {
                     if ( !strncmp("lib", dir_file->d_name, 3) ) {
                         sprintf(full_path, "%s/%s", next_path.c_str(), dir_file->d_name);
-                        void* handle = dlopen(full_path, RTLD_NOLOAD | RTLD_GLOBAL);
+								if(verbose) {
+									printf("SST-DL: Attempting dynamic load of %s\n", full_path);
+								}
+                        void* handle = dlopen(full_path, RTLD_NOW | RTLD_GLOBAL);
 
-                        if ( nullptr == handle ) {
-                            handle = dlopen(full_path, RTLD_NOW | RTLD_GLOBAL);
+                        if ( nullptr != handle ) {
+                            found_element = true;
 
-                            if ( nullptr != handle ) {
-                                found_element = true;
+									 if(verbose) {
+										printf("SST-DL: Load of %s was successful.\n", full_path);
+									 }
 
-                                // loading a library can "wipe" previously loaded libraries depending
-                                // on how weak symbol resolution works in dlopen
-                                // rerun the loaders to make sure everything is still registered
-                                for ( auto& libpair : ELI::LoadedLibraries::getLoaders() ) {
-                                    // loop all the elements in the element lib
-                                    for ( auto& elempair : libpair.second ) {
-                                        // loop all the loaders in the element
-                                        for ( auto* loader : elempair.second ) {
-                                            loader->load();
-                                        }
+                            // loading a library can "wipe" previously loaded libraries depending
+                            // on how weak symbol resolution works in dlopen
+                            // rerun the loaders to make sure everything is still registered
+                            for ( auto& libpair : ELI::LoadedLibraries::getLoaders() ) {
+                                // loop all the elements in the element lib
+                                for ( auto& elempair : libpair.second ) {
+                                    // loop all the loaders in the element
+                                    for ( auto* loader : elempair.second ) {
+                                        loader->load();
                                     }
                                 }
                             }
-                        }
+                        } else {
+									if(verbose) {
+										printf("SST-DL: Load of %s failed, %s\n", full_path, dlerror());
+									}
+								}
                     }
                 }
             }
--- a/src/sst/core/elemLoader.cc
+++ b/src/sst/core/elemLoader.cc
@@ -80,6 +80,15 @@ ElemLoader::ElemLoader(const std::string& searchPaths) : searchPaths(searchPaths
 
     const char* verbose_env = getenv("SST_CORE_DL_VERBOSE");
     if ( nullptr != verbose_env ) { verbose = atoi(verbose_env) > 0; }
+
+    const char* bind_env = getenv("SST_CORE_DL_BIND_POLICY");
+    if ( nullptr == bind_env ) { bindPolicy = RTLD_LAZY | RTLD_GLOBAL; }
+    else if ( (!strcmp(bind_env, "lazy")) || (!strcmp(bind_env, "LAZY")) ) {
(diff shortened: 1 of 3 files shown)


calewis:
I have a meta question.  I realized it would break backwards compatibility, but is there a reason we don't just have `sst-register` change the `DT_RUNPATH`s in the tools that call `dlopen`?

@nmhamster we talked about why you couldn't just use `dlopen` and all this was needed and the reasons you gave me were along the lines of swapping out two different versions of an element at run time.  I haven't checked that RHEL 7 ships with `chrpath`.  But if we just modify the runpath of the core binaries, that would let users get the element they registered by default and they could choose to swap in a different one by changing `LD_LIBRARY_PATH`.  

Thoughts? I guess this ties into the motivation for this extra code to assist `dlopen` in the first place.

src/sst/core/elemLoader.cc
```diff
@@ -57,68 +57,6 @@ preload_symbols(void)
```
calewis, line 57: Probably should take this opertunity to remove 
```cpp
// README: This is only set if we are not building any elements
// in the split up build this is now default so must always be called.
```
I assume this is a hold over from the monorepo days.

calewis: In fact if `preload_symbols` entire body is dead code lets just remove it?

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
-followError(
-    const std::string& libname, const std::string& elemlib, const std::string& searchPaths, std::ostream& err_os)
+ElemLoader::ElemLoader(const std::string& searchPaths) : searchPaths(searchPaths)
 {
-    std::string so_path = libname + ".so";
-    std::string fullpath;
-    void*       handle;
-
-    std::vector<std::string> paths = splitPath(searchPaths);
+    preload_symbols();
```
calewis, line 78: See above, can we just remove `preload_symbols` entirely?

src/sst/core/elemLoader.h
```diff
@@ -40,9 +35,17 @@ class ElemLoader
...
+     *
+     * @param potElems - vector of potential elements that could contain elements
+     * @return void
      */
-    std::vector<std::string> getPotentialElements();
+    void getPotentialElements(std::vector<std::string>& potElems);
+
+private:
+    std::string searchPaths;
+    bool        verbose;
```
calewis, line 47: Can set `bool verbose = false` here if you want it to always start off false.

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
+    verbose = false;
```
calewis, line 79: Can we just set verbose in the constructor if it's going to be hard coded here.  It can also be set to false in the header with `bool verbose = false;` to avoid needing to make the constructor noisy

src/sst/core/elemLoader.h
```diff
@@ -40,9 +35,17 @@ class ElemLoader
...
+    int         bindPolicy;
```
calewis, line 48: Set `int bindPolicy = RTLD_LAZY | RTLD_GLOBAL;` here or (in the constructor if you don't want to pull in the header with these symbols right here).

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
-        struct stat sbuf;
-        int         ret;
+    const char* verbose_env = getenv("SST_CORE_DL_VERBOSE");
+    if ( nullptr != verbose_env ) { verbose = atoi(verbose_env) > 0; }
 
-        fullpath = path + "/" + so_path;
-        ret      = stat(fullpath.c_str(), &sbuf);
-        if ( ret == 0 ) break;
+    const char* bind_env = getenv("SST_CORE_DL_BIND_POLICY");
+    if ( nullptr == bind_env ) { bindPolicy = RTLD_LAZY | RTLD_GLOBAL; }
```
calewis, line 85: Can skip this if and be more clear if you just construct `bindPolicy = RTLD_LAZY | RTLD_GLOBAL;` Then you only have the cases that `bind_env != nullptr` to contend with.

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
-        // which is totally useless...
-        // Make darn sure we have enough space to hold the error message
-        std::vector<char> err_str(1e6);
-        sprintf(err_str.data(), "Opening element library %s failed: %s\n", elemlib.c_str(), lt_dlerror());
-        err_os << (const char*)err_str.data();
-        followError(libname, elemlib, searchPaths, err_os);
-    }
+    std::vector<std::string> paths   = splitPath(searchPaths);
+
+    char* full_path     = new char[PATH_MAX];
```
calewis, line 101: Reason to prefer `sprintf` and `char *` over a string?

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
-    for ( auto& libpair : ELI::LoadedLibraries::getLoaders() ) {
-        // loop all the elements in the element lib
-        for ( auto& elempair : libpair.second ) {
-            // loop all the loaders in the element
-            for ( auto* loader : elempair.second ) {
-                loader->load();
+    for ( std::string& next_path : paths ) {
+        if ( verbose ) { printf("SST-DL: Searching: %s\n", next_path.c_str()); }
+
+        if ( next_path.at(next_path.size() - 1) == '/' ) {
```
calewis, line 107: `std::string` has a `.back()` method.

calewis: Also `.at()`  does bounds checking, but that is pessimistic here since `size() - 1` should be fine.

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
+        }
+        else {
+            if ( verbose ) { printf("SST-DL: Load was successful.\n"); }
+
+            found_element = true;
+
+            for ( auto& libpair : ELI::LoadedLibraries::getLoaders() ) {
+                // loop all the elements in the element lib
+                for ( auto& elempair : libpair.second ) {
+                    // loop all the loaders in the element
```
calewis, line 130: Comment probably is redundant IMO, I think a for each loop with a variable name loader is self explanatory.  Same with the above about looping the elements in the element lib.

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
-}
+    std::vector<std::string> paths = splitPath(searchPaths);
 
-std::vector<std::string>
-ElemLoader::getPotentialElements()
-{
-    std::vector<std::string> res;
-    lt_dlforeachfile(searchPaths.c_str(), elemCB, &res);
-    return res;
+    for ( std::string& next_path : paths ) {
```
calewis, line 154: `next_path` can be const

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
+        DIR* current_dir = opendir(next_path.c_str());
+
+        if ( current_dir ) {
+            struct dirent* dir_file;
+            while ( (dir_file = readdir(current_dir)) != nullptr ) {
+                // ensure we are only processing normal files and nothing weird
+                if ( (dir_file->d_type | DT_REG) || (dir_file->d_type | DT_LNK) ) {
+                    char* current_file = new char[strlen(dir_file->d_name) + 1];
```
calewis, line 162: Can this just  be a `std::string` I believe you could skip the strcpy https://godbolt.org/z/qjMEPbbEf and thus also skip the `delete[]`

src/sst/core/elemLoader.h
```diff
@@ -40,9 +35,17 @@ class ElemLoader
...
 
     /**
-     * Returns a list of potential element libraries in the search path
+     * Search paths for potential elements and add them to the provided vector
+     *
+     * @param potElems - vector of potential elements that could contain elements
+     * @return void
      */
-    std::vector<std::string> getPotentialElements();
+    void getPotentialElements(std::vector<std::string>& potElems);
```
calewis, line 43: If this is going to be used only by sstinfo. why not just have it return a std::vector instead of pass by reference for output? As in what don't you like about the old behavior?

src/sst/core/sstinfo.cc
```diff
@@ -152,10 +152,10 @@ addELI(ElemLoader& loader, const std::string& lib, bool optional)
...
 processSSTElementFiles()
 {
-    std::vector<bool> EntryProcessedArray;
-    ElemLoader        loader(g_searchPath);
-
-    std::vector<std::string> potentialLibs = loader.getPotentialElements();
+    std::vector<bool>        EntryProcessedArray;
+    ElemLoader               loader(g_searchPath);
+    std::vector<std::string> potentialLibs;
+    loader.getPotentialElements(potentialLibs);
```
calewis, line 158: Why did we make this change? I think the old code is better unless there is a good reason for this.

src/sst/core/sstinfo.cc
```diff
@@ -152,10 +152,10 @@ addELI(ElemLoader& loader, const std::string& lib, bool optional)
 static void
 processSSTElementFiles()
 {
-    std::vector<bool> EntryProcessedArray;
-    ElemLoader        loader(g_searchPath);
-
-    std::vector<std::string> potentialLibs = loader.getPotentialElements();
+    std::vector<bool>        EntryProcessedArray;
```
calewis, line 155: I know you didn't write this as part of the PR, but `vector<bool>` eeek

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +73,114 @@ splitPath(const std::string& searchPaths)
...
-    // loading a library can "wipe" previously loaded libraries depending
-    // on how weak symbol resolution works in dlopen
-    // rerun the loaders to make sure everything is still registered
-    for ( auto& libpair : ELI::LoadedLibraries::getLoaders() ) {
-        // loop all the elements in the element lib
-        for ( auto& elempair : libpair.second ) {
-            // loop all the loaders in the element
-            for ( auto* loader : elempair.second ) {
-                loader->load();
+    for ( std::string& next_path : paths ) {
```
calewis, line 104: `for(std::string const& next_path : ...`  Unless you hate east const then you can make it west :P

calewis (changes requested):

src/sst/core/elemLoader.cc
nmhamster: So much for self-documenting code.

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +56,132 @@ splitPath(const std::string& searchPaths)
...
+            while ( (dir_file = readdir(current_dir)) != nullptr ) {
+                // ensure we are only processing normal files and nothing weird
+                if ( (dir_file->d_type | DT_REG) || (dir_file->d_type | DT_LNK) ) {
+                    char* current_file = new char[strlen(dir_file->d_name) + 1];
+                    std::strcpy(current_file, dir_file->d_name);
+
+                    // does the path start with "lib" required for SST
+                    if ( !strncmp("lib", current_file, 3) ) {
+                        // find out if we have an extension
+                        char* find_ext = strrchr(current_file, '.');
```
calewis, line 169: What if you don't have an extension?

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +56,132 @@ splitPath(const std::string& searchPaths)
...
-    // on how weak symbol resolution works in dlopen
-    // rerun the loaders to make sure everything is still registered
-    for ( auto& libpair : ELI::LoadedLibraries::getLoaders() ) {
-        // loop all the elements in the element lib
-        for ( auto& elempair : libpair.second ) {
-            // loop all the loaders in the element
-            for ( auto* loader : elempair.second ) {
-                loader->load();
+        if ( nullptr == handle ) {
+            if ( verbose ) { printf("SST-DL: Loading failed, error: %s\n", dlerror()); }
```
calewis, line 121: What happens here? Should we do something besides call `dlerror`?

nmhamster: I believe we cannot fault/fatal because we can attempt to load modules in Python and fail. The code for explicit halts was removed some time ago.

src/sst/core/elemLoader.cc
nmhamster: We are explicitly stating `.so` on Linux and `.so` and `.dylib` on macOS as supported.

calewis: Okay, I just read the comment `find out if we have ...` as extensions might be optional.

src/sst/core/elemLoader.cc
feldergast: That was a feature Brandon was very keen on, but hasn't ever actually been used to my knowledge.  The expected behavior today if you import a module that doesn't exist (either because it doesn't find the shared library or the shared library doesn't have a python module) is that you get an exception.

src/sst/core/elemLoader.cc
nmhamster: Its better.

src/sst/core/elemLoader.h
nmhamster: We discsussed this on Mattermost. If we proceed to the App-Store/Model-Store approach, we will have a `std::vector` and just call methods to treat them as an append-like operation.

src/sst/core/elemLoader.h
nmhamster: This is migrated to the constructor, I personally don't like setting variables in the class definition variable listing.

src/sst/core/elemLoader.h
nmhamster: See above.

src/sst/core/sstinfo.cc
nmhamster: I understand this is automatically compiled as a bit field anyway.

src/sst/core/elemLoader.h
calewis: Is this PR the right place to do this?

src/sst/core/elemLoader.cc
calewis: ? Why would you not want to initialize it to the expected default value.

src/sst/core/elemLoader.cc
calewis: 😬

src/sst/core/elemLoader.cc
calewis: This one needs to happen

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +56,132 @@ splitPath(const std::string& searchPaths)
     return paths;
 }
```
calewis, line 57: I'll suggest: 
```cpp
std::vector<std::string>
splitPath(std::string const& searchPaths)
{
    std::vector<std::string> paths;
    std::string::size_type   start = 0, pos = 0;
    while ( (pos = searchPaths.find(':', start)) != std::string::npos ) {
        paths.emplace_back(searchPaths.substr(start, pos - start));
        start = pos + 1;
    }
    paths.emplace_back(searchPaths.substr(start));

    return paths;
}
```
Doesn't need to copy the input string as a clear benefit.

src/sst/core/elemLoader.cc
calewis: @feldergast so we should change this or leave it?

src/sst/core/elemLoader.cc
```diff
@@ -135,92 +56,132 @@ splitPath(const std::string& searchPaths)
...
+                for ( auto& elempair : libpair.second ) {
+                    // loop all the loaders in the element
+                    for ( auto* loader : elempair.second ) {
+                        loader->load();
+                    }
+                }
             }
+
+            // exit the search loop, we have found the library we tried to load
+            break;
```
calewis, line 139: Can just return here and get rid of `found_element`

src/sst/core/elemLoader.cc
calewis: @nmhamster just mark this resolved if you don't want to change it.

src/sst/core/elemLoader.cc
nmhamster: I am suggesting that we leave this for now. I am not re-engineering all the handling of exceptions and crud in this PR. The point here is to keep PRs relatively small, focused - like this one, getting us out from `libtldl` so we can move forward with the CMake and no `libtool`. We should not need to re-engineer the core for every PR.

feldergast: I think the code as is will create the expected result, maybe plus or minus some console output. Further, it looks like there'll be enough information printed to help with debugging. Since no one really tries to import a library and then catches exceptions, I think this works for now.  If we start doing some of the exception handling, we may want to gather the text into a string to send with an exception, but we can revisit that later if it ever comes up.

nmhamster:
@calewis / @feldergast - I think this PR has been done to death, can you approve it so we can merge it please?

calewis (approved):
Approving based on conversations on internal communication platforms.

PR #691, opened 2021-08-09
Made the SSTModelDescription into a loadable module.

feldergast:
- Added eli/simpleInfo.h to allow for quick addition of ELI info to base APIs.
- Changed Factory::Create() to Factory::CreateWithParams() because it really only makes sense to use for elements to require a Params object that has the valid keys set.
- Added new Factory::Create() method that doesn't requite a Params object (or you don't need the valid keys populated).
- main() now looks for all SSTModelDescriptions registered with the ELI database and choses the correct one based off of reported supported file extensions.  This will make it easier to add new models in the future.


src/sst/core/main.cc
```diff
@@ -368,36 +368,44 @@ main(int argc, char* argv[])
...
+            extension_map[y] = x;
+        }
+    }
+
     SSTModelDescription* modelGen = nullptr;
 
     double start = sst_get_cpu_time();
 
     if ( cfg.configFile != "NONE" ) {
         string file_ext = "";
```
gvoskuilen, line 391: It doesn't look like 'file_ext' is used anymore - can we remove this?

feldergast: Probably.  I'm about to do another clang-format PR, how about if I fix it there (unless something else pressing needs to be fixed in this PR)?

nmhamster: @feldergast / @gvoskuilen - I went ahead and merged so that @feldergast can fix this up in the clang-format PR which is coming in due course.

PR #692, opened 2021-08-09
Changing the clang-format script to accept the `-i` option.

calewis:
Providing `-i` will ask clang-format to make the changes for you.  

I tested this with `--format-exe` but bash is not something I'm comfortable doing input parsing with so an extra set of eyes would be nice. 

@allevin I'm picking you since I think you know bash better than most of us.


allevin (approved):
Looks good to me

PR #693, opened 2021-08-09
Remove libltdl directory creation

nmhamster:
Removes `libltdl` generation and integration with SST core.


calewis (approved):
Let's see this pass

nmhamster:
> Let's see this pass

Hold up their mister ... .that's sounding way to confident!

PR #696, opened 2021-08-12
Add performance profiling code

matthewbarondeau:
Erase this information and put your Pull Request comments here
---

----

### DIFF
--- a/src/sst/core/event.cc
+++ b/src/sst/core/event.cc
@@ -24,10 +24,68 @@ const SST::Event::id_type SST::Event::NO_ID = std::make_pair(0, -1);
 
 Event::~Event() {}
 
+extern std::map<std::string, uint64_t> eventHandlers;
+extern std::map<std::string, uint64_t> eventSendCounters;
+extern std::map<std::string, uint64_t> eventRecvCounters;
 
 void Event::execute(void)
 {
+
+    #ifdef EVENT_PROFILING
+    #ifdef HIGH_RESOLUTION_CLOCK
+    auto eventStart = std::chrono::high_resolution_clock::now();
+    #else
+    struct timeval eventStart, eventEnd, eventDiff;
+    gettimeofday(&eventStart, NULL);
+    #endif
+    #endif
+
     delivery_link->deliverEvent(this);
+
+    #ifdef EVENT_PROFILING
+    #ifdef HIGH_RESOLUTION_CLOCK
+    auto eventFinish = std::chrono::high_resolution_clock::now();
+    #else
+    gettimeofday(&eventEnd, NULL);
+    timersub(&eventEnd, &eventStart, &eventDiff);
+    #endif
+
+    // Track receiver processing time
+    auto eventHandler = eventHandlers.find(getLastComponentName());
+    if(eventHandler != eventHandlers.end())
+    {
+        #ifdef HIGH_RESOLUTION_CLOCK
+        eventHandler->second += std::chrono::duration_cast<std::chrono::nanoseconds>(eventFinish - eventStart).count();
+        #else
+        eventHandler->second += eventDiff.tv_usec + eventDiff.tv_sec * 1e6;
+        #endif
+    }
+
+    // Track sending and receiving counters
+    auto eventCount = eventRecvCounters.find(getLastComponentName());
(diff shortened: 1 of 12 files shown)


src/sst/core/simulation.cc
```diff
@@ -596,6 +596,14 @@ void Simulation_impl::setup() {
 
 }
 
+#ifdef EVENT_PROFILING
+std::map<std::string, uint64_t> eventHandlers;
+std::map<std::string, uint64_t> eventRecvCounters;
+std::map<std::string, uint64_t> eventSendCounters;
+uint64_t messageSizeSent = 0;
+uint64_t messageSizeRecv = 0;
```
feldergast, line 604: Can't have these as globals.  I think this will work today because simulation construction is serialized across threads.  However, we will probably need to move to parallel construction for really large simulations at some point, at which time these would be accessed concurrently by multiple threads.  You're going to have to move these data structures into the Simulation objects themselves so that they are limited to access by a single thread at a time.

Actually, on further inspection, it looks like there is the possibility these may be accessed concurrently by different threads.

src/sst/core/event.cc
```diff
@@ -24,10 +24,68 @@ const SST::Event::id_type SST::Event::NO_ID = std::make_pair(0, -1);
...
+    {
+        #ifdef HIGH_RESOLUTION_CLOCK
+        eventHandler->second += std::chrono::duration_cast<std::chrono::nanoseconds>(eventFinish - eventStart).count();
+        #else
+        eventHandler->second += eventDiff.tv_usec + eventDiff.tv_sec * 1e6;
+        #endif
+    }
+
+    // Track sending and receiving counters
+    auto eventCount = eventRecvCounters.find(getLastComponentName());
```
feldergast, line 65: Probably better to use ComponentId's instead of names in this data structure.  The string can get rather long with SubComponents, and this will be a HUGE memory overhead if you're not careful.  You may need to dump the mapping to file somewhere to resolve things later.

src/sst/core/event.cc
```diff
@@ -24,10 +24,68 @@ const SST::Event::id_type SST::Event::NO_ID = std::make_pair(0, -1);
...
+    if(eventCount != eventRecvCounters.end())
+    {
+        eventCount->second ++;
+    }else{
+        if(getLastComponentName() != "")
+        {
+            eventRecvCounters.insert(std::pair<std::string, uint64_t>(getLastComponentName().c_str(), 1));
```
feldergast, line 72: This has the possibility of concurrent accesses to the eventRecvCounters object since it can happen for the first time after the run loop has started.

src/sst/core/event.cc
```diff
@@ -24,10 +24,68 @@ const SST::Event::id_type SST::Event::NO_ID = std::make_pair(0, -1);
...
+    }
+    auto eventSend = eventSendCounters.find(getFirstComponentName());
+    if(eventSend != eventSendCounters.end())
+    {
+       eventSend->second++;
+    }else{
+        // Insert handler and counter for the subcomponent so that all link traffic is monitored
+        if(getFirstComponentName() != "")
+        {
+            eventSendCounters.insert(std::pair<std::string, uint64_t>(getFirstComponentName().c_str(),1));
```
feldergast, line 84: This has the possibility of concurrent accesses to the eventSendCounters object since it can happen for the first time after the run loop has started.

feldergast (changes requested):
I didn't look through the entire change set, but I'm relatively certain you have the possibility of concurrent write access to global data structures that aren't thread safe for writes.  In addition, even if I missed something in that code and you only end up with concurrent reads today, we would like to reserve the ability to move to concurrent building of the simulation (graph -> actually instanced objects) in the future, so we'll need this code to be able to support that.  Please avoid using statics in classes or global data structures whenever possible.  They make verifying that threading will work in all cases a nightmare, and I already have several of these types of potential issues to track down in the existing code.  The Simulation object is the right place to keep data that is global for a thread.

feldergast:
Also, please cancel this pull request and create a new one that targets sstsimulator:devel, not master.  Please make sure you also branched off of current devel.  There have been a lot of recent changes.  Closing this request.

PR #698, opened 2021-08-16
Adding GHA clang-format test

allevin:
This PR will add a Github Actions script that runs on PR creation or commit to a PR.  It will clone the repo on Github's test systems and then run clang-format v12.  It will set labels and add appropriate comments based upon testing.  It will also set/remove the AT: WIP flag to lockout Autotester testing if the format test fails.

### DIFF
--- a/.github/workflows/PR_clang-format_test.yml
+++ b/.github/workflows/PR_clang-format_test.yml
@@ -1,0 +1,173 @@
+name: SST-CORE clang-format TEST
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches-ignore:
+      - master
+      - devel
+# Push option may be used in the future for developers forks
+#  push:
+#    branches-ignore:
+#      - master
+#      - devel
+
+# This script is used to do a quick and dirty clang-format v12 check on the
+# PR source repo to ensure it is formatted correctly.  It will initially set
+# the AT: WIP label to try to stop the Autottester from processing the PR, and
+# then perform the clang-format test using a 3rd party action.  It will then
+# set labels and add comments in the conversation  for Pass or Failure as
+# appropriate.  Any labels that do not exist will be created.  Also we use
+# the step command "continue-on-error" to prevent a trying to remove a
+# missing label from failing the script.
+
+### NOTE: This script owes a lot to the work of Stephanie Eckles; see:
+### https://dev.to/5t3ph/github-workflows-for-newbies-add-labels-and-comments-to-pull-requests-37da
+### Also look at https://octokit.github.io/rest.js/v18 for available script commands
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  Test_With_clang-format:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]
+        python-version: [3.8]
+    name: Test:${{ matrix.os }}/PY-${{ matrix.python-version }}
+    runs-on: ${{ matrix.os }}
+
+    steps:
+
+# Set the WIP label immediately before testing
+    - name: SET WIP Label before testing
+      uses: actions/github-script@v4
+      with:
+        github-token: ${{secrets.GITHUB_TOKEN}}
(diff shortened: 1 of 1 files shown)


.github/workflows/PR_clang-format_test.yml
```diff
@@ -0,0 +1,171 @@
...
+        script: |
+            github.issues.addLabels({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: ['AT: CLANG-FORMAT PASS']
+            })
+      if: ${{ success() }}
+
+    - name: REMOVE WIP Label Results for Success
```
calewis, line 132: So I need to learn more about actions, but is this going to run regardless of whether or not a user sets the WIP flag? Because we don't want to accidentally remove a WIP flag if it was set by the submitter.

calewis (approved):
Can we check that one comment I have about not removing submitter set WIP, it's an edge case, but we should double check.  I'm going to approve because I think this is good and it worked on my first order test on my fork and we can make changes if/when issues arise.

.github/workflows/PR_clang-format_test.yml
allevin: The AT: WIP label is one way.  Also the user can set [WIP] in the title of the PR to have the same effect.   That might be a better solution for PR's that are not wanted to be processed.

feldergast:
A few questions before we merge this:

1 - When will this run?  Only at PR submission?  After a new commit is pushed?
2 - Will this do more than add WIP to the PR (i.e., is there a check to make sure this is successful that will be required before merging)?  Or, are we leaving in the clang-format test in the autotester to make do the actual check that will be needed before merging?
3 - Should this have run when this PR was submitted (I don't see any comments that it generated), or will it only run once it is committed into devel?

feldergast (changes requested):
Just going to set this as request changes to prevent accidentally merging before we get a few questions answered.  See comments.

allevin:
> A few questions before we merge this:
> 
> 1 - When will this run? Only at PR submission? After a new commit is pushed?
> 2 - Will this do more than add WIP to the PR (i.e., is there a check to make sure this is successful that will be required before merging)? Or, are we leaving in the clang-format test in the autotester to make do the actual check that will be needed before merging?
> 3 - Should this have run when this PR was submitted (I don't see any comments that it generated), or will it only run once it is committed into devel?

@feldergast, @calewis has tested the script and verified its operation (with some minor corner case issues)

1 - From testing (using my forked repo), it will run when the PR is created and after each commit to the PR's related source branch.  However there may be differences between operation that we wont discover till this is merged (My major concern is its operation on PR's coming from forks - which I could not test)

2 - The script affects 3 labels `AT: WIP`, `AT : CLANG-FORMAT PASS`, `AT: CLANG-FORMAT FAIL`.  Only the AT: WIP affects the Autotester the others are just information labels.  If the script fails, it will add a new status (like the pre-merge and pre-test inspections) indicator to the PR that will prevent merging.  This is done by GHA not the Autotester.  Passing will set the status indicator to success and will allow merging.
  * 2.1 - The Autotester and nightly runs will still continue to call the Jenkins job that performs the clang-format test using the script in the core.  This GHA script uses an 3rd party action to perform the clang-format test.
  * 2.2 - There are 2 concerns with this script:
     *  If the developer sets the `AT: WIP` label before this script runs, and the script passes, the WIP label will be removed.
     * There is a race condition between this script and the Autotester run.  The majority of the time this script will be run within 30 seconds of the PR create/commit and then set the WIP flag.  However if the stars align, the AT could process the PR just before the script runs (this is unlikely, but possible)

3 - I though it might run also, however since its not officially part of the Github repo yet, that may be why it did not run.

The script is fairly benign, and once merged, we can submit a few test PR's to ensure that it is working correctly.  If we have problems with it, we can temporarily turn of GHA till the issue can be resolved.

feldergast (approved):
Questions answered.  I'll go ahead and merge for further testing.

PR #714, opened 2021-08-30
Include updates

feldergast:
Removed some unused core tests and changed files to use quotes instead of angle brackets for sst header includes.


.clang-format
```diff
@@ -82,17 +82,20 @@ IncludeBlocks: Regroup
...
+    Priority:        4
+    SortPriority:    4
+  - Regex:           '("[^\(sst\)].*")'
     Priority:        3
     SortPriority:    3
-  - Regex:           '("[^\(sst\)].*")'
-    Priority:        2
-    SortPriority:    2
-  - Regex:           '("sst/core/.*")'
+  - Regex:           '("sst/core/testElements/.*")'
```
calewis, line 90: Why do we need to single this specific folder out?

calewis (approved):
Seems reasonable to me, although I'm not sure why `testElements` needs to be called out specifically.

.clang-format
feldergast: Even though testElements is part of the core repository, it should be treated like an element library.  Hence, its header files should come before other core header files.

gvoskuilen (approved):

calewis (approved):

gvoskuilen (approved):

Issue #711, opened 2021-08-26
CMake build should error if autotools build artifacts are present


calewis:
We are making the decision that the same directory shouldn't be used for cmake and autotools builds, since there may be issues with the fact that the autotools build puts files like `sst_config.h` in the source directory and we might include those folders leading to header mismatches.  

git worktree is a nice way to have separate branches in separate folders, but this will make it slightly painful to use a single source dir for a specific branch for both autotools and cmake.  

@nmhamster, in fact if we want to revisit the reason for this, It does seem like that would make it hard for core developers to ensure their changes work in both cmake and autotools builds during the period that we have both.   Imagine you add a file and test that it works in autotools build, but before you push you want to test that it works in the cmake build as well 🤷‍♀️ if we block the cmake build that will be difficult.

Issue #710, opened 2021-08-26
CMake build should block in source builds


calewis:
Title is self explanatory.

PR #717, opened 2021-08-31
Cmake block in-source builds. 

calewis:
This will prevent the user from doing one of two things
1. The user can't invoke `cmake experimental/` in the top level source dir
2. The user can't use cmake if they have already run an in-source autotools build, which is detected by the presence of sst_config.h or build_info.h in the source folder. 

In both cases the cmake configure will quickly error. 

This should address #710 and #711


feldergast (approved):
I’m not a crake expert, but these changes look good to me.

Issue #719, opened 2021-09-01
Cmake use variable to hold top-level source


calewis:
This is a recommendation that Jeremy made that I agree with, but I didn't want to hold up #712 to do it.

berquist:
Done in #720, but will likely be undone in the future once CMake "graduates".

Issue #716, opened 2021-08-31
Format cmake


calewis:
Similar to clang-format, just set a format file for cmake and be done with it.

jleidel:
is this still an open issue?

berquist:
We should have a clang-format config file in the repo, even if it's the current tool defaults, and have it run in GHA like clang-format does.

jleidel:
EG, point at this one? https://github.com/sstsimulator/sst-core/blob/master/.clang-format

berquist:
Oops, typo fixed.

berquist:
@jleidel See #963.  I don't have any particular style preferences, even though I have some style habits, but if there's anything specific you think should be enforced, please comment.

I could see enforcing dangling trailing paren important because changes from
```cmake
X)
```
to
```cmake
X,
Y)
```
unfairly modify `X` and make it harder to blame, but I'm not picky.

berquist:
Done via #963 and #964.

PR #720, opened 2021-09-01
Cmake format and source dir variable renaming

calewis:
In this pr I did two things

1. Ran cmake-format with the default format for the tool
2. Changed all of the `${sst-core_SOURCE_DIR}/..` to `${SST_TOP_SRC_DIR}` which was one of the suggestions by Jeremy and will also make it easier to move cmake out of experimental at some point in the future. 

I'm not really certain how we want to handle cmake formatting in the future, but I wanted to get the files into a standardized state before we really start having users.


calewis:
This should address #719 and #716

gvoskuilen (approved):

PR #722, opened 2021-09-02
Handler update

feldergast:
Created a new core-wide handler base class

### DIFF
--- a/src/sst/core/ssthandler.h
+++ b/src/sst/core/ssthandler.h
@@ -39,7 +39,7 @@ namespace SST {
 // using HandlerBase = SSTHandlerBaseNoArgs<return_type_of_callback>;
 
 // template <return_type_of_callback, typename classT, typename dataT = void>
-// using Handler = SSTHandlerNoArgs<classT, dataT>;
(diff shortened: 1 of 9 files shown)


src/sst/core/ssthandler.h
```diff
@@ -0,0 +1,168 @@
...
+
+// template <typename classT, typename dataT = void>
+// using Handler = SSTHandler<return_type_of_callback, arg_type_of_callback classT, dataT>;
+
+// Or:
+
+// using HandlerBase = SSTHandlerBaseNoArgs<return_type_of_callback>;
+
+// template <return_type_of_callback, typename classT, typename dataT = void>
+// using Handler = SSTHandlerNoArgs<classT, dataT>;
```
gvoskuilen, line 42: Should this be:
`using HandlerBase = SSTHandlerNoArgs<return_type_of_callback, classT, dataT>;`

feldergast: Not quite, but you are right that line 42 is wrong, it should just be "using Handler = ..." (not HandlerBase).  I'll fix it.

gvoskuilen (approved):

PR #728, opened 2021-09-15
Fix include paths which can affect Apple XCode 12.5 with new C++ Compilers

nmhamster:
Removes the include of base directories that can affect Apple XCode 12.5 builds.


feldergast (approved):
Looks good.

calewis (approved):
LGTM

PR #734, opened 2021-10-08
Performance Instrumenting - new rebase

donofrio:
This is an updated version of the Performance Instrumenting (#703) pull request.  Relevant changes:
 - HandlerBase now has a GetId() function rather than the pointer being used as an identifier. This ID is created by the factory and is unique per Clock Handler. 
 - New test added: test_PerfComponent which checks the current build configuration to test if the performance counter data is sane.  test_PerfComponent is very similar to test_Component except fewer elements and each clock loop takes more clock cycles to ensure the performance counters are working. This tests passes regardless of whether the performance tracking is enabled or not


donofrio:
Merged as part of #768

PR #742, opened 2021-10-18
Replace sprintf with snprintf.  

calewis:
I skipped the pythonConfig files since they were a bit more complicated. 

The motivation here is to try and track down the OS X issue that has misbehaving output.


feldergast:
Superseded by #832

PR #748, opened 2021-11-02
Config output updates

feldergast:


src/sst/core/cfgoutput/jsonConfigOutput.cc
```diff
@@ -62,58 +66,93 @@ to_json(json::json& j, StatPair const& sp)
...
 
     auto compMap = graph->getComponentMap();
     auto linkMap = graph->getLinkMap();
 
-    json::json outputJson;
+    json::ordered_json outputJson;
+
+    // Put in the global param sets
+    for ( const auto& set : Params::getGlobalParamSetNames() ) {
+        for ( auto kvp : Params::getGlobalParamSet(set) ) {
```
calewis, line 149: Do you want to copy `kvp`?

src/sst/core/cfgoutput/pythonConfigOutput.cc
```diff
@@ -67,9 +68,14 @@ void
 PythonConfigGraphOutput::generateCommonComponent(const char* objName, const ConfigComponent* comp)
 {
     if ( !comp->params.empty() ) {
+        // Add local params
         fprintf(outputFile, "%s.addParams(", objName);
         generateParams(comp->params);
         fprintf(outputFile, ")\n");
+        // Add global param sets
+        for ( auto x : comp->params.getSubscribedGlobalParamSets() ) {
```
calewis, line 76: Probably no need to copy `x`

src/sst/core/cfgoutput/pythonConfigOutput.cc
```diff
@@ -208,12 +218,27 @@ PythonConfigGraphOutput::generate(const Config* cfg, ConfigGraph* graph)
     fprintf(outputFile, "sst.setProgramOption(\"timebase\", \"%s\")\n", cfg->timeBase.c_str());
     fprintf(outputFile, "sst.setProgramOption(\"stopAtCycle\", \"%s\")\n\n", cfg->stopAtCycle.c_str());
 
+    // Output the global params
+    fprintf(outputFile, "# Define the global parameter sets:\n");
+    std::vector<std::string> global_param_sets = Params::getGlobalParamSetNames();
+    for ( auto& x : global_param_sets ) {
+        fprintf(outputFile, "sst.addGlobalParams(\"%s\", {\n", x.c_str());
+        for ( auto y : Params::getGlobalParamSet(x) ) {
```
calewis, line 226: Copy pairs of strings?

src/sst/core/params.cc
```diff
@@ -480,6 +480,54 @@ Params::getArrayTokens(const std::string& value, std::vector<std::string>& token
...
+    for ( auto x : it->second ) {
+        ret[getParamName(x.first)] = x.second;
+    }
+    return ret;
+}
+
+std::vector<std::string>
+Params::getGlobalParamSetNames()
+{
+    std::vector<std::string> ret;
```
calewis, line 500: If we know the size of the global params we should reserve

src/sst/core/params.cc
```diff
@@ -480,6 +480,54 @@ Params::getArrayTokens(const std::string& value, std::vector<std::string>& token
...
+        ret.push_back(x.first);
+    }
+    return ret;
+}
+
+
+std::vector<std::string>
+Params::getLocalKeys() const
+{
+    std::vector<std::string> ret;
```
calewis, line 511: same

src/sst/core/params.cc
```diff
@@ -480,6 +480,54 @@ Params::getArrayTokens(const std::string& value, std::vector<std::string>& token
...
+    }
+    return ret;
+}
+
+
+std::vector<std::string>
+Params::getSubscribedGlobalParamSets() const
+{
+    std::vector<std::string> ret;
+    // Skip the first item because those are local params
```
calewis, line 523: Should reserve

src/sst/core/params.h
```diff
@@ -616,6 +616,65 @@ class Params : public SST::Core::Serialization::serializable
     static void
     insert_global(const std::string& set, const key_type& key, const key_type& value, bool overwrite = true);
 
+
```
calewis, line 619: I think it's dangerous to put functions in a public place and then expect people not to call them.  If I'm using an editor with completion I might just see a method `getGlobalParamSet(std::string const&) -> std::map<std::string, std::string>` and not realize that you didn't intend for this to be called by users.

calewis (commented):
A few minor things, the not copying + reserving probably doesn't really matter, but I do think that comments are insufficient to stop people from calling methods you put in the `SST` namespace.

src/sst/core/params.h
feldergast: Yea, this is a tricky issue.  Technically, the whole configuration interface (including Params) needs to be public because you can implement inputter and outputter classes in element libraries (or will soon be able to).  However, that doesn't mean they should be used by Component/SubComponent.  I'm afraid there's no way to differentiate that other than documentation.  For now, I'm not claiming backward compatibility and scourging people from using them because they could still in flux.  Once they've stabilized and all the APIs for writing your own inputter/outputter in you element library are finalized, the comments will change to reflect where they should/should not be used.

src/sst/core/cfgoutput/jsonConfigOutput.cc
feldergast: Fixed

src/sst/core/cfgoutput/pythonConfigOutput.cc
feldergast: Fixed

src/sst/core/cfgoutput/pythonConfigOutput.cc
feldergast: Fixed

src/sst/core/params.cc
feldergast: Fixed

src/sst/core/params.cc
feldergast: Fixed

src/sst/core/params.cc
feldergast: Fixed

src/sst/core/params.h
calewis: Can't you just do 
```c++
namespace detail {
 static void insert_global(...);
}
```
It doesn't stop someone from calling it, but the extra namespace make it's clear this is Core guts and isn't required to remain compatible or be stable.

feldergast: Yes, but then when I finalize the interface I have to go back and remove the detail namespace from everything.  Remember, this does need to be public outside of core once it's finalized and a few other pieces are in place.

src/sst/core/params.cc
```diff
@@ -497,7 +497,7 @@ Params::getGlobalParamSet(const std::string& name)
 std::vector<std::string>
 Params::getGlobalParamSetNames()
 {
-    std::vector<std::string> ret;
+    std::vector<std::string> ret(global_params.size());
```
calewis, line 500: I meant `std::vector<T>::reserve` https://en.cppreference.com/w/cpp/container/vector/reserve, 

This code will construct a vector of size `global_params.size()` which will have default constructed strings and then additionally push back another `size` number of strings.

feldergast: Arg.  Thanks.  I mess that up from time to time with vectors.  I'll go back and fix it.

calewis (approved):
LGTM

PR #751, opened 2021-11-04
Adding JSON model reader infrastructure

jleidel:
Adding JSON model reader infrastructure using PR #748 as the basis for the format.  This infrastructure allows users to specify model inputs using a standard JSON format.  

Supports
- components
- subcomponents
- recursive subcomponents (subcomponents of subcomponents)
- links
- program options

Does not support:
- statistics (forthcoming based upon the changes to the writer output)

### DIFF
--- a/src/sst/core/model/json/jsonmodel.cc
+++ b/src/sst/core/model/json/jsonmodel.cc
@@ -13,12 +13,6 @@
 
 #include "sst/core/model/json/jsonmodel.h"
 
-#ifdef SST_CONFIG_HAVE_MPI
-DISABLE_WARN_MISSING_OVERRIDE
-#include <mpi.h>
-REENABLE_WARNING
-#endif
-
 #include <string>
 
 DISABLE_WARN_STRICT_ALIASING


src/sst/core/model/json/jsonmodel.h
```diff
@@ -0,0 +1,125 @@
...
+#define SST_MODEL_JSON_COMPONENTID_IDX     1
+#define SST_MODEL_JSON_CONFIGCOMPONENT_IDX 2
+
+    typedef std::vector<std::tuple<std::string, ComponentId_t, ConfigComponent*>> CompTuple;
+    CompTuple                                                                     configGraphTuple;
+
+    void          recursiveSubcomponent(ConfigComponent* Parent, nlohmann::basic_json<> compArray);
+    void          discoverProgramOptions(json jFile);
+    void          discoverComponents(json jFile);
+    void          discoverLinks(json jFile);
```
calewis, line 82: Just one quick point before a bigger review, unless the json objects are shallow copy, can these be `const&` or do you use the `jFile` for temp storage in these functions?

src/sst/core/model/json/jsonmodel.cc
```diff
@@ -0,0 +1,390 @@
...
+// Copyright (c) 2009-2021, NTESS
+// All rights reserved.
+//
+// This file is part of the SST software package. For license
+// information, see the LICENSE file in the top level directory of the
+// distribution.
+
+#include "sst/core/model/json/jsonmodel.h"
+
+#ifdef SST_CONFIG_HAVE_MPI
```
feldergast, line 16: Is there are reason to include mpi in this file?  I don't see MPI used anywhere.

jleidel: This was copied over from the python model reader, so I can remove it.

src/sst/core/model/json/jsonmodel.h
jleidel: they can be const

jleidel: on second thought, no.  it causes an assertion inside nholman::json

src/sst/core/model/json/jsonmodel.cc
jleidel: removed in commit cc7003ef

feldergast (commented):
Okay to test

feldergast (approved):

PR #775, opened 2022-01-05
TestFrameworks addtions

gvoskuilen:
Increase default test time limit to 120s from 60s
Add statistic diff function to diff statistics dumped to console


feldergast (approved):
I did not fully parse out the regex, but looks good to me.

PR #779, opened 2022-01-10
Improved the error message printed when a library fails to load.

feldergast:
The loader will now gather all loading errors from dlopen for existing files.  These will be printed as an error if the library doesn't load and as a warning if it does find a valid library to load.


gvoskuilen (approved):
Looks good.

PR #780, opened 2022-01-10
Fixed build when --disable-mpi is used in configure

feldergast:


calewis (dismissed):
Seems fine, while not part of this commit, it's slightly confusing that `myRank` would need to call `MPI_Barrier` world_size - (myrank + 1) number of times? 

How without more context the goal of this loop is very confusing.

gvoskuilen (dismissed):

gvoskuilen (approved):

PR #783, opened 2022-01-12
Added checks for valid names for links and components.

feldergast:


src/sst/core/configGraph.cc
```diff
@@ -30,6 +31,48 @@ namespace SST {
 
 // static bool zero_latency_warning = false;
 
+// Functions to check component and link names
+static int   bad_comp_name_count = 0;
+static int   bad_link_name_count = 0;
+static regex valid_name_regex =
+    regex("((([_][a-zA-Z0-9])|[a-zA-Z])[_a-zA-Z0-9]*)(.(([_][a-zA-Z0-9])|[a-zA-Z])[_a-zA-Z0-9]*)*");
+
+static const int max_invalid_name_prints = 10;
```
calewis, line 40: This can probably be `constexpr` unless you want to keep it similar to the `bad_...`.

src/sst/core/configGraph.cc
```diff
@@ -30,6 +31,48 @@ namespace SST {
 
 // static bool zero_latency_warning = false;
 
+// Functions to check component and link names
+static int   bad_comp_name_count = 0;
+static int   bad_link_name_count = 0;
+static regex valid_name_regex =
```
calewis, line 37: All these statics can just be wrapped in an anonymous namespace and save on putting `static` in front of them. It doesn't change behavior, just how you write it.

src/sst/core/configGraph.cc
```diff
@@ -682,23 +725,10 @@ ConfigGraph::checkForStructuralErrors()
...
-    if ( !ret.second ) {
-        output.fatal(CALL_INFO, 1, "ERROR: trying to add Component with name that already exists: %s\n", name.c_str());
-    }
-    return cid;
-}
-
 ComponentId_t
 ConfigGraph::addComponent(const std::string& name, const std::string& type)
 {
+    checkForValidComponentName(name);
```
calewis, line 731: Return value ignored?

src/sst/core/configGraph.cc
```diff
@@ -30,6 +31,48 @@ namespace SST {
...
 
+// Functions to check component and link names
+static int   bad_comp_name_count = 0;
+static int   bad_link_name_count = 0;
+static regex valid_name_regex =
+    regex("((([_][a-zA-Z0-9])|[a-zA-Z])[_a-zA-Z0-9]*)(.(([_][a-zA-Z0-9])|[a-zA-Z])[_a-zA-Z0-9]*)*");
+
+static const int max_invalid_name_prints = 10;
+
+bool
```
calewis, line 42: If the goal is to eventually error in this function you could make it return `void` then when you are ready to error you can add the error code later.  

As written it looks like you want the caller of this function to read the `bool` and decide what to do about it, but that's not how the code that calls this is using it.

src/sst/core/configGraph.cc
```diff
@@ -22,13 +22,126 @@
 
 #include <algorithm>
 #include <fstream>
+#include <regex>
```
gvoskuilen, line 25: Remove this include (in next PR OK)

src/sst/core/configGraph.cc
```diff
@@ -745,6 +775,7 @@ ConfigGraph::addLink(
     ComponentId_t comp_id, const std::string& link_name, const std::string& port, const std::string& latency_str,
     bool no_cut)
 {
+    checkForValidLinkName(link_name);
```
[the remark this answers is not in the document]
gvoskuilen: Still needs to be fixed. @feldergast promises to do it in the next PR.

gvoskuilen (approved):

PR #787, opened 2022-01-18
Added param, port and slot name checking.

feldergast:
Also created a central place for the name check code in namecheck.{h,cc}.

### DIFF
--- a/src/sst/core/namecheck.h
+++ b/src/sst/core/namecheck.h
@@ -1,0 +1,61 @@
+// Copyright 2009-2021 NTESS. Under the terms
+// of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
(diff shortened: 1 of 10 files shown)


src/sst/core/namecheck.h
```diff
@@ -0,0 +1,61 @@
...
+       following formats: %d, %(some documentation)d.  The use of dots
+       and wildcards can be turned on and off with the proper flags.
+
+       @param name Name to check for validity
+
+       @param allow_wildcard If true, the %d format is considered
+       valid, if false, using %d will result in a failed check
+
+       @param allow_dot If true, a dot ('.') can be used in the name,
+       if falise, using a dot will result in a failed check
```
gvoskuilen, line 41: "falise" --> "false"

gvoskuilen (approved):
Looks good.

PR #818, opened 2022-03-30
Deprecated Simulation::getSimulation() 

feldergast:
Added all APIs that weren't already in BaseComponent to the BaseComponent API.

### DIFF
--- a/src/sst/core/baseComponent.h
+++ b/src/sst/core/baseComponent.h
@@ -109,20 +109,28 @@ public:
      */
     virtual void printStatus(Output& UNUSED(out)) { return; }
 
+    /** Get the core timebase */
+    UnitAlgebra getCoreTimeBase() const;
     /** Return the current simulation time as a cycle count*/
     SimTime_t   getCurrentSimCycle() const;
     /** Return the current priority */
     int         getCurrentPriority() const;
     /** Return the elapsed simulation time as a time */
     UnitAlgebra getElapsedSimTime() const;
+    /** Return the end simulation time as a cycle count*/
+    SimTime_t   getEndSimCycle() const;
     /** Return the end simulation time as a time */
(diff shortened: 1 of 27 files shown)


src/sst/core/baseComponent.h
```diff
@@ -115,6 +115,8 @@ class BaseComponent
     int         getCurrentPriority() const;
     /** Return the elapsed simulation time as a time */
     UnitAlgebra getElapsedSimTime() const;
+    /** Return the end simulation time as a cycle count*/
+    SimTime_t   getEndSimCycle() const;
```
gvoskuilen, line 119: We've got `getEndSimCycle()` and `getFinalSimTime()`. Can we change these to use the same term while we're at it? Changing getEndSimCycle to getFinalSimCycle is easiest since it's a new function anyways, but I do like "End" better.

feldergast: We can change them.  We can do either getFinalSimCycle() and gerFinalSimTime() or we can do getEndSimCycle() and getEndSimTime().  In the case of the second, I would just add both and deprecate getFinalSimCycle().  I don't have a strong preference, but I think "End" is somewhat better as well.

src/sst/core/simulation.h
```diff
@@ -91,6 +102,12 @@ class Simulation
      */
     virtual void requireEvent(const std::string& name) = 0;
 
+    /** Signifies that a library is required for this simulation.
+     *  @param name Name of the library
+     */
+    virtual void requireLibrary(const std::string& name) __attribute__((deprecated(
+        "requireEvent() has been deprecated in favor or requireLibrary() and will be removed in SST 13."))) = 0;
+
```
gvoskuilen, line 110: Deprecation warning should be on requireEvent not requireLibrary.

gvoskuilen (approved):

PR #826, opened 2022-04-15
Recurse subcomponents

lpsmodsim:
I noticed that the dot graph output was only handling first order subcomponents, and not displaying subcomponents of subcomponents or so on. Updated the code to be recursive for subcomponent dot graph output


gvoskuilen (approved):
Looks good.

PR #832, opened 2022-04-21
Remove use of sprintf

feldergast:
Added format_string() function and changed all sprintf() call to use either format_string() or snprintf().  Also marked sstrng.h deprecated.


gvoskuilen (approved):
Looks good to me.

Issue #837, opened 2022-04-25
Catch HDF5 exceptions by const ref


pdbj:
HDF5 exceptions are polymorphic, must be caught by ref not value.

Patch [Catch exception by const ref](https://github.com/pdbj/sst-core/commit/00716437f19ac9cd152b6c78dff5fdcc1be01e74)

pdbj:
Rebased [Catch exception by const ref](https://github.com/pdbj/sst-core/commit/e3dbc8ee54830ddcc8929e59b670c290a63bbe51)

pdbj:
Has this been merged yet?

feldergast:
I thought we had added it to another PR, but I'm not seeing the changes in devel or master.  We'll need to do a separate PR for it.

PR #858, opened 2022-05-18
adding logic to catch hdf5 exceptions

jleidel:
Adding logic to catch HDF5 exceptions; This is a a fix for Issue #837 from @pdbj.  

The original fix from @pdbj's `wip` branch is here: https://github.com/pdbj/sst-core/commit/e3dbc8ee54830ddcc8929e59b670c290a63bbe51


feldergast:
Closing.  Fixed in #874

PR #876, opened 2022-10-06
Threading option fixes.

### LINKED ISSUE #871: --num_threads option should be --num-threads
`--num_threads` option is the only option which uses underscore `_` instead of dash `-` to separate words.  

Even the error message in `Config::setNumThreads` uses dash

An option `-num-threads` should be provided.  (Keep `num_threads` for backwards compatibility)

### LINKED ISSUE #872: Python option sst.setThreadCount doesn't work
New Issue for sst-core
----------------------
1 - Detailed description of problem or enhancement
Trying to set the thread count from Python doesn't work, using either of these calls:

```{py}
nthreads = 2
sst.setThreadCount(nthreads)
sst.setProgramOption('num_threads', nthreads)
```
2 - Describe how to reproduce the issue

a.  Add one of those calls to any Phython script
b.  Run the script with 
```
$ sst --num_threads=4 ...
```
c.  The script will run with the number of threads from the command line (4), not what was set in the Python (2)

3 - What's happening
In `src/sst/core/main.cc`:
* Line 512:  the command line is parsed by the Config object `cfg`
* Line 521:  `world_size.thread` is set with the value from cfg, which at this point only has command line input
* Line 575: finally execute the Python, but the change to thread count is never propagated to `world_size`

[Edit: added the important word "never"]

feldergast:
- Change command line option --num_threads to --num_threads.  Fixes #871
- Fix setting of threads in the Python input file.  Fixes #872

### DIFF
--- a/src/sst/core/config.cc
+++ b/src/sst/core/config.cc
@@ -798,7 +798,11 @@ static const struct sstLongOpts_s sstOptions[] = {
         "verbose", 'v', "level", "Verbosity level to determine what information about core runtime is printed",
         &ConfigHelper::incrVerbose, &ConfigHelper::setVerbosity, true),
     DEF_ARG(
-        "num_threads", 'n', "NUM", "Number of parallel threads to use per rank", &ConfigHelper::setNumThreads, true),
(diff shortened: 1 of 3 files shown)


jpkenny (approved):
LGTM

PR #877, opened 2022-10-07
Updates to ELI and sst-info

feldergast:
- Re-added SST_ELI_REGISTER_SUBCOMPONENT (same as SST_ELI_REGISTER_SUBCOMPONENT_DERIVED but we'll move away from that over time)
- Added attributes to Components and SubComponents
- Cleaned up the human readable output of sst-info
- SubComponent no longer inherits from Module


jpkenny (approved):
LGTM

PR #878, opened 2022-10-27
First attempt at metis partitioner

bremerm31:
This is a rough draft of a metis partitioner for SST. The partitioner takes the `PartitionGraph` and partitions it using Metis' `METIS_PartGraphKway`. This is a serial partitioner, but we have found in practice it's able to partition graphs with up to millions of edges in a reasonable amount of time.

The partition is currently unweighted as it assigns a weight of one per component, and a weight of 1 to each link. Simple statistics such as edge cut fraction and imbalance are outputted.

The `CSRMat` class is taken from a different project and thus may be slightly overdesigned. The main idea is to take a set of weighted vertices and a set of weighted edges. These are then formatted in a manner that is parsable by Metis.

Lastly, I took the `config/sst_check_metis.m4` from SST-elements.

Todos
 - Need to update header guard for `util.h`
 - Need to remove dependence on 64-bit version of Metis

### DIFF
--- a/config/sst_check_metis.m4
+++ b/config/sst_check_metis.m4
@@ -1,0 +1,53 @@
+AC_DEFUN([SST_CHECK_METIS],
+[
+  sst_check_metis_happy="yes"
+  sst_check_metis_requested="yes"
+
+  AC_ARG_WITH([metis],
+    [AS_HELP_STRING([--with-metis@<:@=DIR@:>@],
+      [Use Metis package installed in optionally specified DIR])],
+    [sst_check_metis_requested="yes"],
+    [sst_check_metis_requested="no"])
+
+  AS_IF([test "x$with_metis" = "xno"], [sst_check_metis_happy="no"])
+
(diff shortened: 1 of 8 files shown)


bremerm31:
Hi @gvoskuilen,

@cychan-lbnl had mentioned that you were interested in our Metis partitioner for sst-core. I thought I'd open a PR since that would make it the easiest for you to see what changes we had to introduce. Please let me know how you'd like to best proceed. I'm happy to re-write accordingly until it's ready to be merged.

Hope this helps,

Max

feldergast:
That's for submitting this PR.  I have some other commitments, so it may be a bit before I can dive into it to make any suggestions.

gvoskuilen:
@bremerm31 Thanks! We've been considering integrating Metis and this will be helpful in making it happen. Like @feldergast , it'll be a bit before I get a chance to look. But very appreciated!

bremerm31:
Sounds good! There's definitely some obvious edits to make and I will also fix the clang formatting issue.

PR #900, opened 2023-01-10
cmake builds

bpswenson:
git clone https://github.com/sstsimulator/sst-core.git
cd sst-core && mkdir build && cd build
cmake ../experimental
make -j32

OLD BEHAVIOR
sstsim.x and other targets fail to build with linking errors

NEW BEHAVIOR
build completes successfully


bpswenson:
Would you take a CMake build for elements?

gvoskuilen (approved):

PR #915, opened 2023-02-13
#32 sst-info needs to be better at finding things

bpswenson:
RE:  #32 Updated sst-info to provide more assistance when unable to find library.


gvoskuilen:
@bpswenson This has failed the clang-format check. Can you run the formatting script (./scripts/clang-format-test.sh) with clang-format v12 and update the branch? Once it passes that check, we can shepherd this through quickly. Thanks!

bpswenson:
sorry I keep meaning to do this.  brew has clang-format@11 and clang-format@8 and clang-format@15 but that script seems to only allow 12 so I need to get that.

feldergast:
@bpswenson I’m working on fixing the library search paths in sst-info so they match what SST is doing.  I can just add these changes to that PR if you’d like since I’ll already be editing that file.  That PR will likely go in next Monday.

feldergast:
These changes were included in #1033.  Closing.

PR #917, opened 2023-02-14
Move to C++17

gvoskuilen:


configure.ac
```diff
@@ -32,12 +32,12 @@ AC_CACHE_SAVE
...
 
 dnl Work out the C++ standard which we are going to use
-AX_CXX_COMPILE_STDCXX_1Y
-AX_CXX_COMPILE_STDCXX_0X
+AX_CXX_COMPILE_STDCXX_17(noext)
 
-AS_IF([test "x$found_cxx1y" = "xyes"],
-        [CXXFLAGS="$CXXFLAGS $SST_CXX1Y_FLAGS"],
-        [CXXFLAGS="$CXXFLAGS $SST_CXX0X_FLAGS"])
+CXXFLAGS="$CXXFLAGS -std=c++17"
```
jpkenny, line 37: I think you should just leave this line out. I believe the macro will just redefine CXX (CXX="$CXX $switch") and will detect different switches for different compilers. Probably not a problem for the compilers we support, but it's the way it's intended to work. (Do your compile commands have duplicate -std=c++17?)

configure.ac
```diff
@@ -32,12 +32,12 @@ AC_CACHE_SAVE
 LT_INIT([shared disable-static dlopen])
 
 dnl Work out the C++ standard which we are going to use
-AX_CXX_COMPILE_STDCXX_1Y
-AX_CXX_COMPILE_STDCXX_0X
+AX_CXX_COMPILE_STDCXX_17(noext)
```
jpkenny, line 35: If you leave off the noext argument (leave it blank) I don't think it hurts anything and could potentially make some obscure hypothetical compiler that doesn't support strict conformance work. Not a big deal.

jpkenny (commented):
A couple changes to consider.

configure.ac
gvoskuilen: noext is required but I will fix the macro so that it doesn't overwrite CXX

jpkenny (commented):
Let's just stop AX_CXX_COMPILE_STDCXX from adding -std flag to CXX/CXXCPP and call it a day.

configure.ac
gvoskuilen: That flag is required there but I am removing -std=c++17 from CXX so it's cleaner.

jpkenny (approved):
Looks good

PR #945, opened 2023-04-21
bump version to 13.0.0

jmlapre:


gvoskuilen:
@jmlapre The version updates should not occur to the devel branch. Only the release branch should have these changes.

PR #957, opened 2023-05-27
README providing context for profiling capability

researcherben:
Explanation to developers looking to use the experimental  profiling capability


feldergast:
Info now provided in extended help introduced in #977.  Closing.

PR #960, opened 2023-06-21
Minor tidying

### LINKED ISSUE #736: Fix GCC-10 warnings
When compiling with GCC-10 (10.2.1 and 10.3 on two different systems) I encountered a number of minor warnings, all fixed by this patch.  This is based on v11.0.0_FINAL
[0001-Fix-GCC10-compilation-warnings.patch.txt](https://github.com/sstsimulator/sst-core/files/7324887/0001-Fix-GCC10-compilation-warnings.patch.txt)
.

berquist:
Closes #736

These are all the patches from https://github.com/sstsimulator/sst-core/issues/736#issuecomment-1125512738 except for the case fallthrough, where the original code no longer exists.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 **/.deps/
 **/.dirstamp
 **/Makefile.in
+**~
 *.o
 .libs/
 .deps/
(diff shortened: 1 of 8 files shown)


jpkenny (approved):
Looks good

PR #962, opened 2023-06-22
add MPI include directories and definitions

jmlapre:
Experimental CMake config omits some important commands:

```
  include_directories(SYSTEM ${MPI_INCLUDE_PATH})
  add_definitions(-DSST_CONFIG_HAVE_MPI)
```

### DIFF
--- a/experimental/CMakeLists.txt
+++ b/experimental/CMakeLists.txt
@@ -54,6 +54,8 @@ endif()
 option(SST_DISABLE_MPI "Compile without MPI" OFF)
 if(NOT SST_DISABLE_MPI)
   find_package(MPI REQUIRED)
+  include_directories(SYSTEM ${MPI_INCLUDE_PATH})
+  add_definitions(-DSST_CONFIG_HAVE_MPI)
 endif()
 
 # sst sets all the variables needed for our *.in files


jpkenny (commented):
If one uses the MPI compiler wrappers, which I believe is standard practice these days, adding the include directories shouldn't be necessary. What situation is the "include_directories" fixing?

berquist:
The way to avoid the `SYSTEM` include is to link `SST::SST` or whatever the target is called to `MPI::MPI` via `target_link_libraries`.  If that doesn't work, then something is wrong the SST target.

jmlapre:
> If one uses the MPI compiler wrappers, which I believe is standard practice these days, adding the include directories shouldn't be necessary. What situation is the "include_directories" fixing?

When following the instructions in the [README](https://github.com/sstsimulator/sst-core/blob/master/experimental/CMAKE_README.md) it mentions running cmake like so:
```
cmake ../experimental
```
which will not work.  Also, setting `CXX` to `mpicxx` should be avoided as per this [comment](https://github.com/dealii/dealii/issues/5510) from the maintainer of the FindMPI cmake module.

jpkenny:
@jmlapre And isn't ```SST_CONFIG_HAVE_MPI``` already defined in sst_config.h?

jpkenny:
@jmlapre

> Also, setting `CXX` to `mpicxx` should be avoided as per this [comment](https://github.com/dealii/dealii/issues/5510#issuecomment-349761025) from the maintainer of the FindMPI cmake module.

Ok, following the "general advice" from that comment is going to be very error prone because most users don't even know what the "compiler you want to work with" is. That said, if they just set CXX to mpicxx they're normally going to be fine for common mpi implementations.

However, I don't think this changeset is a sufficient fix for mpi implementations that don't provide wrappers -- at a minimum I assume MPI_<lang>_LIBRARIES needs to be added to whatever the LD_FLAGS equivalent is.

PR #964, opened 2023-07-10
Fixes for cmake-format GitHub Action

berquist:
For #716, a follow-up to #963.  Apparently we are using an older form of the GitHub script language and I blindly updated the version for the CMake format checker.  The path also wasn't updated to the experimental location.

There is a non-syntax update that needs to be made, which is that the action only runs `cmake-format --check`, not the additional and different `cmake-lint`, but shouldn't be solved in this PR.

### DIFF
--- a/.github/workflows/PR_cmake-format_test.yml
+++ b/.github/workflows/PR_cmake-format_test.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   pull_request_target:
     branches-ignore:
-    - master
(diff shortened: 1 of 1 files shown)


berquist:
It's failing because the action run in the PR is not using the branch's updated `PR_clang-format_test.yml`, rather the broken one that was merged to `devel`.

jpkenny:
@berquist should we be concerned about the cmake-format test?

PR #974, opened 2023-07-17
Allow generic debug flags

berquist:
Fixes https://github.com/sstsimulator/sst-elements/issues/1558

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -139,8 +139,8 @@ AC_DEFINE_UNQUOTED([SST_INSTALL_PREFIX], ["$prefix"], [Defines the location SST 
 AC_CACHE_SAVE
 
 dnl Remove flags like -g, -O, and -W, that are not needed by other elements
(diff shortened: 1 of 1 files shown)


berquist:
Something is going on in the cmake-format action even though I pinned the version :(

berquist:
> Something completely unrelated is going on in the cmake-format action even though I pinned the version :(

The fix to GHA requires https://github.com/puneetmatharu/cmake-format-lint-action/pull/6 and then bumping the action version.

berquist:
https://github.com/puneetmatharu/cmake-format-lint-action/pull/6 is merged; I need to bump the action version then rebase this PR.

berquist:
Explanation of what this does:

`SST_EXPORT_*FLAGS` are so that elements can be build with the same relevant `CFLAGS` and `CXXFLAGS` as core. For debugging, previously only you could compile with `-g` and have that passed on properly. If you compiled with `-g3` or `-ggdb`, core would compile fine, but `SST_EXPORT_*FLAGS` would get mangled and elements wouldn't compile.  This fixes the regex to stop this mangling and also isolates it in a function so it doesn't appear more than once.

feldergast (approved):

PR #977, opened 2023-07-24
Extended help

feldergast:


src/sst/core/config.cc
```diff
@@ -679,14 +727,14 @@ Config::insertOptions()
...
+    DEF_ARG_EH(
         "enable-profiling", 0, "POINTS",
         "Enables default profiling for the specified points.  Argument is a semicolon separated list specifying the "
         "points to enable.",
-        std::bind(&ConfigHelper::enableProfiling, this, _1), true);
+        std::bind(&ConfigHelper::enableProfiling, this, _1), std::bind(&ConfigHelper::getProfilingExtHelp), true);
     DEF_ARG(
         "profiling-output", 0, "FILE", "Set output location for profiling data [stdout (default) or a filename]",
-        std::bind(&ConfigHelper::setProfilingOutput, this, _1), true);
+        std::bind(&ConfigHelper::setProfilingOutput, this, _1), true)
```
jpkenny, line 737: inconsistencies in using ";" after macro bothers me a bit stylistically, but I don't suppose actually hurts anything

jpkenny (approved):
Looks good

src/sst/core/config.cc
feldergast: That was just a missed semicolon.  I used semicolons here since the macro calls are intermixed with regular function calls and it makes the auto-indentation happier if the semicolons are present. I'll fix it now since there is something else going on with the autotester right now so I can't get it merged until that is fixed.

berquist:
> Status Flag 'Pull Request AutoTester' - Failure: Timed out waiting for job SST__AutotestGen2_NewFW_sst-test_OMPI-4.1.4_PY3.6_sst-elements to start: Total Wait = 303
> 
>     * Other jobs have been previously started - We must stop them...

I don't like this behavior...it's only because there's a running build for this particular Jenkins job (that I started manually).

jpkenny (approved):
Looks good

PR #983, opened 2023-09-01
Fix filesystem operations when unregistering elements

### LINKED ISSUE #981: Edge cases when unregistering elements
> We see one or more of the following issue:
> - When unregistering a component (via sst-register -u), the component still resides in the ~/.sstsimulator/ configuration file.  We wrote a manual script that we run with `make clean` that uses sed to remove it by force. (see https://github.com/tactcomplabs/rev/blob/master/scripts/uninstall.sh)
> - When registering a component, we sometimes see the registration appear multiple times in the configuration file.  This generally happens when we pull multiple branches of the same component source tree, build and install each branch w/o uninstalling manually.  EG, sst-register doesn't appear to catch collisions of the same component name w/ different lib paths.I wasn't made aware of how to reproduce this.

berquist:
Closes #981

@jleidel See if you're still having trouble with this patch.


berquist:
Apparently support for `filesystem` is annoying with GCC 8.  I will see if switching to the old POSIX layer calls works instead of adding a bunch of logic to the build system to support `filesystem`.

jleidel:
> Apparently support for `filesystem` is annoying with GCC 8. I will see if switching to the old POSIX layer calls works instead of adding a bunch of logic to the build system to support `filesystem`.

Since GCC8 only has partial support for C++17, you have to do `--std=c++17 -lstdc++fs`.  We have moved to building with GCC-11 for this exact reason.  Annoying, I know...

berquist:
@jleidel Sorry for the runaround, can you take a look now that it passes all our testing?

jleidel:
> @jleidel Sorry for the runaround, can you take a look now that it passes all our testing?

No worries!  Testing now...

jleidel:
> @jleidel Sorry for the runaround, can you take a look now that it passes all our testing?

Fix is confirmed!

feldergast (approved):

PR #1009, opened 2023-10-19
Interactive SST-Info

bliu1013:
Add interactive mode for SST-Info, activated by flag `-i` / `--interactive`.


feldergast:
For core PRs, we prefer that you do a rebase against the latest devel branch instead of using merges. Also, unless there is some reason to preserve all the commits, it's best practice to squash them all into a single commit.  We can do this as part of merging the PR, but if you need to rebase already, it's easy to do the squash locally as well. Typically we use an interactive rebase (git rebase -i devel) to do both of these at the same time. I'll contact you on how best to do this.

feldergast:
Also, just noticed that you did a merge from sstsimulator:master, which could cause strange artifacts in devel if we merge this.  I'm going to close this for now until we can investigate further.

PR #1010, opened 2023-10-19
String format specifiers for typedefs

berquist:
These defines are used like `PRIu64` etc. for typedefs in elements.

There's also a fix for an unused return value.

### DIFF
--- a/src/sst/core/statapi/statgroup.cc
+++ b/src/sst/core/statapi/statgroup.cc
@@ -50,7 +50,7 @@ bool
 StatisticGroup::containsStatistic(const StatisticBase* stat) const
 {
     if ( isDefault ) return true;
-    std::find(stats.begin(), stats.end(), stat);
+    std::ignore = std::find(stats.begin(), stats.end(), stat);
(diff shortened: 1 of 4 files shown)


src/sst/core/statapi/statgroup.cc
```diff
@@ -50,7 +50,7 @@ bool
 StatisticGroup::containsStatistic(const StatisticBase* stat) const
 {
     if ( isDefault ) return true;
-    std::find(stats.begin(), stats.end(), stat);
+    std::ignore = std::find(stats.begin(), stats.end(), stat);
```
feldergast, line 53: I assume this was added to squash a warning, but the warning actually revealed a bug.  The code should check the return value of the find and return the proper bool.  So, something like this:

if ( std::find(stats.begin(), stats.end(), stat) != stats.end() ) return true;

We need to fix this instead of suppressing the warning.

feldergast (changes requested):

feldergast (approved):

PR #1011, opened 2023-10-19
Add interactive sst-info

bliu1013:
Adding first version of interactive SST-info, activated using flag `-i` / `--interactive`


bliu1013:
Not sure why it's duplicating commits here, I ran:

`git fetch upstream`
`git rebase upstream/devel`
`git push --force`

Any suggestions on how to fix this?

feldergast (changes requested):
Generally things look good, but we need to bracket all the new code in #ifdefs that ensure that ncurses is installed.  That way if there's a machine that doesn't have it installed for some reason, sst will still compile.

bliu1013:
@feldergast I've added in the #ifdefs and the error messages from the other PR. However, do you know if the code would actually configure and compile correctly without curses due to the  changed code in `configure.ac`? I'm not too familiar with how autotools works, but it seems to me that either the AX_WITH_CURSES macro or the linker flag could throw an error anyways:

```
AX_WITH_CURSES
if test "x$ax_cv_ncursesw" != xyes && test "x$ax_cv_ncurses" != xyes; then
    AC_MSG_ERROR([requires either NcursesW or Ncurses library])
fi
LDFLAGS="$LDFLAGS -lncurses"
```

bliu1013:
Going to go ahead and close this PR and try again

PR #1021, opened 2023-12-12
Add interface for accessing CustomReq/Resp data

bwhitchurch:
This is related to #1001. Introduces an interface so we can leapfrog sst-core and sst-elements to ensure builds continue to pass as we deal with this pernicious memory leak.

### DIFF
--- a/src/sst/core/interfaces/stdMem.h
+++ b/src/sst/core/interfaces/stdMem.h
@@ -914,6 +914,8 @@ public:
             return str.str();
         }
 
+        CustomData* getData() { return data; }
+
         CustomData* data; /* Custom class that holds data for this event */
         Addr        iPtr; /* Instruction pointer */
         uint32_t    tid;  /* Thread ID */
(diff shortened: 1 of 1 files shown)


gvoskuilen (approved):
Looks good to me as a first step. Once this merges, I'll look at the elements PR.

bwhitchurch:
I hate to keep bumping this PR, but I think it may be necessary to add a `releaseData()` (just a thunk of `unique_ptr::release()`) interface to handle ownership transfers from request to response (happens in MemH). This could be done by directly accessing the `data` member, but it seems like encapsulating it in a method call would make future development more streamlined. Thoughts?

gvoskuilen:
Yes, a function to let ownership be explicitly passed between CustomReqs/Resps would be better than manually setting data to nullptr. Go ahead and add it to this PR.

bwhitchurch:
@gvoskuilen i will add that in. Can we mark this WIP for the mean time? I am travelling so it might be a couple days before I'm able to get it done.

gvoskuilen:
@bwhitchurch Thanks!

gvoskuilen (approved):
Looks good, thanks!

PR #1027, opened 2024-01-08
Add SSTConfigVersion.cmake to allow cmake find_package with version

beardhatcode:
With this commit SST can be found with version restrictions in CMake

    find_package(SST 13 REQUIRED)

Before this commit cmake would error with:

    Could not find a configuration file for package "SST" that is compatible
    with requested version "13".

    The following configuration files were considered but not accepted:

    /lib/cmake/SST/SSTConfig.cmake, version: unknown

(The code of this change was made by one of my colleagues, who allowed me to make a PR for this.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -153,6 +153,7 @@ AC_CONFIG_FILES([
   doc/Doxyfile
   share/Makefile
   share/SSTConfig.cmake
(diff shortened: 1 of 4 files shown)


jmlapre (approved):
Looks good.

PR #1029, opened 2024-01-09
Time fault

feldergast:
Adds a check to make sure time does not go backwards within a simulation object.  It will call fatal() if it detects a time fault. The most likely cause is actually that the 64-bit core time variable overflowed.

Also includes a change to serialize threads in creating Links because interleaving the creation of Links leads to very long deletion times when there are a large number of links.


jpkenny (approved):
Looks good to me

PR #1043, opened 2024-01-30
Use exceptions for UnitAlgebra errors

### LINKED ISSUE #1037: UnitAlgebra should raise exceptions rather than call `abort.fatal`
From https://github.com/sstsimulator/sst-core/blob/962d589c3b7dc7398d2de360915ebf833aa2577e/tests/test_UnitAlgebra.py#L153

```python
# Check to make sure we get exceptions for invalid arguments.  Since
# it is the same function, really only need to check one operator.
# Need to check using the UnitAlgebra as both the first and second
# argument.  Also need to check equals and not equals separately,
# because they can fall back on just checking to see if they are the
# exact same object, in which case, then won't generate an error.
```

berquist:
Closes #1037

Here is a proposal for adding exceptions to the core public API.  Doxygen is missing until we decide on the form of exceptions.

Thoughts:
- Since these exceptions are specific for `UnitAlgebra`, I've added them inside the `UnitAlgebra` class rather than just the header.
- There is a base `UnitAlgebraException` in case a consumer (elements, some other code) wants to catch a more generic exception type rather than one of the more specific types.
- The base `UnitAlgebraException` inherits from `logic_error` rather than `exception` since these are fundamentally not runtime errors.
- As a development guideline, it seems like reasons/messages given by C++ exceptions should be actionable directly Python side: the message passed to `PyErr_SetString` should always be `exception_instance.what()`.

For the tests, I used `assert` statements rather than the existing `correct_throw` idiom.

### DIFF
--- a/src/sst/core/model/python/pymodel_unitalgebra.cc
+++ b/src/sst/core/model/python/pymodel_unitalgebra.cc
@@ -45,7 +45,13 @@ unitAlgebraInit(UnitAlgebraPy_t* self, PyObject* args, PyObject* UNUSED(kwds))
     PyErr_Clear();
     if ( PyArg_ParseTuple(args, "O", &str_obj) ) {
         PyObject* vstr = PyObject_CallMethod(str_obj, (char*)"__str__", nullptr);
-        self->obj      = SST_ConvertToCppString(vstr);
+        try {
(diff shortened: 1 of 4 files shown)


feldergast (approved):
Looks good.  This will be a nice addition.

PR #1052, opened 2024-03-05
Update curses detection for Autotools and CMake

berquist:
* For Autotools, switch to using `ncurses6-config` (like `python-config`) to get library and include locations.  This should fix any issues where a user-specified or custom curses install may conflict with a system install.
* Add CMake support for linking against curses.
* For my Spack-provided curses, it adds `-D_XOPEN_SOURCE` to the compile line, and this conflicts with the one present in `Python.h`.  Rather than trying to handle this in the build system it was easier to silence it in the one source file that complains.


feldergast (approved):
Looks good based on my limit m4 knowledge.

PR #1064, opened 2024-03-22
Remove Python 2 testing infrastructure

berquist:
This removes all of the testing infrastructure that tries to distinguish between Python 2 and 3.  It also removes the minimum Python version check in favor of the one performed by Autotools at configure time.

### DIFF
--- a/src/sst/core/testingframework/sst_test_engine_loader.py
+++ b/src/sst/core/testingframework/sst_test_engine_loader.py
@@ -33,50 +33,7 @@ TESTFRAMEWORKSFILES = ["sst_unittest.py", "sst_unittest_support.py",
 TEST_ELEMENTS = 0
 TEST_SST_CORE = 1
 
-REQUIRED_PY_MAJ_VER_2 = 2               # Required Python2 Major Version
-REQUIRED_PY_MAJ_VER_2_MINOR_VER = 7     # Required PY2 Minor Version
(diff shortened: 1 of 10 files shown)


feldergast (approved):
Looks good. I am a bit concerned that we may need to check minor versions again at some point since Python seems to make some major changes between minor versions, but I think we can add those back in if they become needed.

berquist:
My reason for removing the checks Python side is that I assume
- incompatibilities are more likely to pop up when compiling against CPython, rather than at test time, and
- the likelihood of using a different Python for testing than the one compiled against is low, and this different version would cause a problem during testing that didn't appear during compilation.

If either of these are bad assumptions, then we should not make this change.

feldergast:
> My reason for removing the checks Python side is that I assume
> 
> * incompatibilities are more likely to pop up when compiling against CPython, rather than at test time, and
> * the likelihood of using a different Python for testing than the one compiled against is low, and this different version would cause a problem during testing that didn't appear during compilation.
> 
> If either of these are bad assumptions, then we should not make this change.

I think your reasoning is sound.  This is definitely more of a problem with the Cpython, but even the last changes with strings broke things.  Thankfully there were ways to change the code to work with all versions we currently support, but there is definitely a change, even if a small one, that we may not be so lucky next time.  I think the odds are low enough that we can remove the checks and just add them back in later.  The exact checks we may someday have to make in the code will be different than the ones you removed anyway.

PR #1071, opened 2024-04-24
Interactive SST-Info update

bliu1013:
Update 'find' function, add error checking, and add bold/highlight


jpkenny:
This probably still fails due to using new[] to allocate a multidimensional array. The discussion here outlines two different ways to do this (the "pointer" approach being the closest to how the code currently works).
https://cplusplus.com/forum/beginner/251637/

bliu1013:
> @bliu1013 This probably still fails due to using new[] to allocate a multidimensional array. The discussion here outlines two different ways to do this (the "pointer" approach being the closest to how the code currently works). https://cplusplus.com/forum/beginner/251637/

Looks like it worked anyways, but yeah I think just doing a std::vector<std::vector> would be more C++-ey

gvoskuilen (approved):

PR #1075, opened 2024-05-09
Initial checkpointing support

feldergast:
Initial checkpointing support

- Currently only supports serial runs
- Updates to serialization to support checkpointing
  - Added ability to turn on pointer tracking
  - Added ability to serialize non-pointer variables, but track address of data as a pointer.
- Changed templating for handlers to support checkpointing
  - Function pointer is now a non-type template parameter
- Added additional serialization tests and new checkpointing tests
- Added a function to ConfigBase so you could query whether or not an option was set on the command line.

Other incidental changes:

- Remove old references to SharedRegion
- Removed type checking in testing comparison filters.  This was done because the filter is called for every line of output and the type checks apparently aren't very efficient and were a performance bottleneck for tests with a lot of output
- Added Python Module output to sst-info
- Added fflush to all output in TraceFunction


jmlapre (dismissed):
Everything looks OK.  I would maybe add a comment here to make it clear that checkpointing is only supported within a single machine or at least not across two different architectures.

jmlapre (dismissed):

jpkenny (changes requested):
Found some typos and style complaints. No major technical issues that I see (it's a lot of code).

jmlapre (approved):
quick review

PR #1092, opened 2024-06-17
Update checkpoint branch - statistics

gvoskuilen:
- Add checkpoint support for statistics 
 - HDF5 not supported yet
- NotSerializable macro no longer required for classes that inherit from a class that uses ImplementVirtualSerializable
- Output no longer inherits from serializable
- Minor general clean up in statapi classes
- Fixed issue in JSON statistic output where the outputsimtime parameter had no effect
- Fixed issue where params were serializable but not checkpointable due to missing keysets in checkpoint


feldergast (approved):
Looks good for an intermediate check-in.  Will need to clean up debug output at a later date.

PR #1093, opened 2024-06-20
Modified TraceFunction object to require that an environment variable…

feldergast:
… be set to enable output.  This makes it possible to leave the TraceFunction calls in while developing, but still be able to run and pass testing.

### DIFF
--- a/src/sst/core/output.cc
+++ b/src/sst/core/output.cc
@@ -24,6 +24,7 @@
 #include <atomic>
 #include <cerrno>
 #include <cinttypes>
+#include <cstdlib>
 #include <cstring>
 #include <string>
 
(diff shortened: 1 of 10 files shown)


src/sst/core/output.cc
```diff
@@ -567,18 +579,72 @@ TraceFunction::output(const char* format, ...) const
...
+    va_start(args, format);
+    size_t n = vsnprintf(buf, 200, format, args);
+    va_end(args);
+
+    if ( n >= 200 ) {
+        // Generated string longer than buffer
+        delete[] buf;
+        buf = new char[n + 1];
+        va_start(args, format);
+        vsnprintf(buf, 200, format, args);
```
gvoskuilen, line 598: Should this be 'n' instead of 200?

gvoskuilen (dismissed):

gvoskuilen (approved):

gvoskuilen (approved):

PR #1107, opened 2024-07-24
Reorganize serialization code in preparation for larger changes.

feldergast:


src/sst/core/serialization/serialize_serializable.h
gvoskuilen, line 1: Does serialize_serializable.h need to include serializable.h so that the functionality doesn't break or is that already included implicitly somewhere?

feldergast: In general, files should not have been including serialize_serializable.h since it was already being included by serializable.h.  We should be okay not including it.

gvoskuilen (approved):

PR #1114, opened 2024-07-31
Remove redundant initializeTestModule_SingleInstance() methods

jmlapre:
The implementation of `initializeTestModule_SingleInstance()` is essentially the same across all of the `SSTTestCase` subclasses.  There's no need to re-implement them in each class.  We can cut ~15 lines from each of the 19 `testsuite_default_*.py` files.

### DIFF
--- a/src/sst/core/testingframework/sst_unittest.py
+++ b/src/sst/core/testingframework/sst_unittest.py
@@ -53,15 +53,6 @@ class SSTTestCase(unittest.TestCase):
         This class is derived from Python's unittest.TestCase which provides an
         basic resource for how to develop tests for this frameworks.
     """
-    module_init = 0
-    module_sema = threading.Semaphore()
-
-    @classmethod
-    def initializeTestModule_SingleInstance(cls):
-        with cls.module_sema:
-            if cls.module_init != 1:
-                # Put your single instance Init Code Here
-                cls.module_init = 1
 
     def __init__(self, methodName):
         # NOTE: __init__ is called at startup for all tests before any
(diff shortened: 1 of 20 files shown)


src/sst/core/testingframework/sst_unittest.py
```diff
@@ -48,6 +48,15 @@ class SSTTestCase(unittest.TestCase):
         This class is derived from Python's unittest.TestCase which provides an
         basic resource for how to develop tests for this frameworks.
     """
+    module_init = 0
+    module_sema = threading.Semaphore()
+
+    @classmethod
```
feldergast, line 54: @gvoskuilen  and I discussed this and we think that since this function doesn't do anything useful we should either remove it or have it be empty and have documentation on how to properly do the semaphore if it's needed.  I would only leave the empty function if it's required to avoid errors.

feldergast:
Looks like this has conflicts with what's in devel, so you'll need to rebase against current devel and force push the branch. Please don't do a merge.

jpkenny:
@jmlapre:
I don't know the details of the testing framework enough.
I agree the code is redundant, but shouldn't it be moved to the parent class then? I feel like that "module_init = 1" probably has to happen somewhere and it looks like this changeset would eliminate it.
Probably @feldergast needs to review this.

feldergast:
> @jmlapre: I don't know the details of the testing framework enough. I agree the code is redundant, but shouldn't it be moved to the parent class then? I feel like that "module_init = 1" probably has to happen somewhere and it looks like this changeset would eliminate it. Probably @feldergast needs to review this.

I think it's fine if module_init doesn't get set.  It only seems to be used with the semaphore to make sure the code only executes once.  No other part of the testing framework access a variable named module_init.  If a test suite needs to have code that executes only once, then they would need to add in the code that was removed and then add their execute once code in the designated place. We'll need to make sure that we have documentation on what should be added to the test suite in this case. I'm sure this was just part of whatever code template was originally created and didn't have a good comment about when the code was actually needed and that it could be removed if it wasn't.

feldergast (approved):

PR #1126, opened 2024-09-06
Update and pin version of clang-format-lint-action

berquist:
This fixes the error in the `clang-format` part of CI:
```
  File "/run-clang-format.py", line 25, in <module>
    from distutils.util import strtobool
ModuleNotFoundError: No module named 'distutils'
```
See https://github.com/kokkos/kokkos/pull/7295 for another example.

This is not the version of `clang-format` used, which is specified by `clangFormatVersion: 12` (see line 75).

All PRs on core will fail until this is merged.


berquist:
Wrong base branch, see https://github.com/sstsimulator/sst-core/pull/1127

PR #1127, opened 2024-09-06
Update and pin version of clang-format-lint-action

berquist:
This fixes the error in the `clang-format` part of CI:
```
  File "/run-clang-format.py", line 25, in <module>
    from distutils.util import strtobool
ModuleNotFoundError: No module named 'distutils'
```
See https://github.com/kokkos/kokkos/pull/7295 for another example.

This is not the version of `clang-format` used, which is specified by `clangFormatVersion: 12` (see line 75).

All PRs on core will fail until this is merged.

Redo of https://github.com/sstsimulator/sst-core/pull/1126

### DIFF
--- a/.github/workflows/PR_clang-format_test.yml
+++ b/.github/workflows/PR_clang-format_test.yml
@@ -67,7 +67,7 @@ jobs:
 
 #   Running the github action for clang format
     - name: Run Action clang-format-lint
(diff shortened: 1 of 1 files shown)


berquist:
See https://github.com/berquist/sst-core/actions/runs/10743566168/job/29798572611 for an example of this passing.

gvoskuilen (approved):

PR #1136, opened 2024-09-25
deprecate testing_is_PIN2_used()

jmlapre:
We are deprecating `testing_is_PIN2_used()` at some future release.  This PR simply adds a warning to any code that calls this function.


jmlapre:
Closes https://github.com/jmlapre/sst-core/issues/6

PR #1137, opened 2024-09-25
deprecate simple host_os_get_* methods

jmlapre:
`host_os_get_*()` : Eliminate the ones that are just wrapping a Python function in favor of the Python function. Most tests at most have to make a distinction between osx & linux. Occasionally Ubuntu vs RHEL needs to be treated differently but only w.r.t. Pin in my experience.


jmlapre:
Closes https://github.com/jmlapre/sst-core/issues/7

gvoskuilen (approved):

PR #1148, opened 2024-10-02
make helper function internal

jmlapre:
`testing_check_is_scenario_filtering_enabled()` should be internal. It's only used in `skip_on_scenario()`

I feel this change is small and localized enough we don't need to deprecate it first.  There are no references to it anywhere besides its usage in `skip_on_scenario()`.  Thoughts?  @gvoskuilen @feldergast 

Closes https://github.com/jmlapre/sst-core/issues/8

### DIFF
--- a/src/sst/core/testingframework/sst_unittest_support.py
+++ b/src/sst/core/testingframework/sst_unittest_support.py
@@ -361,7 +361,7 @@ def host_os_get_num_cores_on_system():
 # SST Skipping Support
 ################################################################################
 
-def testing_check_is_scenario_filtering_enabled(scenario_name):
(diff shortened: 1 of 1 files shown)


feldergast:
> `testing_check_is_scenario_filtering_enabled()` should be internal. It's only used in `skip_on_scenario()`
> 
> I feel this change is small and localized enough we don't need to deprecate it first. There are no references to it anywhere besides its usage in `skip_on_scenario()`. Thoughts? @gvoskuilen @feldergast
> 
> Closes [jmlapre#8](https://github.com/jmlapre/sst-core/issues/8)

I agree that you probably don't need to deprecate.

gvoskuilen:
I agree

gvoskuilen (approved):

Issue #987, opened 2023-09-13
feature request: Python module with stubs and documentation for IDEs


researcherben:
## Detailed description of enhancement

Python module of stubs with documentation to facilitate use of VSCode and other IDEs when writing Python driver files.

As an example, how should a user of SST know what is hidden in the Python module referenced in this file?

<https://github.com/sstsimulator/sst-elements/blob/master/src/sst/elements/simpleSimulation/tests/test_simpleCarWash.py>

A user might look at <http://sst-simulator.org/sst-docs/> and find <http://sst-simulator.org/sst-docs/docs/config/general/setProgramOption>

This feature request is for the same information to be made available in IDEs like VSCode by providing a Python module with stubs and documentation.

## Provide sha1 of all relevant sst repositories (sst-core, sst-elements, etc)

This could be distributed as part of sst-core.

## Fill out Labels, Milestones, and Assignee fields as best possible

I don't have permissions to set the assignee, label, or milestone for this issue.

PR #1151, opened 2024-10-03
Add initial set of Python type annotations

berquist:
In the test framework, the use of custom classes to extend `unittest` (and the use of `unittest` in general) makes it difficult to understand what objects are getting passed around.  This PR adds an initial set of (incomplete) type annotations and a `mypy` configuration.  `mypy` will run type checking just like a compiler for functions and methods that have their signatures annotated with types.

At least for the test framework, this is also part of https://github.com/sstsimulator/sst-core/issues/987.

This PR does not enable checking the type annotations with `mypy` via `pre-commit`, since the annotations are incomplete and may still be wrong in a few places while they are being added.  It's easier to iterate on the annotations and the configuration rather than do everything all at once.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,19 @@
+[tool.mypy]
+explicit_package_bases = true
+mypy_path = "$MYPY_CONFIG_FILE_DIR/src/sst/core/testingframework"
+
+warn_unused_ignores = true
+
+warn_return_any = true
+warn_unused_configs = true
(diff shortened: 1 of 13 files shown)


gvoskuilen (approved):

berquist:
Waiting for the Autotester to make a pass before rebasing.

gvoskuilen (approved):

PR #1152, opened 2024-10-04
Fix call to os.path.isfile()

jmlapre:
`os.path.isfile("/etc/rocky-release", " ")` is incorrect and shouldn't have the additional `" "` passed to it.


jmlapre:
Closes https://github.com/jmlapre/sst-core/issues/22

gvoskuilen (approved):

PR #1167, opened 2024-10-24
Create README.md

researcherben:
content is an explanation from @feldergast

### DIFF
--- a/src/sst/core/profile/README.md
+++ b/src/sst/core/profile/README.md
@@ -1,0 +1,12 @@
+Profiling experimental, see the output of `sst -h`
+```bash
+Advanced Options - Profiling (EXPERIMENTAL)
+   --enable-profiling=POINTS      (S) Enables default profiling for the specified points.  Argument is a semicolon separated list specifying the points to enable.
+   --profiling-output=FILE        (S) Set output location for profiling data [stdout (default) or a filename]
+```
+
+Profiling is "supported" in the context that it is not going away and the SST team will respond to bug reports.  It is not necessarily in its final form and some aspects of it may change before it is finalized.
+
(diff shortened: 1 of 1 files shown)


feldergast:
The following information is available by running:
  sst --help=enable-profiling

Profiling Points [EXPERIMENTAL]:

NOTE: Profiling points are still in development and syntax for enabling profiling tools, as well as available profiling points is subject to change.  However, it is intended that profiling points will continue to be supported into the future.

  Profiling points are points in the code where a profiling tool can be instantiated.  The profiling tool allows you to collect various data about code segments.  There are currently three profiling points in SST core:
   - clock: profiles calls to user registered clock handlers
   - event: profiles calls to user registered event handlers set on Links
   - sync: profiles calls into the SyncManager (only valid for parallel simulations)

  The format for enabling profile point is a semicolon separated list where each item specifies details for a given profiling tool using the following format:
   name:type(params)[point]
     name: name of tool to be shown in output
     type: type of profiling tool in ELI format (lib.type)
     params: optional parameters to pass to profiling tool, format is key=value,key=value...
     point: profiling point to load the tool into

Profiling tools can all be enabled in a single instance of --enable-profiling, or you can use multiple instances of --enable-profiling to enable more than one profiling tool.  It is also possible to attach more than one profiling tool to a given profiling point.

Examples:
  --enable-profiling="events:sst.profile.handler.event.time.high_resolution(level=component)[event]"
  --enable-profiling="clocks:sst.profile.handler.clock.count(level=subcomponent)[clock]"
  --enable-profiling=sync:sst.profile.sync.time.steady[sync]

I think this covers the information and it will be updated as new profile points are added and if anything changes in the syntax.  We are anticipating that this will become a fully supported feature in SST 15.

feldergast:
Closing.  Information is available using sst --help.  Also, profiling will no longer be considered experimental as of SST 15.

Issue #1169, opened 2024-10-29
Add [[noreturn]] attribute to SST::Output::fatal()


leekillough:
**Describe the solution you'd like**
Add `[[noreturn]]` attribute to `SST::Output::fatal()`

`[[noreturn]]` allows greater optimization and avoids warnings about missing return values in functions which call `SST::Output::fatal()` but don't `return` values.

**Describe alternatives you've considered**
Add dummy `return` after calls to `SST::Output::fatal()` to avoid warnings

feldergast:
I think this makes a lot of sense, but the fatal() function as written is not actually no return.  There are corner cases in threaded runs where two threads calling fatal() at the same time will cause the one that reaches the end of the function last to return (you can't have MPI_abort()/exit() called more than once).  If you'd like to put in a PR to add the attribute, then you'll probably need to throw an exception at the end of the function to actually make the function no return.

leekillough:
Speaking of handling race conditions, I don't know if the current code uses `call_once()`, but a fast and simple method is local static initialization (the so-called Meyers Singleton) at the point of exit:

```c++
static int exit_once = (MPI_exit(), 0);
std::terminate();
```

That static initialization can be placed at the point of the exit in any part of a function, and only one thread will call it. If it releases the lock and other threads continue, they will call `std::terminate()`. But it probably will never release the lock before `MPI_exit()` is called and (never) returns.

leekillough:
If there are several places in SST which may call `MPI_exit()`, then the above code can be placed in a wrapper to guard against race conditions, e.g.:
```c++
[[noreturn]] void SST_exit() {
  static int exit_once = (MPI_exit(), 0);
  std::terminate();
}
```
The API can be changed if exiting with a specific exit code needs to be supported.

PR #1175, opened 2024-11-09
Add [[noreturn]] attribute to Output::fatal, add SST_Exit() function

leekillough:
Fixes https://github.com/sstsimulator/sst-core/issues/1169 by adding `[[noreturn]]` to `SST::Output::fatal()`.

Adds a `SST::SST_Exit()` function which ensures only a single thread exits if multiple threads call it simultaneously.

### DIFF
--- a/src/sst/core/exit.cc
+++ b/src/sst/core/exit.cc
@@ -12,6 +12,7 @@
 #include "sst_config.h"
 
 #include "sst/core/exit.h"
+#include <exception>
 
 #include "sst/core/warnmacros.h"
 #ifdef SST_CONFIG_HAVE_MPI
@@ -204,6 +205,25 @@ Exit::serialize_order(SST::Core::Serialization::serializer& ser)
     ser& m_idSet;
     ser& end_time;
     ser& single_rank;
+}
+
+[[noreturn]] void
+SST_Exit(int exit_code)
+{
+    // Make sure only one thread calls MPI_Abort() or exit() in the
+    // case where two threads call fatal() at the same time
+    // Only one thread initializes the function-local static variable
+    // Other threads are blocked
+
+#ifdef SST_CONFIG_HAVE_MPI
+    // If MPI exists, abort
+    static int exit_once = (MPI_Abort(MPI_COMM_WORLD, exit_code), 0);
+#else
+    static int exit_once = (exit(exit_code), 0);
+#endif
+
+    // Should never get here
+    std::terminate();
 }
 
 } // namespace SST
--- a/src/sst/core/exit.cc
+++ b/src/sst/core/exit.cc
@@ -12,7 +12,6 @@
 #include "sst_config.h"
 
 #include "sst/core/exit.h"
-#include <exception>
 
 #include "sst/core/warnmacros.h"
 #ifdef SST_CONFIG_HAVE_MPI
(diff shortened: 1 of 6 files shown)


src/sst/core/exit.cc
```diff
@@ -12,6 +12,7 @@
 #include "sst_config.h"
 
 #include "sst/core/exit.h"
+#include <exception>
```
feldergast, line 15: Per the coding standards, includes of standard headers should go after all sst includes.  This is probably contributing to the clang-format failure.

src/sst/core/exit.h
```diff
@@ -117,6 +117,8 @@ class Exit : public Action
     bool single_rank;
 };
 
+[[noreturn]] void SST_Exit(int exit_code);
```
feldergast, line 120: Not sure if exit.h is the best place for this function to live.  One of the issues is that exit.h is being installed, though it really shouldn't be, and SST_Exit() shouldn't be part of the public API.  The other issue is that even though the names match, the functionality isn't really related to the Exit object.  Of the header files that aren't installed, I think simulation_impl.h is probably the best place for the SST_Exit() function (not as part of the Simulation_impl object, however).  Since The function won't be part of the non-core API, I think it's fine to leave as is for now if you don't want to change it and we can reevaluate later.

feldergast (commented):
Thanks for putting this together.  In general, things look good.  Just need to address the clang-format errors and we can get it released for testing.

src/sst/core/exit.cc
leekillough: I ran `clang-format` but my version is newer than v12 and I didn't feel like trying to go back to an older version.

src/sst/core/exit.h
leekillough: Okay I can move it anywhere, to prevent it from being publicly exported / documented. The `exit.h` and `exit.cc` seemed logical, even if this function is unrelated to the Exit object.

leekillough: Oh, and since `Exit` is a common name which might conflict with other uses, I used `SST::SST_Exit` instead of `SST::Exit`. If this deviates from normal SST naming conventions, such as case and underscores, it can be renamed. In Rev we use `camelCase` for variables / functions and `CamelCase` for types.

src/sst/core/exit.cc
feldergast: That's not promising for moving clang-format forward if it doesn't follow the header ordering settings in the .clang-format file.  You can look at the diff of what it got with what it wanted in the "checks" tab in the PR.  Just select the clang-format check and click in the Matrix box to bring up the results of the test.

leekillough: It's too late to fix it now, but `clang-format` should have a `version 1.2.3` directive for `.clang-format` files so that it behaves like a certain version. But as the number of versions increases, this becomes hard to test, and complicates the code to behave differently for different versions, although if this versioning scheme is done from the beginning, the versioned tests can be developed incrementally.

src/sst/core/simulation_impl.h
```diff
@@ -603,6 +604,9 @@ void wait_my_turn_start(Core::ThreadSafe::Barrier& barrier, int thread, int tota
 
 void wait_my_turn_end(Core::ThreadSafe::Barrier& barrier, int thread, int total_threads);
 
+// Function to exit, guarding against race conditions if multiple threads call it
```
feldergast, line 607: Looks like this function needs to be hoisted above the uses of it in the same file.  This is causing all the testing builds to fail.

leekillough:
I'll investigate the failure. Maybe the `simulation_impl.h` file is not being included in the `Output` files.

I don't mind taking this slowly. I want to make sure it is safe.

feldergast:
Just looked at the log.  It appears that mpi.h is not included in simulation.cc.

leekillough:
> Just looked at the log. It appears that mpi.h is not included in simulation.cc.

Should we conditionally include `mpi.h` based on the MPI macro?

feldergast:
> > Just looked at the log. It appears that mpi.h is not included in simulation.cc.
> 
> Should we conditionally include `mpi.h` based on the MPI macro?

Yes.  We generally do this for including mpi.h:

#ifdef SST_CONFIG_HAVE_MPI
DISABLE_WARN_MISSING_OVERRIDE
#include <mpi.h>
REENABLE_WARNING
#endif

The extra macros are to suppress some of the known warnings in openmpi's version of mpi.h.

PR #1176, opened 2024-11-12
Fix type annotations in sst_unittest_support.py

berquist:
There are also a few examples of how annotations would look in the test suite itself.

There is an unfortunate violation of the signature of one of the comparison functions by https://github.com/sstsimulator/sst-elements/blob/a5adba225d4262f67b5c6b09c7bfdfe5af8717cd/src/sst/elements/ember/tests/testsuite_default_ember_otf2.py#L54, so it is possible that others could be using the code this way too (passing a single filter instead of a list of filters).  Maybe adding a deprecation warning would be good so that the `isinstance` check can go away in the future.

### DIFF
--- a/src/sst/core/testingframework/sst_unittest_support.py
+++ b/src/sst/core/testingframework/sst_unittest_support.py
@@ -375,7 +375,11 @@ def _testing_check_is_scenario_filtering_enabled(scenario_name: str) -> bool:
 
 ###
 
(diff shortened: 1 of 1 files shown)


feldergast:
The filtered compares were designed to take either a list of filters or a single filter. It was not an oversight.

feldergast (approved):

PR #1179, opened 2024-12-03
Remove dead log calls and consolidate logic in PIN support functions

berquist:


src/sst/core/testingframework/sst_unittest_support.py
```diff
@@ -127,19 +127,13 @@ def testing_is_PIN_loaded() -> bool:
...
-    #log_debug("testing_is_PIN_loaded() - Intel_PIN_Path = {0}; Valid Dir = {1}".format(pin_path, pindir_found))
     return pindir_found
 
 def testing_is_PIN_Compiled() -> bool:
     global pin_exec_path
     pin_crt = sst_elements_config_include_file_get_value(define="HAVE_PINCRT", type=int, default=0, disable_warning=True)
     pin_exec = sst_elements_config_include_file_get_value(define="PINTOOL_EXECUTABLE", type=str, default="", disable_warning=True)
-    #log_debug("testing_is_PIN_Compiled() - Detected PIN_CRT = {0}".format(pin_crt))
-    #log_debug("testing_is_PIN_Compiled() - Detected PIN_EXEC = {0}".format(pin_exec))
-    pin_exec_path = pin_exec
```
feldergast, line 139: Not sure if it's needed, but I noticed that the side effect of setting the global variable pin_exec_path is not in the new code.  Was this intended?

berquist: This wasn't intended, so I either need to put it back or check in all of the core and elements Python code for its usage.

berquist:
In elements, `pin_exec_path` was only used for one thing, printing the Pin location in some test logging.  I've put it back since we decided that `sst_unittest_support.py` is public, so someone else could be using it, and since it's a variable I don't think we can attach a deprecation warning to it.  Ideally this would be replaced with a function.

I did remove the `HAVE_PINCRT` lookup, since that variable isn't global and isn't used anywhere.

feldergast (approved):

PR #1212, opened 2025-02-18
Simplify SST_ELI_COMPILED_VERSION

leekillough:
- Remove redundant `static` definition of `SST_ELI_COMPILED_VERSION` in `elibase.cc`
- Use:
  -  `std::string`, `std::to_string` and mutable `delim`iter
- instead of
  -  `std::stringstream`, `operator<<` and more complicated `bool` variable
- to create version string
- Change `#include <sstream>` to `#include <ostream>` since `std::ostream` is used but `std::stringstream` is no longer used

### DIFF
--- a/src/sst/core/eli/elementinfo.cc
+++ b/src/sst/core/eli/elementinfo.cc
@@ -15,7 +15,7 @@
 
 #include "sst/core/eli/elibase.h"
 
-#include <sstream>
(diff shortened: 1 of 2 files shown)


feldergast (approved):
Looks good.  I like the mutable delimiter pattern and may have to use that other places in the code.

PR #1218, opened 2025-02-21
Use C++17 nested namespaces; remove using namespace std

leekillough:
This changes the code to use C++17 [`Nested::Namespace::Syntax`](https://en.cppreference.com/w/cpp/language/namespace).

It also removes `using namespace std` from a few files, since that's generally bad practice.

In the random number generator, it removes redundant namespace prefixes since it is already in the `SST::RNG` namespace.

### DIFF
--- a/src/sst/core/baseComponent.cc
+++ b/src/sst/core/baseComponent.cc
@@ -919,9 +919,7 @@ BaseComponent::serialize_order(SST::Core::Serialization::serializer& ser)
     }
 }
 
-namespace Core {
-namespace Serialization {
-namespace pvt {
+namespace Core::Serialization::pvt {
 
 static const long null_ptr_id = -1;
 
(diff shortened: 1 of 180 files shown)


feldergast (approved):

feldergast (dismissed):

leekillough:
This PR will fail tests until https://github.com/sstsimulator/sst-elements/pull/2460 is merged or `SST::RNG::MersenneRNG` is exported to the global namespace in SST Core. Before this PR, SST Core imported all `SST::RNG` symbols into the global namespace, which is surely incorrect.

(It was also done incorrectly, since `using namespace SST::RNG;` appeared _before_ the `SST::RNG` symbols were defined in the header file, which means that a file which included SST Core headers could not be sure that the `using namespace SST::RNG;` at the top of a header file imported all of its symbols, since the `using namespace SST::RNG;` occurred _before_ the definitions of the symbols, which means the compiler is free to take the state of the `SST::RNG` namespace at the time of the `using namespace SST::RNG;`, ignoring any definitions in `SST::RNG` which occur afterwards.)

If a `using SSTMersenneRNG = SST::RNG::MersenneRNG;` alias needs to be added to the global namespace, it can be, but all SST Elements except Merlin currently use the full `SST::RNG::MersenneRNG` name.

feldergast (approved):

PR #1224, opened 2025-02-26
Remove compiler warnings; move all MPI stuff into sst_mpi.h

leekillough:
Remove compiler warnings
  - Use `(void)!` to ignore return values marked `[[nodiscard]]` or `__attribute__((warn_unused_result))` (the `!` is needed for GCC to ignore it since simply casting to `(void)` [is insufficient](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425))
  - Initialize `std::set` key value to its type's default value during (de)serialization to remove compiler warning about it being uninitialized
  - Suppress `-Wcast-function-type` when `#include <mpi.h>`

Move all MPI header handling to `sst_mpi.h` -- now it handles the warning suppressions, `#include <mpi.h>`, and `#define`s `UNUSED_WO_MPI()`, all depending on whether MPI is enabled or not. What used to take several lines in each file, now only requires `#include "sst/core/sst_mpi.h"`. (The name `sst_mpi.h` was chosen because naming it `mpi.h` would cause it to include itself when it `#include <mpi.h>`, since the angle bracket include paths include the `src/sst/core` directory.)


src/sst/core/sst_mpi.h
```diff
@@ -0,0 +1,37 @@
...
+// All rights reserved.
+//
+// This file is part of the SST software package. For license
+// information, see the LICENSE file in the top level directory of the
+// distribution.
+
+#ifndef SST_CORE_SST_MPI_H
+#define SST_CORE_SST_MPI_H
+
+#include "sst_config.h"
```
feldergast, line 15: sst_config.h is not allowed to be included in header files and is required to be included first in all .cc files.  It's a quirk of the auto tools generating the file without guards for multiple inclusion.

feldergast (changes requested):
Other than the inclusion of sst_config.h in sst_mpi.h, this looks good.

src/sst/core/configGraph.h
```diff
@@ -138,7 +138,11 @@ class ConfigStatistic : public SST::Core::Serialization::serializable
 
     ConfigStatistic() : id(stat_null_id) {}
 
-    ConfigStatistic(const ConfigStatistic& o) : id(o.id), params(o.params), shared(o.shared), name(o.name) {}
+    ConfigStatistic(const ConfigStatistic&) = default;
+    ConfigStatistic(ConfigStatistic&&)      = default;
+    ConfigStatistic& operator=(const ConfigStatistic&) = default;
+    ConfigStatistic& operator=(ConfigStatistic&&) = default;
+    ~ConfigStatistic() override                   = default;
```
leekillough, line 145: This prevents a `-Wdeprecated-copy` warning, which is a misleading warning name since it doesn't involve deprecated function but instead warns about using `operator=` copy-assignment on a class which has a user-defined copy-constructor, but which doesn't have a user-defined copy-assignment operator, even if a default one can be generated. The above `= default` declarations declare that default-generated ones can and _must_ be generated, or it's an error.

Since the copy constructor of `ConfigStatistic` was simply copying member-by-member from the source, and since assignment should be done in the same fashion, we simply declare them `= default` so that the member-by-member copy/move/assignment can and _must_ be generated. If anything gets in the way, such as `const` or reference members (for assignment), it is an error. The destructor here is declared `override` but still `= default` so it ensures a `virtual` base class destructor. (In C++11 and later, `= default` is preferred instead of an empty body `{}` for destructors.)

feldergast (approved):

feldergast:
Last testing failure was a timeout on a git clone command.

Issue #1165, opened 2024-10-14
checkpoint-wall-clock not recognizing simple seconds-only format


kpgriesser:
New Issue for sst-core
NOTE: If you are reporting an issue for a specific element library, please submit request in ssts-element repository.
----------------------
1 - Detailed description of problem or enhancement

Referencing the checkpoint-wall-period documentation:
`The wall time option accepts a time formatted as HH:MM:SS, MM:SS, or SS. Alternately an integer with a time unit (h, m, or s) can be given. The simulation time option accepts an integer with a time unit of s or Hz and SI prefixes are allowed.`

This works for values < 62s (and strange number) but returns a parsing error for other values. 

```
sst --checkpoint-wall-period=62s
ERROR: Argument passed to "--checkpoint-wall-period" could not be parsed. Argument = [62s]
Valid formats are:
        %H:%M:%S
        %M:%S
        %S
        %Hh
        %Mm
        %Ss
```

Perhaps documentation can be updated for this release to read just:
`The wall time option must be specified in the time format as HH:MM:SS, MM:SS, or SS. `

2 - Describe how to reproduce the issue

`sst --checkpoint-wall-period=62s`

3 - What Operating system(s) and versions 

ubuntu 22.04.5

4 - What versions of external libraries (MPI, etc.)

openmpi 4.1.2

5 - Provide sha1 of all relevant SST repositories (sst-core, sst-elements, etc)

```
commit 335de84417671898414e7f4d242c5ab109d5f963 (HEAD -> devel, origin/devel)
Author: Scott Hemmert <kshemme@sandia.gov>
Date:   Mon Oct 7 16:30:05 2024 -0600

    Bug fixes for parallel checkpointing. (#1158)
    
    - Fixes #1156 - Initial Sync interval needed to take Checkpoint time into account.
    - Fixes #1157 - Moved barrier for Exit restart to the proper place.
```
    -

gvoskuilen:
This is a limitation of using `strptime` to parse the time string and will be an issue for any command line option that accepts a wall-time. We will fix this in a future release to be more permissive. 
Function: `ConfigBase::parseWallTimeToSeconds`

PR #1231, opened 2025-03-05
Allow user to specify seconds-only wall time in either S or Ss format

kpgriesser:
This PR allows the user to specify seconds-only wall time in either S or Ss format ( e.g. >60s ). Fixes issue #1165.
ConfigBase::parseWallTimeToSeconds(...) is modified to first check if the argument is of the form S or Ss and parses the value directly. If that fails the normal path is followed.

### DIFF
--- a/src/sst/core/configBase.cc
+++ b/src/sst/core/configBase.cc
@@ -60,8 +60,8 @@ uint32_t
 ConfigBase::parseWallTimeToSeconds(const std::string& arg, bool& success, const std::string& option)
 {
     // first attempt to parse seconds only. Assume \d+[s] until it's not
-    char* pos;
-    unsigned long  seconds = strtoul(arg.c_str(), &pos, 10);
+    char*         pos;
+    unsigned long seconds = strtoul(arg.c_str(), &pos, 10);
     while ( isspace(*pos) )
         ++pos;
     if ( *pos == 's' || *pos == 'S' ) {
--- a/src/sst/core/configBase.cc
+++ b/src/sst/core/configBase.cc
@@ -59,14 +59,29 @@ ConfigBase::parseBoolean(const std::string& arg, bool& success, const std::strin
 uint32_t
 ConfigBase::parseWallTimeToSeconds(const std::string& arg, bool& success, const std::string& option)
 {
+    // first attempt to parse seconds only. Assume \d+[s] until it's not
+    char* pos;
+    unsigned long  seconds = strtoul(arg.c_str(), &pos, 10);
+    while ( isspace(*pos) )
+        ++pos;
+    if ( *pos == 's' || *pos == 'S' ) {
+        while ( isspace(*++pos) )
+            ;
+        if ( !*pos ) {
+            if ( seconds <= UINT32_MAX ) {
+                success = true;
+                return seconds;
+            }
+        }
+    }
+
     static const char* templates[] = { "%H:%M:%S", "%M:%S", "%S", "%Hh", "%Mm", "%Ss" };
     const size_t       n_templ     = sizeof(templates) / sizeof(templates[0]);
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wmissing-field-initializers"
     struct tm res = {}; /* This warns on GCC 4.8 due to a bug in GCC */
 #pragma GCC diagnostic pop
-    char*    p;
-    uint32_t seconds;
+    char* p;
 
     for ( size_t i = 0; i < n_templ; i++ ) {
         memset(&res, '\0', sizeof(res));


src/sst/core/configBase.cc
```diff
@@ -59,14 +59,30 @@ ConfigBase::parseBoolean(const std::string& arg, bool& success, const std::strin
...
+            seconds = std::stoul(str, &pos, 10);
+            if ( pos == str.size() ) {
+                success = true;
+                return seconds;
+            }
+        }
+        catch ( const std::exception& e ) {
+        }
+    }
+
```
leekillough, line 78: I would write it:
```c++
uint32_t
parseWallTimeToSeconds(const std::string& arg, bool& success, const std::string& option)
{
    // first attempt to parse seconds only. Assume \d+[s] until it's not
    char* pos;
    auto  seconds = strtoul(arg.c_str(), &pos, 10);
    while ( isspace(*pos) )
        ++pos;
    if ( *pos == 's' || *pos == 'S' ) {
        while ( isspace(*++pos) )
            ;
        if ( !*pos ) {
            if ( seconds <= UINT32_MAX ) {
                success = true;
                return seconds;
            }
        }
    }
```
- It avoids memory allocation by not creating `str`
- It avoids exceptions by using C `strtoul` instead of C++ `std::stoul`
- It allows `s` or `S` suffix
- It allows spaces before and after `s` or `S`
- It falls back to the old code if it doesn't match Perl regexp `^\s*[0-9]*\s*[Ss]\s*$` or is out of range of `uint32_t`

src/sst/core/configBase.cc
```diff
@@ -59,14 +59,29 @@ ConfigBase::parseBoolean(const std::string& arg, bool& success, const std::strin
 uint32_t
 ConfigBase::parseWallTimeToSeconds(const std::string& arg, bool& success, const std::string& option)
 {
+    // first attempt to parse seconds only. Assume \d+[s] until it's not
+    char* pos;
+    auto  seconds = strtoul(arg.c_str(), &pos, 10);
```
gvoskuilen, line 64: Change 'auto' to 'unsigned long'. Otherwise looks good.

leekillough: `<rant>`

The reason `auto` was originally used, was to put the type declaration in one place, namely the return value of the function (`strtoul`, `strtoull`, etc.), and to prevent differences between the declared type of the variable and the return type of the function, which could lead to truncation or other unexpected behavior.

Using `auto` makes the type of a variable, or the return type of a function, etc., be the same as the value it is being set to, returning, etc., when knowing the exact type used is less important than the meaning of the variable or function. C++14 adds `auto` to function return types and generic lambdas; C++17 adds `auto` for non-type template parameters, so that you can write a template which performs the same operation  on an entity regardless of its type; and C++20 adds `auto` for function parameters. 

Some people say it's confusing, but I actually think it's clearer, when the variable, function or template parameter is "more important in function than in form". In the example cited here, the variable is declared `auto` because its precise type is not central to the operation, and the code says "Give me a variable which can hold the result of `strtoul()` or some other function which performs a certain operation. I don't care about its type, just the operational function. If later I decide to change it to `strtoull()` or some other function, the variable will follow the function."

Using `auto` is also to shield yourself from changes in an external library, or to use opaque types whose implementation details or exact layout is not important. You write your code once with `auto`, and if the function, variable, etc. changes, you don't have to change your code as long as it still supports the same functions/methods/etc. on the type.

Using `auto` is similar to using `let` in functional languages, where the operation is more important than the type.

Now, when doing things like writing code for processor registers and instructions, I don't use `auto` because the instructions and registers have a very specific size. For those, I use types like `uint32_t` and `uint64_t` rather than `auto` or even `unsigned long`.

feldergast: I've read a lot of online discussions on this topic and there seems to be people on both sides of this.  Many think auto should be used everywhere for many of the reasons you describe.  Others argue auto should only be used when necessary, and not out of convenience, to enhance code understandability. I can see both sides of the argument, but for SST, we've fallen more towards the later.  In my mind, most of what you describe above is convenience, not necessity.  It is convenient not to have to change my code if I change the type of a variable.  On the other hand, if I go and change the type everywhere in the code, then people reading the code will know what types they're dealing with.  It is convenient not to have to change my code if an external library changes.  On the other hand, if an external library changes in a way visible to my code, I'd want to review my code anyway to make sure it's still interacting in a proper way. For how rarely these types of things happen, I find the extra work to be a good trade-off for better accessibility to developers (especially new developers, many of whom are not C++ gurus). 

For me, having a specific type instead of auto would never be less understandable, though it may not always be more understandable.  In C++, form and function are often linked, i.e., you can't understand the code without knowing the types you're operating on. The cost of entry to understanding a new code base goes up dramatically when auto is used everywhere because understanding the data you're operating on becomes challenging. In SST we intentionally decided to err on the side of readability over convenience in the hopes of being able to enable more users to contribute. I do recognize this can cause some pain in writing code.  I also recognize that there are cases where auto could help with things like avoiding truncation I that is something we give up with this standard, and we could look at refining the way we use auto to account for that.  Though I typically see compiler warnings if I haven't specifically downcast the bit width.

As to this specific case, in the end, I think either auto or unsigned long are both as readable.  But, where do you draw the line?  What's easily understood by one may not be easily understood by another. In developing SST, we have decided not to use auto except when it's necessary.  We also make exceptions for things like iterators, but it's best to only use it there if the types you're dealing with are immediately recognizable in the surrounding code.   If there's no way to do what you want without using auto, then a comment about how to determine types would be welcome.

feldergast (approved):

Issue #1066, opened 2024-04-04
Support for SST::Statistics handlers in JSON input files


jleidel:
**Is your feature request related to a problem? Please describe.**
The current JSON reader/writers for simulation inputs/outputs do not support including statistics directives.  It would be ideal for JSON to handle SST::Statistics directives in the same manner as the Python inputs.

**Describe the solution you'd like**
Add the necessary standard handlers to enable and configure statistics for components, subcomponents, component/subcomponent groups and simulation-wide.  This will likely require some notional internal API changes in order to provide sufficient functionality and interoperability.

**Describe alternatives you've considered**
There aren't really any alternatives if we truly want JSON inputs to be equivalent to Python inputs

**Additional context**
I can mock up the potential JSON verbiage if need be.  Ideally, we'll need: 

1. simulation wide statistic enablers
2. component/group enablers
3. per component/subcomponent enablers

gvoskuilen:
I made assignments so we can track this. If you can mock something up and identify what needs to changed/added, that'd be great.

jleidel:
Code was merged; closing this for now

PR #1232, opened 2025-03-05
Adding global statistics and statistics groups for JSON reader and writer

jleidel:
Adding global statistics and statistics groups for JSON reader and writer.  This targets Issue #1066.  Currently passes all the standard `sst-core` tests.


feldergast (changes requested):
At quick glance, these changes look good.  But, we'll need to make sure the new functionality is being tested. Probably the easiest way to add them is to add them to the config_input_output test suite.  To do that, add a stat to sst/core/testElements/message_mesh/enclosingComponent.{h,cc} and enable it in the input file.  This should test both stat output and input for json.

PR #1233, opened 2025-03-06
Use modern techniques to delete copy constructors and assignment operators

leekillough:
This uses modern methods of deleting copy constructors and copy assignment operators, with `= delete`.

Copy assignment operators declared `void operator=()` were changed to `class& operator=()` to become a copy assignment operator as defined by the language, so that the default one can be deleted.

For some classes, a default constructor previously explicitly declared `private` to prevent its use outside of the class was removed, because the declaration of the copy constructor causes the deletion of the default constructor.

### DIFF
--- a/src/sst/core/checkpointAction.h
+++ b/src/sst/core/checkpointAction.h
@@ -84,10 +84,9 @@ public:
     NotSerializable(SST::CheckpointAction);
 
 private:
-    CheckpointAction() {}
-    CheckpointAction(const CheckpointAction&);
+    CheckpointAction(const CheckpointAction&) = delete;
+    CheckpointAction& operator=(const CheckpointAction&) = delete;
 
-    void operator=(CheckpointAction const&);
     void createCheckpoint(Simulation_impl* sim); // The actual checkpoint operation
 
     RankInfo       rank_;          // RankInfo for this thread/rank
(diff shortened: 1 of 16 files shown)


berquist (approved):
I think separately there should also be a policy on move, now that the copy constructors and assignment are explicitly deleted.

leekillough:
> I think separately there should also be a policy on move, now that the copy constructors and assignment are explicitly deleted.

Or, at least, [The Rule of 3](https://en.cppreference.com/w/cpp/language/rule_of_three), or [The Rule of 5](https://en.cppreference.com/w/cpp/language/rule_of_three) if moving is significantly faster than copying (such as if new memory is dynamically allocated during copying but can simply be reused during moving, or if deep copies are expensive while shallow copies just move a few pointers around), or moving is allowed and copying is not, e.g. such as `std::unique_ptr`.

`= default` can be used instead of `= delete` if the behavior of copying, moving or assigning a class member-by-member (including direct bases) is the desired behavior. If the compiler cannot automatically generate copying/ moving/ assignment member-by-member, then explicitly declaring `= default` will cause a compilation error while leaving `= default` out will implicitly delete the method without explicitly declaring it `= delete`, while explicit `= delete` is preferred.

For C-style `struct`s without explicit constructors, the default behavior can be relied upon without explicitly declaring constructors or assignment operators. The Rule of 3/5/0 says that if you declare one of (copy constructor, copy assignment operator, destructor), you should declare all 3, even if only `= default`, and if you declare one of (move constructor, move assignment operator), you should declare all 5 of the aforementioned.

PR #1238, opened 2025-03-07
adding json stats support

jleidel:
see the code

fix for issue #1066

### DIFF
--- a/tests/test_MessageMesh.py
+++ b/tests/test_MessageMesh.py
@@ -18,6 +18,9 @@ sst.setProgramOption("stop-at", "10us")
 x_size = int(sys.argv[1])
 y_size = int(sys.argv[2])
 
+sst.setStatisticOutput("sst.statOutputCSV")
+sst.enableAllStatisticsForAllComponents()
+
 # Calculate number of routers and endpoints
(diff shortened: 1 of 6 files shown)


tests/test_MessageMesh.py
```diff
@@ -18,6 +18,9 @@
 x_size = int(sys.argv[1])
 y_size = int(sys.argv[2])
 
+sst.setStatisticOutput("sst.statOutputCSV")
```
feldergast, line 21: Using the default name for stats output file in the test could cause random failures when running testsuites in parallel.  However, I think this is a problem in other testsuites as well and we don't currently run the tests in parallel so we can reevaluate this one when we look at the others.

feldergast (approved):

tests/test_MessageMesh.py
jleidel: I can override it with a specific name if need be

feldergast: Let's leave it as is for now.  This just got me thinking about parallel testsuites running and I found errors in other testsuites as well.  We'll tackle them all together.

Issue #1053, opened 2024-03-07
GitHub Actions infrastructure for testing


berquist:
Autotester 2 will be based upon self-hosted GitHub runners.  In order to support this we need to be able to build and test within a GHA-based environment in general.  (Right now this is outsourced entirely to bare metal and VMs accessed through Jenkins.)  To start with, we will use their free hosted runners with
- Ubuntu 22.04 on x86_64
- macOS 13 on x86_64
- macOS 14 on M1
We had a few internal discussions about costs of provided macOS runners.  They are 100% free, even the ARM ones, just like the Linux ones.  These are VMs, not containers; we may eventually want a Rocky-, UBI-, or Alma-based container as a free runner, but these will be covered by AT2-based self-hosted runners no matter what.

AT2 will not use labels to control whether or not it runs, but whether or not a PR is a [draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request).  The basic setup using free runners should work this way too.

The lint checks also use labels for interacting with AT1.  This will be replaced with calling the lint checks before running tests/AT2 and stopping the workflow using GH-provided built-in rules.

In order to migrate incrementally, all the new GH-provided infrastructure can be added without touching anything existing.  For the formatting/lint checks, have https://github.com/pre-commit/action (not the provided service pre-commit.ci) run using https://pre-commit.com/, and only once this passes, start building and testing.

---

The setup for elements should look similar.  There is an outstanding question about how to deal with the possible race that AT1 detects, where a commit is being made to core that would break an approved commit to elements and is not yet the `HEAD` of core that the elements commit sees.  This is a problem that is more general than AT2 and requires further investigation.  I don't expect it to need a complete redo of any initial work that doesn't account for repository synchronization.

PR #1239, opened 2025-03-07
Fix pre commit and label interaction

berquist:
(Redo of #1237)

We've discussed on and off that there is a race condition with setting the labels between the script-based format checks for C++ and CMake.  There will soon be an addition problem that Autotester 2 will not play properly with these two isolated GHA workflows that currently only interact via labels.  Instead, AT2 will run via a hierarchical set of GHA workflows at the currently mostly unused entry point https://github.com/sstsimulator/sst-core/blob/e992a3fd54178adf2ba9018c35335d32ac8d9f7f/.github/workflows/ci.yml (#1053).

In preparation for this, some time ago (#1055) I added this driver that calls instead https://pre-commit.com/ which has been running simultaneously.  It properly pins linter/hook versions, without the extra isolated workflows and shell scripts, and it failing will prevent any following tests or workflows from running.

Since AT1 still needs to work with labels, I modified the pre-commit workflow to just set the `clang-format` ones when it passes or fails, with updated status messages.


berquist:
Draft while I figure out why this workflow can't access labels but the standalone ones can.

berquist:
A better version of this fix will appear as part of the AT2 rework.

PR #1248, opened 2025-03-13
Clean up UNUSED() macro

leekillough:
This removes the `UNUSED()` macro from function _declarations_ rather than _definitions_ and uses the C++17 `[[maybe_unused]]` attribute rather than GCC/Clang attributes.

This is a much smaller version of https://github.com/sstsimulator/sst-core/pull/1247 which imposes no incompatibilities.


feldergast (approved):
Looks good.  Thanks for changing the macro to use the new [[maybe_unused]]

feldergast (dismissed):

feldergast:
The compilers on the ARM platforms don't seem to like the [[maybe_unused]] attribute on one of the function declarations:  

```
08:28:46 ../../../src/sst/core/serialization/impl/serialize_array.h:91:23: error: 'maybe_unused' attribute cannot be applied to types
08:28:46     void operator()(T UNUSED(arr[N]), serializer& UNUSED(ser), const char* UNUSED(name))
08:28:46                       ^
08:28:46 ../../../src/sst/core/warnmacros.h:17:23: note: expanded from macro 'UNUSED'
08:28:46 #define UNUSED(x) x [[maybe_unused]]
```

src/sst/core/warnmacros.h
```diff
@@ -14,7 +14,7 @@
 #ifndef SST_CORE_WARNMACROS_H
 #define SST_CORE_WARNMACROS_H
 
-#define UNUSED(x) x __attribute__((unused))
+#define UNUSED(x) x [[maybe_unused]]
```
hughes-c, line 17: I haven't checked with the arm compiler, but maybe `#define UNUSED(x) [[maybe_unused]] x`?

leekillough: > I haven't checked with the arm compiler, but maybe `#define UNUSED(x) [[maybe_unused]] x`?

I think that fails compilation with some current usages of `UNUSED()`. The macro calls would need to be changed. See my previous commend on the combinations of `[[maybe_unused]]` and their relationships to how the macro is called.

The current PR on Core builds on almost all platforms, because it puts `[[maybe_unused]]` after the macro argument, which is usually a variable name without a type. If we use your suggestion, all macro calls would need to be changed to have the complete type and variable passed to the macro, like `UNUSED(int x[10])`, while it now uses `int UNUSED(x[10])` or `int UNUSED(x)[10]`.

leekillough:
> The compilers on the ARM platforms don't seem to like the [[maybe_unused]] attribute on one of the function declarations:
> 
> ```
> 08:28:46 ../../../src/sst/core/serialization/impl/serialize_array.h:91:23: error: 'maybe_unused' attribute cannot be applied to types
> 08:28:46     void operator()(T UNUSED(arr[N]), serializer& UNUSED(ser), const char* UNUSED(name))
> 08:28:46                       ^
> 08:28:46 ../../../src/sst/core/warnmacros.h:17:23: note: expanded from macro 'UNUSED'
> 08:28:46 #define UNUSED(x) x [[maybe_unused]]
> 08:28:46                       ^
> ```

I have pushed a change from `UNUSED(arr[N])` to `UNUSED(arr)[N]`, which should fix this issue, because the attribute should follow the name (in certain simple declarations, some attributes can precede the whole declaration and apply to all variables, but attributes which follow variables apply only to those variables. See https://github.com/sstsimulator/sst-elements/pull/2475#issuecomment-2750003122 ).

feldergast (approved):

PR #1249, opened 2025-03-14
Add override, [[noreturn]] and format attributes

leekillough:
Add `override` for virtual functions which are overriden (as warned by `-Wsuggest-override`).

For default constructors with all base classes and class members default-initialized `()` and an empty constructor body, and for destructors with an empty body, move the definition from the `.cc` file to the `.h` file with a `= default` definition, except for one case in which a comment specifically wants it to not be inlined for some reason..

Add `explicit` to constructors with a single parameter which should not perform implicit conversions. I built sst-core and sst-elements and got no errors, but if some of the constructors need to perform implicit conversions, then `explicit` can be removed.

Add `= delete` declarations for copy constructors and copy assignment operators for classes which do not need them (or which had marked them as `private`). Not all classes which can have deleted copy constructors and assignment operators were modified by this PR, but many were.

Add virtual destructors `virtual ~func() = default;` or `~func() override = default;` to classes with virtual functions and no declared destructor.

Add `__attribute__((format(...)))` to `printf`-style functions which take format strings. For non-static member functions, the `this` pointer is implicitly the first parameter, so the 2nd and 3rd arguments of the `format()` attribute are increased by one.

Add `[[noreturn]]` to non-virtual functions which do not return, such as calling `Output::fatal()`. (If the function is virtual, then `[[noreturn]]` has little benefit unless the function is marked as `final`, which is dangerous to do because SST Elements and other client code might override SST Core classes, and without a careful analysis of which functions and classes are really `final` in SST Core, it is too risky to add `final` to functions or classes, so `[[noreturn]]` is left off of virtual functions.)

Use `std::make_unique()` for variable-length arrays instead of `array[N]` when `N` is not constant. VLAs are non-standard, and allocate on the stack instead of the heap, which increases the chance of stack overflows.

Replace `volatile` counters with `std::atomic` types, except in one case where the assembly code is listed and analyzed and `volatile` is used for a timing delay. `volatile` is [deprecated in C++20](https://www.youtube.com/watch?v=KJW_DLaVXIY), and most uses of it are replaceable with atomic variables. 

Reformat `SST_ELI_REGISTER_SUBCOMPONENT_API()` macro calls to have one argument per line.

Reformat long `TimeVortex` string messages to be split across lines.

Reformat `sst-config` `print_usage()` to use one call to `fputs()` instead of dozens of consecutive calls to `fprintf()`.

Rename `sst-info`'s `dprintf()` to `sst_dprintf()` since `dprintf()` is a POSIX C function which prints to an integer file descriptor.

Could have been done, but wasn't:

Default constructors and destructors defined with empty bodies in the header could be replaced with `= default` which is more specific than `{}` and could conceivably be more optimized. To minimize diffs, these `{}` default constructors and destructors in headers were left alone. Only if they were newly added, had `override` added, or moved from `.cc` to `.h` files, was `= default` used instead of `{}`.

`virtual` could be removed wherever `override` was used. `override` implies `virtual`, but it was felt that leaving `virtual`, which is at the beginning, would make it more readable. If `virtual` existed before, it was not removed simply because `override` was added, but if neither `virtual` nor `override` existed and `override` applied because the function was `virtual` in a base class, then only `override` was added.

### DIFF
--- a/src/sst/core/impl/timevortex/timeVortexPQ.cc
+++ b/src/sst/core/impl/timevortex/timeVortexPQ.cc
@@ -153,7 +153,10 @@ public:
     TimeVortexPQ() : TimeVortexPQBase<false>() {} // For serialization only
     ~TimeVortexPQ() {}
 
-    void serialize_order(SST::Core::Serialization::serializer& ser) { TimeVortexPQBase<false>::serialize_order(ser); }
+    void serialize_order(SST::Core::Serialization::serializer& ser) override
+    {
+        TimeVortexPQBase<false>::serialize_order(ser);
+    }
 
     SST_ELI_EXPORT(TimeVortexPQ)
 };
@@ -166,15 +169,19 @@ public:
         TimeVortexPQ_ts,
         "sst",
         "timevortex.priority_queue.ts",
-        SST_ELI_ELEMENT_VERSION(1,0,0),
-        "Thread safe verion of TimeVortex based on std::priority_queue.  Do not reference this element directly, just specify sst.timevortex.priority_queue and this version will be selected when it is needed based on other parameters.")
-
+        SST_ELI_ELEMENT_VERSION(1, 0, 0),
+        "Thread-safe verion of TimeVortex based on std::priority_queue.  Do not reference this element directly; just"
+        " specify sst.timevortex.priority_queue and this version will be selected when it is needed based on other"
+        " parameters.")
 
     TimeVortexPQ_ts(Params& params) : TimeVortexPQBase<true>(params) {}
     TimeVortexPQ_ts() : TimeVortexPQBase<true>() {} // For serialization only
     ~TimeVortexPQ_ts() {}
 
-    void serialize_order(SST::Core::Serialization::serializer& ser) { TimeVortexPQBase<true>::serialize_order(ser); }
+    void serialize_order(SST::Core::Serialization::serializer& ser) override
+    {
+        TimeVortexPQBase<true>::serialize_order(ser);
+    }
 
     SST_ELI_EXPORT(TimeVortexPQ_ts)
 };
--- a/src/sst/core/interfaces/simpleNetwork.h
+++ b/src/sst/core/interfaces/simpleNetwork.h
@@ -38,7 +38,8 @@ namespace SST::Interfaces {
 class SimpleNetwork : public SubComponent
 {
 public:
-    SST_ELI_REGISTER_SUBCOMPONENT_API(SST::Interfaces::SimpleNetwork, int)
+    SST_ELI_REGISTER_SUBCOMPONENT_API(SST::Interfaces::SimpleNetwork,
+                                      int)
 
     /** All Addresses can be 64-bit */
     using nid_t = int64_t;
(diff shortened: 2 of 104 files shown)


src/sst/core/impl/timevortex/timeVortexPQ.cc
```diff
@@ -162,19 +165,20 @@ class TimeVortexPQ_ts : public TimeVortexPQBase<true>
 {
 public:
     SST_ELI_REGISTER_DERIVED(
-        TimeVortex,
-        TimeVortexPQ_ts,
-        "sst",
-        "timevortex.priority_queue.ts",
-        SST_ELI_ELEMENT_VERSION(1,0,0),
-        "Thread safe verion of TimeVortex based on std::priority_queue.  Do not reference this element directly, just specify sst.timevortex.priority_queue and this version will be selected when it is needed based on other parameters.")
+        TimeVortex, TimeVortexPQ_ts, "sst", "timevortex.priority_queue.ts", SST_ELI_ELEMENT_VERSION(1, 0, 0),
```
feldergast, line 168: Convention is to have one item per line.  This should remain as it currently is.

src/sst/core/interfaces/simpleNetwork.h
```diff
@@ -174,7 +174,7 @@ class SimpleNetwork : public SubComponent
     {
 
     public:
-        SST_ELI_REGISTER_SUBCOMPONENT_API(SST::Interfaces::SimpleNetwork::NetworkInspector,std::string)
+        SST_ELI_REGISTER_SUBCOMPONENT_API(SST::Interfaces::SimpleNetwork::NetworkInspector, std::string)
```
feldergast, line 177: Looks like this one doesn’t follow convention and should be changed to one item per line at some point.

feldergast (changes requested):
There are a lot of irons in the fire right now and this touches a broad range of files.  For now, please just leave the virtual keywords in place as they aren’t actually wrong, just redundant.  Go ahead and remove any virtual keyword where you make another change to the function (add override or no return, for example), but leave the rest.

feldergast:
I should also point out that some (maybe many in the core) instances of having both virtual and override is intentional.  There are some base APIs that are really only there for the core to interact with and then the user facing APIs inherit from those.  In that case, we like to have the virtual keyword on the functions the user should be overriding. It is easy to miss the override keyword at the end of the declaration, so having the virtual on the front of it is good.  As an example, the Activity base class is used for core interactions, but the user really only sees the Event and Action classes. All virtual functions in those classes are marked virtual, even if they inherit them from Activity.  You’ll want to watch out for these cases and leave the virtual keyword intact.

src/sst/core/impl/timevortex/timeVortexPQ.cc
leekillough: It was a `clang-format` difference. Newer versions apparently do that.

leekillough:
> I should also point out that some (maybe many in the core) instances of having both virtual and override is intentional.

Okay, `override` implies `virtual`, I prefer being less verbose, and it follows the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override)

PR #1252, opened 2025-03-16
Add = delete declarations for copy constructors and copy assignment operators which are not needed

leekillough:
Add `= delete` declarations for copy constructors and copy assignment operators for classes which do not need them (or which had marked them as private). Not all classes which can have deleted copy constructors and assignment operators were modified by this, but many were.


leekillough:
C++ legalese note:

The move constructors (`class(class&&)`) and move assignment operators (`class& operator=(class&&)`) are not explicitly declared as `= delete` because by deleting the copy constructor (`class(const class&) = delete`) and the copy assignment operator (`class& operator=(const class&) = delete`), the move operators are implicitly declared as deleted, and we do not need to explicitly declare them as such.

If a class (like `std::unique_ptr`) needs to support moves but not copying, then the move operations can be explicitly declared as `= default` or be explicitly defined.

In this PR there is one or two classes which need copy construction and assignment, and those are declared as `= default` instead of as `= delete`. Previously they were not explicitly declared, but were implicitly declared and defined by the compiler. By explicitly declaring them `= default`, we ensure that they exist and can be default-generated as member-by-member copies/assignments, and if they cannot be default-generated, it results in a build error.

feldergast (approved):

PR #1255, opened 2025-03-16
Use std::make_unique() for variable-length arrays instead of array[N] when N is not constant

leekillough:
Use `std::make_unique()` for variable-length arrays instead of `array[N]` when `N` is not constant. VLAs are non-standard, and allocate on the stack instead of the heap, which increases the chance of stack overflows.


leekillough:
@gvoskuilen

gvoskuilen (approved):

PR #1258, opened 2025-03-16
Replace volatile counters with std::atomic types

leekillough:
Replace `volatile` counters with std::atomic types, except in one case where the assembly code is listed and analyzed and `volatile` is used for a timing delay.  `volatile` is [deprecated](https://www.youtube.com/watch?v=KJW_DLaVXIY).

### DIFF
--- a/src/sst/core/interprocess/tunneldef.h
+++ b/src/sst/core/interprocess/tunneldef.h
@@ -22,8 +22,10 @@
 
 #include "sst/core/interprocess/circularBuffer.h"
 
-#include <inttypes.h>
+#include <atomic>
+#include <cinttypes>
 #include <unistd.h>
+#include <utility>
 #include <vector>
 
 namespace SST::Core::Interprocess {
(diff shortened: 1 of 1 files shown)


gvoskuilen:
@leekillough Do you know if these changes are compatible with PinCRT? This file gets compiled into both SST and a Pintool so there are restrictions on what it can include and the atomic needs to work in a shared memory region between SST & the pintool.

leekillough:
> @leekillough Do you know if these changes are compatible with PinCRT? This file gets compiled into both SST and a Pintool so there are restrictions on what it can include and the atomic needs to work in a shared memory region between SST & the pintool.

No, I don't know anything about PinCRT. I would expect than an atomic would work unless the shared memory gets remapped after the std::atomic is constructed. But if multiple processes/threads/whatever access mapped memory with atomic instructions, it should produce the desired effects that `volatile` produced. 

In reviewing the PinCRT documentation, release notes, etc., C++11 support has been only added recently, and in a 2023 release note, [it says that C++11 atomic are not supported](https://groups.io/g/pinheads/message/13595).

>       Pin on Linux C++ 17 Support
>
> o   Pin on Linux now supports large portions of C++ 17 standard with the following known issues:
>    §  RTTI and exceptions are not supported (as in previous Pin versions)
>    §  Thread safety and Concurrency support library are not supported
>    ·       The above applies also to std::atomic_xxx(std::shared_ptr) utility functions
>    ·       Users must continue using Pin Thread API and LOCK primitives for multi threading scenarios and concurrency control

PR #1260, opened 2025-03-16
Inline trivial default constructors and destructors

leekillough:
For default constructors with all bases and class members default-initialized `()` and an empty constructor body, and for destructors with an empty body, move the definition from the `.cc` file to the `.h` file with a `= default` definition.

Default constructors and destructors defined with empty bodies in the header could be replaced with `= default` which is more specific than `{}` and could conceivably be more optimized. To minimize diffs, these `{}` default constructors and destructors in headers were left alone. Only if they were newly added or moved from `.cc` to `.h` files, was `= default` used instead of `{}`.


feldergast (approved):

feldergast:
This PR is now unmergeable.  Please resolve conflicts.

feldergast (approved):

PR #1264, opened 2025-03-17
Output directory fix

feldergast:
Added a SST::Util::Filesystem to track the --output-directory option and provide helper functions for ensuring files generated in the core are create there. Also added the ability to specify the filename format for checkpoints.


src/sst/core/checkpointAction.cc
```diff
@@ -22,13 +22,52 @@
 #include "sst/core/timeConverter.h"
 
 // #include <filesystem>
```
hughes-c, line 24: Should remove the commented line

src/sst/core/checkpointAction.cc
```diff
@@ -53,7 +92,37 @@ CheckpointAction::CheckpointAction(
...
+        dir_format_  = format;
+        file_format_ = format;
+    }
+    else {
+        dir_format_  = format.substr(0, split);
+        file_format_ = format.substr(split + 1);
+    }
+
+    if ( (0 == this_rank.rank) ) {
+        // Check to make sure that there is at least on of %n or %t to
```
hughes-c, line 110: one of...

src/sst/core/main.cc
```diff
@@ -703,6 +702,18 @@ main(int argc, char* argv[])
...
     }
 
+    // Set up the Filesystem object with the output directory
+    bool out_dir_okay = Simulation_impl::filesystem.setBasePath(cfg.output_directory());
+    if ( !out_dir_okay ) {
+        fprintf(
+            stderr,
+            "ERROR: Directory specified with --output-directory (%s) is not valid.  Most likely causes are that the "
+            "user "
+            "does not have permissions to write to this path, or a file of the same name exists.\n",
```
hughes-c, line 712: You don't support overwriting a file?

src/sst/core/util/filesystem.cc
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    // Check to see if this existing directory is writeable
+    if ( isDirectoryWritable(parent.string()) ) {
+        base_path_ = absolute_path;
+        return true;
+    }
+
+    return false;
+}
```
hughes-c, line 136: Already briefly discussed this but this seems pretty heavy-weight. We should look for a better solution for the future.

hughes-c (changes requested):

src/sst/core/main.cc
feldergast: This is trying to create a directory that has the same name as a file.  If you use the fopen() or of of ofstream() calls, then you can overwrite a file.

hughes-c (approved):

feldergast:
Testing failures were python errors in the testing framework code.  Setting retest.

Issue #1225, opened 2025-02-26
Add ability to break to interactive console on SIGUSR1/2


skuntz:
**Is your feature request related to a problem? Please describe.**
If a program is running longer than expected, I want to break into an interactive console to inspect the current state. 

**Describe the solution you'd like**
Add sst.rt.interactive.cosole to the real time actions that can be triggered by SIGUSR1/2

**Describe alternatives you've considered**
None

**Additional context**

skuntz:
I have code that supports a new sst.rt.interactive realtime action in sst-core to support using sigusr1/2 to break into an interactive console. I explicitly disallow it for sigalrm since that didn't make sense to me.  I will clean that up and submit a PR. 

As far as support for breaking into interactive console from a component, I will change my example to use the baseComponent functionality instead of the realtime action directly.

skuntz:
Fixed in #1266

PR #1266, opened 2025-03-18
Interactive console realtime action

skuntz:
This PR add supports for a realtime action (sst.rt.interactive) to break to an interactive console from sigusr1 or sigusr2. It causes an error if used with sigalrm. Issue #1225 
---

----

### DIFF
--- a/src/sst/core/realtime.h
+++ b/src/sst/core/realtime.h
@@ -154,7 +154,7 @@ public:
 
     InteractiveRealTimeAction();
     void execute() override;
-    bool validSigalrmAction() override { return false; }
+    bool isValidSigalrmAction() override { return false; }
 };
 
 /* Wrapper for RealTimeActions that occur on a time interval */
--- a/src/sst/core/realtime.h
+++ b/src/sst/core/realtime.h
@@ -35,9 +35,13 @@ class ExitCleanRealTimeAction : public RealTimeAction
 {
 public:
     SST_ELI_REGISTER_REALTIMEACTION(
-        ExitCleanRealTimeAction, "sst", "rt.exit.clean", SST_ELI_ELEMENT_VERSION(0, 1, 0),
+        ExitCleanRealTimeAction, 
+        "sst", 
+        "rt.exit.clean", 
+        SST_ELI_ELEMENT_VERSION(0, 1, 0),
         "Signal handler that causes an immediate, but non-emergency shutdown. This is the default action for the "
-        "'--exit-after' option.")
+        "'--exit-after' option."
+    )
 
     ExitCleanRealTimeAction();
     virtual void execute() override;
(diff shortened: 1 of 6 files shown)


src/sst/core/simulation.cc
```diff
@@ -284,6 +284,11 @@ Simulation_impl::setupSimActions(Config* cfg, bool restart)
             std::string action;
             bool        found;
             parseSignalString(alarmstr, action, params);
+            if ( action == "sst.rt.interactive" ) {
```
feldergast, line 287: We probably shouldn't be testing for specific handler types.  If one type needs to check, then there are likely others that will as well. May be best to add a function in the base to class that returns a bool to indicate whether it is supported on sigalrm or not.  This check would then need to be moved to the proper place in the code where that function could be called.  Optionally, we may be able to add this information to the ELI registration call, but I'd need to look through the ELI base code to see how easy that may or may not be.

src/sst/core/realtime.h
```diff
@@ -116,6 +116,19 @@ class HeartbeatRealTimeAction : public RealTimeAction
     static Core::ThreadSafe::Barrier exchange_barrier_;
 };
 
+
+// SKK based on CoreStatusRealTimeAction
+class InteractiveRealTimeAction : public RealTimeAction
+{
+public:
+    SST_ELI_REGISTER_REALTIMEACTION(
```
feldergast, line 124: Convention for using the ELI registration macros is to put one item per line.

feldergast (changes requested):

src/sst/core/realtime.h
skuntz: It looks like my formatting is the same as the other ELI registrations in the file. Can you clarify what needs to change?

feldergast: If there are others in the file, don't worry about it.  I'll do a sweep before the release to make sure all the ELI calls follow the right convention.

skuntz: If this is the format you're referring to, I'm happy to update the ELI calls in the realtime.h file.
```
// REGISTER THIS COMPONENT INTO THE ELEMENT LIBRARY
    SST_ELI_REGISTER_COMPONENT(
        coreTestPerfComponent,
        "coreTestElement",
        "coreTestPerfComponent",
        SST_ELI_ELEMENT_VERSION(1,0,0),
        "CoreTest Test Perf Component",
        COMPONENT_CATEGORY_PROCESSOR
    )
```

feldergast: That is the format.  If you want to change them all in the file, that would be great.

src/sst/core/simulation.cc
skuntz: I added a virtual function validSigalrmAction based on my understanding of what we discussed in the last meeting. Let me know if that isn't what you had in mind.

PR #1268, opened 2025-03-20
Fix serialization to build in both mapped and un-mapped mode

leekillough:
Fixes https://github.com/tactcomplabs/sst-tools/issues/5 (cross-repo).

Remove the `name` parameter from serialization `operator()(T&, serializer& ser, const char* name)`, since it makes it too complicated, and caused compilation errors. It would be better to store mapped information in the `serializer` class state than to pass mapped `name` arguments at every level of the call hierarchy and to need to split the serialization  functions into mapped and non-mapped versions.

Add `ser.setMapName()` and `ser.getMapName()` to set and get mapped names, which are only used in certain places ~~(right now it's just a `const char*` pointer, but it can be turned into a `std::string` if necessary)~~. The `SST_SER()` macro will call the `sst_map_object()` function, which will set the quoted variable name to be used in the map before the serialization call, and clear the name after the serialization call, so that the name does not need to be passed around in every call.

`operator()` functions which used to be duplicated with a mapped and a non-mapped version, the mapped version of which was only valid in `serializer::MAP` mode, are now combined into one where the `serializer::MAP` mode, already handled in a `switch` statement, expects `getMapName()` to return the name.

Rewrite the default `serialize_impl<>` class to use modern C++ techniques like `if constexpr` and `decltype(auto)`. `tPtr` is a `const` reference variable which refers to either `t` or `&t` depending on whether `t` is a pointer or not, so that pointer and non-pointer code can be merged into one, by using the `tPtr` pointer to make method calls, and `tPtr` also reflects the new value of `t` when `t` is a pointer which gets assigned to in `UNPACK` mode. The `dependent_false` template is not needed anymore, since we can simply use a single `static_assert()` at the top of our function to assert that the type, with any top-level pointer removed, is a non-polymorphic class type.

Write a generalized `SST::Core::to_string()` which operates like `std::to_string` and also handles enumeration types, and generalize `SST::Core::from_string()`  to handle enumeration types. Right now, enumeration types are inputted and outputted as integers of the underlying type. There were build errors when the mapped and non-mapped serialization codes were merged, because the mapped code depended on `std::to_string` which isn't defined for enumeration types but which was depended on when the `serialize_impl` for "fundamental and `enum` types" created a mapping object.

For array serialization templates, use `size_t` instead of `int` for the size of the array.

The Core tests pass, and Elements builds, but there are some Elements test failures which need investigating and fixing.

### DIFF
--- a/src/sst/core/baseComponent.cc
+++ b/src/sst/core/baseComponent.cc
@@ -963,7 +963,7 @@ SerializeBaseComponentHelper::unpack_basecomponent(serializable_base*& s, serial
 }
 
 void
-SerializeBaseComponentHelper::map_basecomponent(serializable_base*& s, serializer& ser, const char* name)
+SerializeBaseComponentHelper::map_basecomponent(serializable_base*& s, serializer& ser, const std::string& name)
 {
     if ( nullptr == s ) return;
 
--- a/src/sst/core/baseComponent.h
+++ b/src/sst/core/baseComponent.h
@@ -1124,7 +1124,7 @@ public:
 
     static void unpack_basecomponent(serializable_base*& s, serializer& ser);
 
-    static void map_basecomponent(serializable_base*& s, serializer& ser, const char* name);
+    static void map_basecomponent(serializable_base*& s, serializer& ser, const std::string& name);
 };
 
 } // namespace pvt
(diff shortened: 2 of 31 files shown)


feldergast:
Thanks for putting this together.  Minus a few details, this looks very similar to how I first implemented things (including holding name state in the serializer) and was my preferred solution.  However, some constraints in the interactions between using serialization for checkpointing and mapping mode (where performance requirements are low) and using the same serialization for event synchronization (where performance requirements are high) led me to the two function solution.  I’m on vacation for spring break this week and saw this as I was triaging emails and wanted to give a heads up so you don’t spend too much time on it before we can talk about it.  I was under a time constraint in the original implementation and gave up on the one function solution, but maybe we can work something out that gets us back there while fulfilling all the constraints (which haven’t been fully documented yet since mapping mode is still experimental).

PR #1269, opened 2025-03-20
Add include guard to sst_config.h

gvoskuilen:
Simplifies use of the generated sst_config.h by giving it an include guard.


leekillough:
LGTM

jmlapre (approved):

jmlapre:
@gvoskuilen We should probably do this for elements as well...

PR #1271, opened 2025-03-23
Generalize serialize to support more containers in a generalized way, and store mapping name in a stack stored in the serializer

leekillough:
In addition to the [comments here](https://github.com/sstsimulator/sst-core/pull/1268#issue-2933593765), this PR:
- Places the names used in the serializer for mapping into _a stack_ in the `serializer` class. The names can be pushed on and popped off the stack to support multiple recursion levels of serialization. A name will be pushed on the stack when the `SST_SER()` macro is called, and then when serializations occur recursively, they will push another name (or no name if the current top name should be retained), etc., and when a serialization ends, the stack will be popped. The name does not need to be passed through the serialization function arguments, and the functions do not need to be duplicated with and without the name arguments. `ser.getMapName()` will return the last name pushed on the stack. (Another name for these map name functions can be used if desired.)
- Refactors `serialize_impl` to support more containers in a generalized way
  - Make the default `serialize_impl` template for user-defined non-polymorphic classes much shorter, using the same code for pointers and non-pointers
  - Add `serialize_insertable.h` and `serialize_utility.h` to support all sequenced and associative containers in one specialization 
    - `std::vector`, `std::vector<bool>`, `std::list`, `std::deque`, `std::map`, `std::multimap`, `std::set`, `std::multiset`, `std::unordered_map`, `std::unordered_multimap`, `std::unordered_set`, `std::unordered_multiset`
    - This "insertable" `serialize_impl` partial specialization is selected based on the existence of class types and methods it depends on, such as `value_type`, `begin()`, `end()`, `push_back()` and `insert()`, rather than listing specific classes by name
  - Add `serialize_adapter.h` to support `std::stack`, `std::queue` and `std::priority_queue` adapter classes as a pass-through serialization of the underlying container
  - Add `serialize_tuple.h` to support `std::tuple` and `std::pair`.


leekillough:
> This change now requires the macro to have a semicolon.
> 
> ```
> /Users/kgriesser/work/sst-bench/sst-bench/large-stat-chkpnt/large-stat-chkpnt.cc:60:26: error: expected ';' after expression
>    60 |     SST_SER(clockHandler)
> ```
> 
> If we want to keep it this way, which would be my preference as well, the documentation should also reflect this change: https://sst-simulator.org/sst-docs/docs/guides/features/checkpoint

Yes, we should decide on this as we move towards using the macro interface. I prefer to require the caller to use a semicolon, so that the _function-like macro_ is called the same as a function is. If there is a concern that the macro might need to be expanded into multiple statements, then the way to do it is to wrap it in a `do{ ... } while(0)` loop, which still requires the terminating semicolon, but which is usable as a single statement (such as in nested `if` statements).

leekillough:
> @leekillough I am unable to compile sst-elements with these changes. Can you check on your side?

I pushed more changes, but I only tested sst-core. I'll now test sst-elements.

src/sst/core/impl/interactive/simpleDebug.cc
```diff
@@ -71,7 +71,7 @@ SimpleDebugger::cmd_pwd(std::vector<std::string>& UNUSED(tokens))
 void
 SimpleDebugger::cmd_ls(std::vector<std::string>& UNUSED(tokens))
 {
-    std::vector<std::pair<std::string, SST::Core::Serialization::ObjectMap*>> vars = obj_->getVariables();
+    auto& vars = obj_->getVariables();
```
feldergast, line 74: Shouldn't use auto here, but we can change it back in a future PR so we can get the majority of the changes merged.

feldergast (approved):
Looks good.

PR #1278, opened 2025-03-28
Fix config output

gvoskuilen:
* Fixes config output to handle stat groups and shared stats correctly
* Fix several typos in help text
* Add testing for shared stats across subcomponents
* Fix issue where nested subcomponents didn't have access to shared_stats unless their immediate parent also had the shared_stat.

This fixes the underlying issues and the python output; fix for JSON outputs is still needed.


feldergast (approved):
Looks good.

PR #1282, opened 2025-04-06
Replace incorrect usage of std::is_fundamental with std::is_arithmetic

leekillough:
Replace all uses of the word and type traits `fundamental` with `arithmetic`, because `fundamental` is the wrong description -- the wrong named requirement -- for its uses.

--------

In C++ nomenclature and [`<type_traits>`](https://en.cppreference.com/w/cpp/header/type_traits), there are [different type categories](https://en.cppreference.com/w/cpp/language/type):
- Integral, `std::is_integral`: All signed and unsigned integer and `char` types, including `bool`
- Floating-point, `std::is_floating_point`: `float`, `double`, `long double`, extensions like `_Float16`
- Enumeration, `std::is_enum`: Scoped (`enum class`) and unscoped (`enum`) enumeration types
- Pointer, `std::is_pointer`: A pointer to an object or function, like `T*` or `void (*)()`
- Null pointer, `std::is_null_pointer`: `std::nullptr_t`, `decltype(nullptr)`, the type of a null pointer constant
- `void`, `std::is_void`
- Class, `std::is_class`
- Union, `std::is_union`
- Array, `std::is_array`
- Function, `std::is_function`
- Lvalue reference, `std::is_lvalue_reference`
- Rvalue reference, `std::is_rvalue_reference`
- Member object pointer, `std::is_member_object_pointer`
- Member function pointer, `std::is_member_function_pointer`

Composite type categories:
- Arithmetic, `std::is_arithmetic`: Integral and floating-point
- Scalar, `std::is_scalar`: Arithmetic, enumeration, pointer, pointer-to-member, and `std::nullptr_t`
    - not a reference, function, array, class, union, or `void`
- Object, `std::is_object`: Scalar, array, class or union
    - not a reference, function or `void`
- Reference, `std::is_reference`: Lvalue or Rvalue reference
- Member pointer, `std::is_member_pointer`: A pointer to a class member object or function
- Fundamental, `std::is_fundamental`: Arithmetic, `std::nullptr_t` and `void`.
    - The word _fundamental_ is used in the sense of the word _elemental_: That the type cannot be broken down any further into smaller pieces, like a pointer and its pointee, a class and its members, an enumeration and its underlying integral type, etc.
- Compound, `std::is_compound`: Array, class, union, enumeration, function, pointer to {object, function, member}, reference
    - The opposite of fundamental

--------

The word "Fundamental", and `std::is_fundamental`, `std::is_fundamental_v`, etc. were used incorrectly, when the correct term is _Arithmetic_: Integral types (including `char` and `bool`) and Floating-point types.

In places where "Fundamental" was used, I do not think that null pointer constants and `void` intended to be included:
  - "Yes, I'd like to serialize a `void` object. Oh wait, I can't declare one, much less pass it to a function!!!"
  - "I'd like to serialize a null pointer constant, not a general pointer to an object or function. I realize that it can only have one value, but there are cases where I'd like to call a serialization function or instantiate a template with `std::nullptr_t` when normally a pointer to an object or function is required, and well, you don't consider those object and function pointers to be fundamental types anyway."

All cases of the word `fundamental` have been replaced with `arithmetic`, preserving case, and the phrase `a fundamental` has been replaced with `an arithmetic`.

It should be remembered that although the word "fundamental" sounds cool, like "Well, it's a fundamental type in the C++ language -- that sounds cool, so we'll handle it", the word "fundamental" is not used casually in the C++ language definition: It means a built-in type which cannot be broken down into smaller components like a pointer and its pointee can, and it includes `std::nullptr_t` and `void`. That is almost certainly not the intended use of the word "fundamental" in SST. "Arithmetic" is the better word, and is the C++ language type category for integral and floating-point, with `std::is_arithmetic` being the `<type_traits>` trait for it.

### DIFF
--- a/src/sst/core/from_string.h
+++ b/src/sst/core/from_string.h
@@ -99,14 +99,16 @@ from_string(const std::string& input)
     return static_cast<T>(from_string<std::underlying_type_t<T>>(input));
 }
 
+///////////////////////////////////////////////////////////////////////////////
+
 template <class T>
 std::enable_if_t<!std::is_enum_v<T>, std::string>
 to_string(const T& input)
 {
-    if constexpr ( std::is_fundamental_v<T> )
+    if constexpr ( std::is_arithmetic_v<T> )
         return std::to_string(input);
     else
-        return typeid(T).name(); // For now, return a string if the type isn't fundamental or handled elsewhere
+        return typeid(T).name(); // For now, return a string if the type isn't handled elsewhere
 }
 
 template <class T>
@@ -116,6 +118,7 @@ to_string(const T& input)
     return std::to_string(static_cast<std::underlying_type_t<T>>(input));
 }
 
+// for std::string no-op, or class types which define operator std::string()
 inline std::string
 to_string(std::string s)
 {
--- a/src/sst/core/serialization/impl/serialize_array.h
+++ b/src/sst/core/serialization/impl/serialize_array.h
@@ -82,10 +82,10 @@ raw_ptr(TPtr*& ptr)
 
 /**
    Version of serialize that works for statically allocated arrays of
-   fundamental types and enums.
+   arithmetic types and enums.
  */
 template <class T, size_t N>
-class serialize_impl<T[N], std::enable_if_t<std::is_fundamental_v<T> || std::is_enum_v<T>>>
+class serialize_impl<T[N], std::enable_if_t<std::is_arithmetic_v<T> || std::is_enum_v<T>>>
 {
     template <class A>
     friend class serialize;
@@ -102,7 +102,7 @@ class serialize_impl<T[N], std::enable_if_t<std::is_fundamental_v<T> || std::is_
    non base types.
  */
 template <class T, size_t N>
-class serialize_impl<T[N], std::enable_if_t<!std::is_fundamental_v<T> && !std::is_enum_v<T>>>
+class serialize_impl<T[N], std::enable_if_t<!std::is_arithmetic_v<T> && !std::is_enum_v<T>>>
 {
     template <class A>
     friend class serialize;
@@ -120,11 +120,10 @@ class serialize_impl<T[N], std::enable_if_t<!std::is_fundamental_v<T> && !std::i
 
 /**
    Version of serialize that works for dynamically allocated arrays of
-   fundamental types and enums.
+   arithmetic types and enums.
  */
 template <class T, class IntType>
-class serialize_impl<
-    pvt::ser_array_wrapper<T, IntType>, std::enable_if_t<std::is_fundamental_v<T> || std::is_enum_v<T>>>
+class serialize_impl<pvt::ser_array_wrapper<T, IntType>, std::enable_if_t<std::is_arithmetic_v<T> || std::is_enum_v<T>>>
 {
     template <class A>
     friend class serialize;
@@ -142,7 +141,7 @@ class serialize_impl<
  */
 template <class T, class IntType>
 class serialize_impl<
-    pvt::ser_array_wrapper<T, IntType>, std::enable_if_t<!std::is_fundamental_v<T> && !std::is_enum_v<T>>>
+    pvt::ser_array_wrapper<T, IntType>, std::enable_if_t<!std::is_arithmetic_v<T> && !std::is_enum_v<T>>>
 {
     template <class A>
     friend class serialize;
--- a/src/sst/core/serialization/serialize.h
+++ b/src/sst/core/serialization/serialize.h
@@ -280,11 +280,11 @@ public:
 
 
 /**
-   Version of serialize that works for fundamental types and enums.
+   Version of serialize that works for arithmetic and enum types.
  */
 
 template <class T>
-class serialize_impl<T, std::enable_if_t<std::is_fundamental_v<T> || std::is_enum_v<T>>>
+class serialize_impl<T, std::enable_if_t<std::is_arithmetic_v<T> || std::is_enum_v<T>>>
 {
 public:
     void operator()(T& t, serializer& ser)
@@ -300,15 +300,14 @@ public:
 };
 
 /**
-   Version of serialize that works for pointers to fundamental types
-   and enums. Note that the pointer tracking happens at a higher
-   level, and only if it is turned on.  If it is not turned on, then
-   this only copies the value pointed to into the buffer.  If multiple
-   objects point to the same location, they will each have an
-   independent copy after deserialization.
+   Version of serialize that works for pointers to arithmetic and enum types.
+   Note that the pointer tracking happens at a higher level, and only if it is
+   turned on. If it is not turned on, then this only copies the value pointed
+   to into the buffer. If multiple objects point to the same location, they
+   will each have an independent copy after deserialization.
  */
 template <class T>
-class serialize_impl<T*, std::enable_if_t<std::is_fundamental_v<T> || std::is_enum_v<T>>>
+class serialize_impl<T*, std::enable_if_t<std::is_arithmetic_v<T> || std::is_enum_v<T>>>
 {
     template <typename>
     friend class serialize;
--- a/src/sst/core/statapi/statbase.h
+++ b/src/sst/core/statapi/statbase.h
@@ -318,12 +318,11 @@ private:
 /**
  \class StatisticCollector
  * Base type that creates the virtual addData(...) interface
- * Used for distinguishing fundamental types (collected by value)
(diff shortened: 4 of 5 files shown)


feldergast:
I'll have to give some thought to this one, because isArithmetic() is definitely not the right description in ObjectMap.  ObjectMap::isFundamental() is used in the way you describe fundamental above: in mapping, it is treated as an elemental, i.e. ObjectMaps will not break it down into smaller parts.  It indicates you can print and set its value. Maybe isElemental() is a better function name in ObjectMap than isFundamental(), but isArithmetic() doesn't make sense there, as you are not guaranteed to be able to do any arithmetic operations on it.

leekillough:
I am using the C++ definition of "fundamental", not the English use of the word "fundamental".

`void` cannot be stored in a variable or object or passed to a function; there can be no `void&` or `void&&`; and `void*` is not a fundamental type.

A `std::nullptr_t` is never used to store or pass values; it is only used to disambiguate function overloads ( [The rationale document for why C++11 added `std::nullptr_t` and `nullptr`](https://open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2431.pdf) ). A program is not going to have a `std::nullptr_t` variable or container value. I suppose it is legal to have a `std::vector<std::nullptr_t>` but it is useless.

With the exclusion of `void` and `std::nullptr_t`, all remaining _Fundamental_ types are _Arithmetic_ as tested by `std::is_arithmetic`, which includes all integral types (including `char` and `bool`) and floating-point types. That is why "Arithmetic" and `std::is_arithmetic` was used.

If "Arithmetic" sounds too complicated for English, "Numeric" can be used, but `std::is_arithmetic` is still the C++ type trait name. If you want to define your own SST type trait to mean the same thing as `std::is_arithmetic` but called something else, you can:
```C++
template<typename T>
using is_elemental = std::is_arithmetic<T>;
```
Names which can be used as descriptions:
- Arithmetic (the formal C++ name)
- Numeric
- Elemental
- Basic
- Builtin
- Simple

Rather than adjectives, maybe simply the nouns "Number" and "Numbers" should be used.

I would not want to use these names:
- Fundamental, because we would need to include `void` and `std::nullptr_t` or be at odds with the C++ formal definition of "Fundamental"
- Atomic, because these are not `std::atomic` types with atomic memory semantics

Regardless of what is used for the English description of the types, `std::is_fundamental` is incorrect today and has been replaced with `std::is_arithmetic`.

If you truly intended to handle `std::nullptr_t` as an object type such as a container key, and serialize/deserialize it, then the code does not currently handle objects of type `std::nullptr_t` correctly even though it includes them in its `std::is_fundamental` test.

feldergast:
I don't debate any of that, but that's also not what I was trying to say.  All I'm saying is that ObjectMap::isArithmetic() is not a correct description, as there will be ObjectMaps for underlying types that don't return true for is_arithmetic<T>, but would return true for ObjectMap::isArithmetic() (the ObjectMap for std::string is one of them).  The current use of ObjectMap::isFundamental(), in the context of how ObjectMaps are used, means approximately: a "built-in" ObjectMap type which cannot be broken down into smaller ObjectMaps (i.e. the ObjectMap has no children ObjectMaps).  Look familiar? I realize in the base language there is a specific meaning for fundamental, so we can avoid it in this context if you think it will cause confusion. But, ObjectMap::isArithmetic() is equally wrong.  We should just avoid the use of any of the specific names in the standard library.  I think ObjectMap::isElemental() is probably as good as any. None of this requires a new is_elemental<T> template, as none of these types will have to be checked with that, only through the isElemental() function call.  To the tools that use ObjectMap, returning true from this function just means that the virtual set() and get() function are expected to work, whereas these functions should not be called for things that return false from this function.

None of this should be interpreted as suggesting that we don't change the use of is_fundamental<T> to is_arithmetic<T>.

leekillough:
I like the names "Simple" or "Basic" rather than "Elemental". They could include `enum`.

I'll update this PR with a new SST trait which starts out as `std::is_arithmetic`. I'll also change the mapping function names and comments to use the new word.

gvoskuilen:
`isSimple` could be misleading in a different way because the data structure represented by the ObjecMap may not be simple even if it is not breakable into smaller ObjectMaps. `isBasic` and `isElemental` better capture the idea that this ObjectMap is not breakable into further ObjectMaps/is a building block for ObjectMaps representing combinations of the basic/elemental ObjectMaps.

leekillough:
I'm curious why `isFundamental()` was chosen to be a virtual function which is overriden and is evaluated at runtime, rather than a type trait which can be specialized and evaluated at compile-time, e.g.
```c++

template< typename T >
inline constexpr bool is_basic_v = std::is_arithmetic_v<T> || std::is_enum_v<T>;

template<>
inline constexpr bool is_basic_v< std::string > = true;

class MyClass {
};

template<>
inline constexpr bool is_basic_v< MyClass > = true;
```
Template specializations can be used instead of virtual function overrides, and they are evaluated at compile-time.

If you don't want to have a separate variable, you can define a `constexpr` variable or function returning it inside the class, but then it won't work for constant expressions if the function needs to be virtual (at least not in C++17).

feldergast:
I suppose you could do it that way, but you would still need an isBasic() (or similarly named function) that can be called by the tools using the ObjectMap.  All the tool has is an ObjectMap pointer and has to determine what is a valid operation on it.  The ObjectMap doesn't necessarily "know" the type of the object it's wrapping, other than it has a string representation of it. The ObjectMaps for "basic" types do know the type of the object they're wrapping, but you'd have to know it was basic before you asked for the type.  Or, I suppose you could have a using type = statement and have the non-basic object use void or similar (we don't want a new template expansion for every type, just for the ones that need to know their type to do get and set).  For me, I'd just as soon call a virtual function to know if the ObjectMap has the features I want.  I don't think doing it at compile time buys you much; this is not a performance sensitive path so doing the branches at runtime is not an issue. As far as code goes, I either override a function or add a template specialization. Not much difference in how much extra code you need.

The one place that having a compile time evaluation could be useful is that you may be able to have a single template expansion work for the non-arithmetic objects that are declared as basic (this already exists for the arithmetic types).  But, there's no guarantee what their APIs look like, so I'd just as soon write a custom class to handle them, then I know they're being handled exactly the way I want them to be instead of in some generic way.

The other thing to consider is that we're thinking we may need to differentiate beyond basic versus non-basic objects for some of the advanced features we're considering.  In this case, there may be other is* functions and I don't want to start doing a templated attribute for each of those. Or, you may have a non-basic type, but you are also allowed to set it with a string, or you can print a string representation of it as a whole (isSettable()/isGettable() maybe).  I think we're likely to move to an API that has virtual functions that lets the tool know which features that object supports.  The isBasic() function is likely to go away in favor of the more granular API.

I wouldn't worry about an is_basic<T>, as it is likely to be obsolete in the near future.

leekillough:
I have changed the code to use the word "basic" instead of "fundamental", and added `is_basic` and `is_basic_v` as traits tests for it.

- Use `std::is_arithmetic_v` instead of `std::is_fundamental_v` in some tests, since `void` and `std::nullptr_t` were not really applicable to them.
- Rename `fundamental` to `basic` to prevent confusion with C++ fundamental types and `std::is_fundamental`.
- Add `is_basic` and `is_basic_v` templates for testing whether a type is basic (`enum` is included even though it wasn't included in the "fundamental" types).
- Move a `loaded_` class `static` variable definition to inside the class as `inline` because previously it could have had ODR violations.
- Make `from_string()` logic on floating-point types simpler. (Instead of `if float {} else if double {} else if long double {} else { double; make compiler happy }`, use `if float{} else if long double {} else double {}`

leekillough:
`is_basic` was added even if it will become obsolete, because:
1. It is intended to replace the uses of `std::is_fundamental` which were used in mapping corresponding to `isFundamental()`.
2. It is best to define this in one place where practical. If the mapping code will be refactored to where `isBasic()` becomes obsolete, and classes are still defined hierarchically, then a `static constexpr bool isBasic` data member or member function can be defined to indicate whether it is basic or not, and as long as the true type of the class is used and not a base class to where virtual overloads are needed, this member will indicate whether it is basic.
3. It is only 4 lines of simple code.

leekillough:
`std::is_fundamental*` has been replaced with `std::is_arithmetic*`. Although there is partial overlap with "fundamental" `ObjectMap`, "fundamental" `ObjectMap` is distinct from C++ fundamental types.

In `NullStatistic`, `static` data member `loaded_` was made `inline` and defined in-class to avoid ODR violations (multiple definitions).

feldergast (approved):

Issue #1240, opened 2025-03-07
Python command ordering induces I/O error with simulated model outputs


jleidel:
Devel branch w/ upstream JSON I/O changes merged on 03/07/25 from PR #1238 

There is an issue with resolving the statistics `frequency` value when reading in model inputs from Python.  I believe the issue is due to some sort of Python command ordering issue.  

I used the following test for this experiment: `test_StatisticsComponent_basic.py`

I tried to do the following: 
```
sst --output-config=first.py --run-mode=init test_StatisticsComponent_basic.py
sst --output-json=second.json --run-mode=init test_StatisticsComponent_basic.py
sst --output-json=third.json --run-mode=init first.py
diff second.json third.json
```

The diff looks like this: 
```
27d26
<       "frequency": "19 ns",
58d56
<       "frequency": "23 ns",
90d87
<       "frequency": "27 ns",
121d117
<       "frequency": "29 ns",
```

Notice that in the `third.json` the `frequency` is missing.  This comes from the SST input when you translate from Python to Python.

gvoskuilen:
There is an additional issue where stats are not correctly enabled with global enables in the generated config file. E.g.,
```
sst --output-config=first.py test_StatisticsComponent_basic.py > orig.out
sst first.py > first.out
diff orig.out first.out
```
The diff shows globally enabled stats are missing from first.out and several stats are diffing as well.

gvoskuilen:
#1278 has a partial fix for the underlying issues and python output. A fix for JSON output is still needed. On the python side I added support for enabling all statistics in the component as well as to handle shared stats within a component and its children.

jleidel:
Fixed in [37d1c19](https://github.com/sstsimulator/sst-core/commit/37d1c196865370123861ac486b4adc39832ef88d) and [6fc989d](https://github.com/sstsimulator/sst-core/commit/6fc989d656017b5c2221074e7c9d6d909f725630)

PR #1286, opened 2025-04-08
adding shared stats writing from JSON; Fix for JSON portion of Issue #1240

jleidel:
Adds shared statistics writing from the JSON config writer.  Takes a similar approach to the Python writer.  This is the remainder of the fix for Issue #1240


feldergast:
The failed test is due to a system upgrade on the failing tester that caused one of the third-party libraries to no longer compile.  I'll send this for retest once that's sorted.

PR #1290, opened 2025-04-10
Add testing_check_is_nightly() function

jmlapre:
Will be required when turning off long-running tests in sst-elements.

### DIFF
--- a/src/sst/core/testingframework/sst_unittest_support.py
+++ b/src/sst/core/testingframework/sst_unittest_support.py
@@ -75,6 +75,15 @@ class SSTTestCaseException(Exception):
 # Commandline Information Functions
 ################################################################################
 
+def testing_check_is_nightly() -> bool:
+    """If Nightly is in the BUILD_TAG it's very likely a Nightly build."""
+    build_tag = os.getenv("BUILD_TAG")
+    if build_tag and "Nightly" in build_tag:
+        return True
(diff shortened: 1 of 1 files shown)


src/sst/core/testingframework/sst_unittest_support.py
```diff
@@ -75,6 +75,15 @@ def __str__(self) -> str:
 # Commandline Information Functions
 ################################################################################
 
+def testing_check_is_nightly() -> bool:
+    """If Nightly is in the BUILD_TAG it's very likely a Nightly build."""
+    build_tag = os.getenv("BUILD_TAG")
+    if build_tag and "Nightly" in build_tag:
```
berquist, line 81:
```suggestion
    if build_tag is not None and "Nightly" in build_tag:
```

berquist (changes requested):
One minor nit

src/sst/core/testingframework/sst_unittest_support.py
jmlapre: That's a little redundant don't you think?

jmlapre: If `build_tag` was `None` then the `if` statement would fail...

gvoskuilen (approved):

PR #1293, opened 2025-04-14
Serialize arrays, making the code much simpler and more general

leekillough:
This rewrites the array serialization, adding support for `std::array` and dynamically allocated arrays of types besides arithmetic and `enum`.

It still supports the existing wrapper classes `SST::Core:Serialization::array(ptr, size)` and `SST::Core::Serialization::raw_ptr(ptr)` which, when serialized, serialize a dynamic array and a `void*` raw pointer, respectively.

`ser& SST::Core::Serialization::array(ptr, size)` during deserialization, will get `size` and allocate `ptr` with `new ELEM_T[size]` where `ELEM_T` is the type that `ptr` points to.

`std::array<ELEM_T,SIZE>` and `ELEM_T[SIZE]`, and pointers to them, are handled with the same templated class. A pointer to a fixed array, like `int (*aptr)[10]`, will be allocated when deserialized.

-----

The old `serializer::array()` function was renamed `serializer::raw()` and no longer uses the parameter syntax `array(T a[N])` because `a` is turned into a pointer, and it cannot match array arguments that way (it would need to use `array(T (&a)[N])` to match a reference to an array). `serializer::raw()` just takes a `void*` and `size_t` and reads/writes the raw data.

-----

Right now the code potentially leaks because it does not call `delete` or `delete[]` on the original pointer before calling `new` or `new[]` to set it. This may need to be fixed but if the pointer is uninitialized, it could pose a problem. If the serialization documentation says that pointers must be initialized to `nullptr` or some value returned by `new` or `new[]` before serializing them, this would help.

-----

The code has been written in such a way as to reduce code size across a large number of types and fixed array sizes, by separating out the parts which depend on the element type and the array size. Where there is a template dependency on the element type or a fixed array size (which is a part of the type), it is necessary to include templated code, and trying to split it out much further would not be beneficial -- I've tried abstracting it out in different ways, and this current way produces a small number of Lines of Code (LOC) while still not making template instantiations grow the code size too large. `if constexpr` is used to make parts of the code conditional and optimized out where possible.

-----

All existing Core and Elements test pass; I did not see where in the tests where array serialization was tested. We may need to add it, particularly for `std::array`.

-----

An `ObjectMapContext` class has been created which can be used to push options for mapping mode on the stack without having to use `std::stack` or any other containers. A pointer in the `serializer` class points to the current `ObjectMapContext`. The creation of a new `ObjectMapContext` points the serializer to it and saves the old context. When destroyed, it automatically restores the old context. A spurious compiler warning about pointers to local variables had to be suppressed (I verified a Bugzilla was submitted against it on GCC 13, which is my current compiler).
```c++
template <class T>
void
sst_map_object(serializer& ser, T& t, std::string_view name = "")
{
    if ( ser.mode() == serializer::MAP ) {
        if ( !name.empty() ) { // Do nothing if name is empty
            // A new ObjectMapContext is created with name and ser points to it
            ObjectMapContext c(ser, name); 
            serialize<T>()(t, ser);
           // c is destroyed and ser's old context is restored
        }
    }
    else {
        serialize<T>()(t, ser);
    }
}
```
Even though `ObjectMapContext` is lightweight and doesn't dynamically allocate memory (except for `std::string`), it still does not need to be used in the non-mapping mode, as shown in the code above. Besides the `name` string, other attributes can be added later and will be accessible through accessor functions in the `serializer` class without having to pass them in every serialization call. Even if you have a bitwise `flags` option passed in macros, it doesn't have to be passed in every function call, just passed to an `ObjectMapContext` constructor and then adding a method to return it (if attributes like flags apply to more than mapping mode, a different name like `SerializerContext` can be used). If performance is a concern with multiple pointer dereferences, `ObjectMapContext` can store any attributes passed to it directly into the `serializer` class (and restore the old one upon destruction).

-----

Because the `SST::Core::Serialization::array(ptr, size)` wrapper class is passed at the time of serialization and is an Rvalue, universal/forwarding references are used, but within the serialization code, the reference is treated as an Lvalue reference so that functions which expect Lvalue references can be called:
```c++
template<typename T>
void operator&(serializer& ser, T&& t)    // T may be plain type or Lvalue reference to it; t is a Rvalue or Lvalue reference
{
   // t is used locally as an Lvalue to call functions which expect Lvalue references
}
```
As the serialization code is changed to use macros, the functions that get called need to support serializing Rvalue references as well as Lvalue references in case wrappers are used. The old code used a kludgy workaround by providing its own overloads for `operator&(ser, SST::Core::Serialization::array ary)` which accepted wrapper classes by value instead of the usual Lvalue reference. But with universal/forwarding references at the top level, and then using the reference name as an Lvalue, that's not necessary.

Cc: @kpgriesser

### DIFF
--- a/src/sst/core/serialization/impl/serialize_array.h
+++ b/src/sst/core/serialization/impl/serialize_array.h
@@ -20,160 +20,150 @@
 #include "sst/core/serialization/serializer.h"
 
 #include <array>
+#include <cstddef>
+#include <string>
 #include <type_traits>
 
 namespace SST::Core::Serialization {
+
+// Serialize arrays
+
 namespace pvt {
 
-template <class TPtr, class IntType>
-class ser_array_wrapper
+// Wrapper classes. They have no declared constructors so that they can be aggregate-initialized.
+template <typename ELEM_T, typename SIZE_T>
+struct array_wrapper
 {
-public:
-    TPtr*&   bufptr;
-    IntType& sizeptr;
-    ser_array_wrapper(TPtr*& buf, IntType& size) : bufptr(buf), sizeptr(size) {}
+    ELEM_T*& ptr;
+    SIZE_T&  size;
 };
 
-// template <class IntType>
-// class ser_buffer_wrapper
-// {
-// public:
-//     void*& bufptr;
-//     IntType& sizeptr;
-//     ser_buffer_wrapper(void*& buf, IntType& size) :
-//         bufptr(buf), sizeptr(size) {}
-// };
-
-template <class TPtr>
-class raw_ptr_wrapper
+template <typename ELEM_T>
+struct raw_ptr_wrapper
 {
-public:
-    TPtr*& bufptr;
-    raw_ptr_wrapper(TPtr*& ptr) : bufptr(ptr) {}
+    ELEM_T*& ptr;
+};
+
+// Functions for serializing arrays element by element
+void serialize_array(
+    serializer& ser, void* data, size_t size, void serialize_array_element(serializer& ser, void* data, size_t index));
+
+void serialize_array_map(
+    serializer& ser, void* data, size_t size, ObjectMap* map,
+    void serialize_array_map_element(serializer& ser, void* data, size_t index, const std::string& name));
+
+// Serialize an array element
+// Separated out to reduce code size
+template <typename ELEM_T>
+void
+serialize_array_element(serializer& ser, void* data, size_t index)
+{
+    ser& static_cast<ELEM_T*>(data)[index];
+}
+
+// Return a new map representing canonical fixed sized array (whether the original array came from ELEM_T[SIZE] or
+// std::array<ELEM_T, SIZE>)
+template <typename ELEM_T, size_t SIZE>
+ObjectMap*
+new_fixed_array_map(void* data)
+{
+    return new ObjectMapContainer<ELEM_T[SIZE]>(static_cast<ELEM_T(*)[SIZE]>(data));
+}
+
+// Serialize an array map element
+// Separated out to reduce code size
+template <typename ELEM_T>
+void
+serialize_array_map_element(serializer& ser, void* data, size_t index, const std::string& name)
+{
+    sst_map_object(ser, static_cast<ELEM_T*>(data)[index], name);
+}
+
+// Whether the element type is copyable with memcpy()
+// TODO: Implement with std::is_trivially_copyable and std::is_aggregate, using reflection to check for troublesome
+// members like pointers
+template <typename T>
+constexpr bool is_trivial_element_v = std::is_arithmetic_v<T> || std::is_enum_v<T>;
+
+// Serialize fixed arrays
+template <typename OBJ_TYPE, typename ELEM_T, size_t SIZE>
+struct serialize_impl_fixed_array
+{
+    void operator()(OBJ_TYPE& ary, serializer& ser)
+    {
+        const auto& aPtr = get_ptr(ary);
+        switch ( ser.mode() ) {
+        case serializer::MAP:
+            serialize_array_map(
+                ser, &(*aPtr)[0], SIZE, new_fixed_array_map<ELEM_T, SIZE>(&(*aPtr)[0]),
+                serialize_array_map_element<ELEM_T>);
+            break;
+
+        case serializer::UNPACK:
+            if constexpr ( std::is_pointer_v<OBJ_TYPE> ) {
+                if constexpr ( std::is_same_v<OBJ_TYPE, ELEM_T(*)[SIZE]> )
+                    ary = new ELEM_T[SIZE];
+                else
+                    ary = new std::remove_pointer_t<OBJ_TYPE>;
+            }
+            [[fallthrough]];
+
+        default:
+            if constexpr ( is_trivial_element_v<ELEM_T> )
+                ser.raw(&(*aPtr)[0], sizeof(ELEM_T) * SIZE);
+            else
+                serialize_array(ser, &(*aPtr)[0], SIZE, serialize_array_element<ELEM_T>);
+            break;
+        }
+    }
 };
 
 } // namespace pvt
-/** I have typedefing pointers, but no other way.
- *  T could be "void and TPtr void* */
-template <class TPtr, class IntType>
-inline pvt::ser_array_wrapper<TPtr, IntType>
-array(TPtr*& buf, IntType& size)
+
+// Serialize fixed arrays and pointers to them
+template <typename ELEM_T, size_t SIZE>
+struct serialize_impl<ELEM_T[SIZE]> : pvt::serialize_impl_fixed_array<ELEM_T[SIZE], ELEM_T, SIZE>
+{};
+
+template <typename ELEM_T, size_t SIZE>
+struct serialize_impl<std::array<ELEM_T, SIZE>> :
+    pvt::serialize_impl_fixed_array<std::array<ELEM_T, SIZE>, ELEM_T, SIZE>
+{};
+
+template <typename ELEM_T, size_t SIZE>
+struct serialize_impl<ELEM_T (*)[SIZE]> : pvt::serialize_impl_fixed_array<ELEM_T (*)[SIZE], ELEM_T, SIZE>
+{};
+
+template <typename ELEM_T, size_t SIZE>
+struct serialize_impl<std::array<ELEM_T, SIZE>*> :
+    pvt::serialize_impl_fixed_array<std::array<ELEM_T, SIZE>*, ELEM_T, SIZE>
+{};
+
+// Serialize dynamic arrays
+template <typename ELEM_T, typename SIZE_T>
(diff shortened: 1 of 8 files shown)


src/sst/core/serialization/serializer.h
```diff
@@ -290,8 +260,31 @@ class serializer
...
+    const ObjectMapContext*        mapContext = nullptr;
+    friend class ObjectMapContext;
+}; // class serializer
+
+/**
+   ObjectMap context which is saved in a virtual stack when name or other information changes.
+   When ObjectMapContext is destroyed, the serializer goes back to the previous ObjectMapContext.
+ */
+
+class ObjectMapContext
```
feldergast, line 272: This seems like a good replacement for the name stack.  However, the plan is to still also pass an integer options value to each serialize_impl call as a bitwise map of options.  These options are needed in SIZE, PACK and UNPACK mode and we're not willing to create a context object in those modes as they are used in the serialization for synchronization fast path.

src/sst/core/serialization/impl/serialize_array.h
```diff
@@ -20,200 +20,197 @@
...
-    template <class A>
-    friend class serialize;
-    void operator()(pvt::ser_array_wrapper<void, IntType> arr, serializer& ser)
+    void operator()(pvt::array_wrapper<ELEM_T, SIZE_T>& ary, serializer& ser)
     {
-        ser.binary(arr.bufptr, arr.sizeptr);
-
-        // TODO: Implement mapping mode
+        switch ( const auto mode = ser.mode() ) {
+        case serializer::MAP:
```
feldergast, line 149: Can you add this mode?  I know of several dynamically allocated arrays that will need this feature immediately.

src/sst/core/serialization/impl/serialize_array.h
```diff
@@ -20,200 +20,197 @@
...
-    friend class serialize;
-    void operator()(pvt::raw_ptr_wrapper<TPtr> ptr, serializer& ser)
+    void operator()(pvt::raw_ptr_wrapper<ELEM_T>& a, serializer& ser)
     {
-        ser.primitive(ptr.bufptr);
-
-        // TODO: Implement mapping mode
+        switch ( ser.mode() ) {
+        case serializer::MAP:
+            // TODO: Implement mapping mode
```
feldergast, line 181: Mapping mode does not not need to be implemented for raw pointers. If it's being serialized as just the value of the pointer, then we don't want to follow it to map the data it points to.

feldergast (changes requested):

feldergast:
Even though changers are requested, I'm going to release this for testing to make sure it is compatible with sst-macro.

src/sst/core/serialization/impl/serialize_array.h
```diff
@@ -20,200 +20,197 @@
...
-inline pvt::ser_array_wrapper<void, IntType>
-buffer(void*& buf, IntType& size)
+// Serialize an array element
+// Separated out to reduce code size
+template <typename ELEM_T>
+void
+serialize_array_element(serializer& ser, void* data, size_t index)
 {
-    return pvt::ser_array_wrapper<void, IntType>(buf, size);
+    ser& static_cast<ELEM_T*>(data)[index];
```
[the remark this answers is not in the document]
leekillough: I know but I didn't want to use it until everything was settled as far as macro names, etc. Should a macro be called or a function, etc.? Is there a different function/macro to be called or a different argument to use when doing sub-objects which don't need annotation, etc.?

I was waiting for all of this stuff to settle.

`sst_map_object()` seems very map-centric and does not sound like a general serialization function.

src/sst/core/serialization/impl/serialize_array.h
```diff
@@ -20,200 +20,197 @@
...
+template <typename T>
+constexpr bool is_trivial_element_v = std::is_arithmetic_v<T> || std::is_enum_v<T>;
 
-/**
-   Version of serialize that works for statically allocated arrays of
-   non base types.
- */
-template <class T, size_t N>
-class serialize_impl<T[N], std::enable_if_t<!std::is_arithmetic_v<T> && !std::is_enum_v<T>>>
+// Serialize fixed arrays
```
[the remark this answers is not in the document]
leekillough: I can add comments, but I thought it was obvious.

src/sst/core/serialization/serializer.h
leekillough: I think the flags should be a mostly-opaque type, such as `SerFlag`. I would not want `uint32_t` or `uint64_t` to be exposed in the API as the type of the flags. There should be a separate `using` (like`typedef`) or `enum class` for the flags, using whatever integer underlying type wanted. The names should be prevented from collision with other names, such as putting them in an `enum class` or an `enum` inside the public section of a class, or a namespace with a short name like `Ser`.

Like how [`RevFlag` is implemented](https://github.com/tactcomplabs/rev/blob/devel/include/RevFlag.h).

src/sst/core/serialization/impl/serialize_array.h
leekillough: I can add mapping mode but I was waiting to see how the `ObjectMap` should be implemented. We have a container `ObjectMap` -- should arrays just use `ObjectMapContainer` or create a `ObjectMapArray` class which inherits from `ObjectMapContainer`?  How will we store the array size in it, and how does the size stored change w.r.t. serialization (I mean, how to keep the `ObjectMapArray` class or whatever in-sync with the array address and size in the program? Or do we only need the `ObjectMapArray` to live the life of a single serialization call and then it can be destroyed and none of the array or other information stored in it matters antmore?).

src/sst/core/serialization/impl/serialize_array.h
leekillough: I can remove that comment then and clean up the code. The previous code had a `TODO` about mapping mode in the raw pointers serialization, but I wasn't sure what needed to be done.

src/sst/core/serialization/serializer.h
feldergast: That's how I've got them implemented.

src/sst/core/serialization/serialize.h
```diff
@@ -336,32 +336,39 @@ class serialize_impl<T*, std::enable_if_t<std::is_arithmetic_v<T> || std::is_enu
 // pointer tracking can be done.
 template <class T>
 void
-sst_map_object(serializer& ser, T& t, std::string name = "")
+sst_map_object(serializer& ser, T&& obj)
```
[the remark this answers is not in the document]
feldergast: The macro interface will always pass a name, so having a version without name is not needed.  

With the new API, there is an option to control whether mapping of an object happens or not, so the name variable should never need to be tested unless you are checking for errors.  So, passing either "" or nullptr should work, but really, the only time the function would be called directly is if you want to have a different name for mapping than the variable name in the class, in which case you wouldn't pass either of those.

I've already changed the name to sst_ser_object() to mirror the SST_SER macro.

src/sst/core/serialization/impl/serialize_array.h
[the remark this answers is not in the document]
feldergast: You can just use SST_SER.

feldergast:
Since this passed testing and we're getting close to release, I'm going to just merge this and we can revisit the comments in another PR.  Getting this merged now will let us start testing functionality (I'll add the tests to my next PR since I'm adding a lot of other tests to the serialization test suite).  There will be some final cleanup across all of the serialization we'll probably want to do as we finalize the APIs and we can address these then.

feldergast (approved):
Looks good. Can address comments in a future PR. See prior comment regarding this.

leekillough:
> Even though changers are requested, I'm going to release this for testing to make sure it is compatible with sst-macro.

Do you want me to merge `sst-macro` into this branch, resolving conflicts and making sure that it conforms to the new macro/function call syntax? Where is `sst-macro`?

feldergast:
> > Even though changers are requested, I'm going to release this for testing to make sure it is compatible with sst-macro.
> 
> Do you want me to merge `sst-macro` into this branch, resolving conflicts and making sure that it conforms to the new macro/function call syntax? Where is `sst-macro`?

No.  That's an unfortunate name clash.  sst-macro is a separate repository. It's a network simulator that can use set-core as its pdes framework.  It isn't a branch of this repo, which I suspect is what you were thinking.

I'll do the merge with my work and figure out all the clashes.  I had also already made the array() function work for general types and based a lot of what I was doing off of that.  Once I get everything merged and all the tests added, it would be great if you could review things.

leekillough:
> Since this passed testing and we're getting close to release, I'm going to just merge this and we can revisit the comments in another PR. Getting this merged now will let us start testing functionality (I'll add the tests to my next PR since I'm adding a lot of other tests to the serialization test suite). There will be some final cleanup across all of the serialization we'll probably want to do as we finalize the APIs and we can address these then.

Please hold off. I have more commits in response to your comments, including the name string.

leekillough:
I will reopen another PR

PR #1295, opened 2025-04-14
Serialize arrays, continued

leekillough:
Continuing from https://github.com/sstsimulator/sst-core/pull/1293, which IMO was merged prematurely.

- More comments were added, to make the code more clear
- The names used for mapping were changed to `const char*` to avoid the overhead of `std::string`. There is an assumption that the string pointed to will be valid throughout the lifetime of the serialization call, which is true even if it is a temporary expression like `std::to_string(i).c_str()`, but if it needs to be stored somewhere after the serialization call, such as in an `ObjectMap`, then a `std::string` or other copy of it will need to be made.
- A `nullptr` was used instead of an empty string `""` for when the string argument is absent. This avoids a pointer dereference to check if the first byte of the string is `'\0'`, and simply tests if the pointer is `nullptr`.
- The `sst_ser_object()` and `sst_ser_object_as_ptr()` functions were added, and macros defined:
```c++
#define SST_SER(obj)        sst_ser_object(ser, (obj), #obj)
#define SST_SER_AS_PTR(obj) sst_ser_object_as_ptr(ser, (obj))
```
- The `ser&` syntax was marked deprecated, and all calls were replaced by a command like this (including this so that you can replicate it elsewhere if you reject this PR):
```
find . -type f -exec grep -q '\bser *&' {} \; -exec perl -pi -e 's/\bser\s*&\s*([^;]+)/SST_SER($1)/g' {} \;
```
- Semicolons were removed from the expansion of the `SST_SER()` macro, because:
    - Calling `SST_SER()` without a terminating semicolon messes with auto-indentation from editors and `clang-format`, because it looks like an unterminated statement
    - They do not work well with `if` statements, since adding an extra semicolon makes this invalid;
```c++
   if (test) 
     SST_SER(x);     <--- if the macro expands with ; this extra semicolon makes the `else` clause invalid
   else
     SST_SER(y);
```

- They do not allow expansion of the macro in C++17 fold expressions, such as this from `serialize_tuple.h`:
```c++
        // Serialize each element of tuple or pair
        std::apply([&](auto&... e) { ((SST_SER(e)), ...); }, t);
```
- A `DISABLE_WARN_DANGLING_POINTER ` warning macro was added to suppress a GCC 13 false warning. It was noticed that when a warning macro is disabled because of the compiler version, the warning macro was expanded to empty, but the `REENABLE_WARNING` macro was still popping from the compiler option stack, popping from an empty stack or removing a previous warning prematurely. The warning macros which should do nothing because of the compiler version were changed to expand to `DIAG_PRAGMA(DIAG_COMPILER, push)` so that the compiler options stack is pushed for its corresponding `REENABLE_WARNING` macro to pop it, even though warnings aren't changed.
- An `ObjectMapArray` class derived from `ObjectMapContainer` was created, adding a runtime size in addition to the pointer in `ObjectMapContainer`. It is too late and drastic to remove the type from the template now.
- Mapping mode was added to the dynamically allocated array serialization which uses `array` wrappers. When the pointer is `void*`, mapping mode is a no-op because there is no way to access array elements until it is cast to a complete type. If `void*` buffers need to be supported in mapping mode, this will require more work.


leekillough:
More changes since the API overhaul:

Make `SerOption` an `enum class` instead of a `namespace`
- Removes unnecessary `ser_opt_t` type, since `SerOption` is now the type name
- Provides the same `SerOption::name` syntax for values
- Improves type safety by preventing implicit conversions between `SerOption` and integers
- `SerOption::none` is the value for no options set, and is more self-documenting than using `0`

Add `operator|()` for `SerOption` to bitwise-OR several `SerOption` values together
- Argument-Dependent Lookup (ADL) finds it in the namespace that `SerOption` is defined in

Change verbose, typing-unfriendly `isSerOptionSet()` to simpler `hasOption()`
- `hasOption(opt, has)` returns true if `opt` contains **all** options contained in `has`
- `hasOption()` is currently placed in `SST::Core::Serializer` but it could be placed in the same namespace as `SerOption` and ADL would find it there

Use `opt` instead of longer names `opts`, `option`, `options` for `SerOption` parameters
- Similar to `ser` short name
- `SerOption opt` parameter declaration makes is pretty clear that it's an option

Make `opt` the third parameter of `sst_ser_object()` and `name` the fourth parameter, and make `SerOption::none` the default `opt` argument and `nullptr` the default `name` argument
```c++
void sst_ser_object(serializer& ser, T&& obj, SerOption opt = SerOption::none, const char* name = nullptr);
```
- Makes all functions/macros have canonical order (`ser`, `obj` [, `opt` [, `name` ] ])
- Since `opt` is used everywhere but `name` is not, `opt` should come before `name`
- Avoids having to pass `name` argument if `opt` argument is non-default
- `nullptr` `name` instead of `""` is faster to test for (one less memory indirection), and it is more likely to produce a runtime error if `name` is used when it is `nullptr`

For `SerOption::as_ptr_elem` semantics, define an `elem_opt` variable which sets `SerOption::as_ptr`, and use that for serializing container elements which should be treated as pointers, but use the unchanged `opt` parameter passed to the function everywhere else
```c++
SerOption elem_opt = hasOption(opt, SerOption::as_ptr_elem) ? opt | SerOption::as_ptr : opt;
```
- Sets `as_ptr` option if `as_ptr_elem` option is set but leaves all other options unchanged, making it future-proof
- Defines `elem_opt` variable to use for element options but leaves `opt` for use elsewhere
- If certain options need to be cleared in `elem_opt`, they can, but the default should be to pass options unchanged for future-proofing. If clearing options is needed, a function similar to `hasOption()` like `clearOption()` should be written, rather than `operator&()`.

Simplify `sst_ser_object()`
- A `serialize` type alias was used to reduce repetition of long expressions
```c++
    using serialize = pvt::serialize<std::remove_reference_t<T>>;
```
- In the `if`/`else` logic, each final action is only listed once, instead of repeated in different decision tree paths

Simplify `sst_ser_or_helper()` to use _binary_ fold expressions which can work with size 0 (_unary_ fold expressions only work with size 0 for the `&&`, `||` and `,` (comma) operators, where a default value is clearly defined)
```c++
return (SerOption::none | ... | args); // Fold expression to perform logical OR
```

Pass options to the array serialization functions, and handle `as_ptr_elem`

-------

Some tests fail for mysterious reasons but I'm investigating

PR #1301, opened 2025-04-18
Deprecate use of shared TimeConverter* objects

gvoskuilen:
Deprecates use of shared TimeConverter* objects in link/clock/simtime calls that use them

Also:
* Adds support for serializing std container pointers 
* Fixes some remaining VLA warnings
* Changes SST_FRIEND_SERIALZE to SST_FRIEND_SERIALIZE

### DIFF
--- a/src/sst/core/baseComponent.cc
+++ b/src/sst/core/baseComponent.cc
@@ -169,7 +169,7 @@ BaseComponent::registerClock(const UnitAlgebra& freq, Clock::HandlerBase* handle
 }
 
 TimeConverter*
-BaseComponent::registerClock(TimeConverter& tc, Clock::HandlerBase* handler, bool regAll)
+BaseComponent::registerClock(TimeConverter tc, Clock::HandlerBase* handler, bool regAll)
 {
     TimeConverter* tcRet = sim_->registerClock(tc, handler, CLOCKPRIORITY);
     registerClock_impl(tcRet, handler, regAll);
@@ -185,7 +185,7 @@ BaseComponent::registerClock(TimeConverter* tc, Clock::HandlerBase* handler, boo
 }
 
 Cycle_t
-BaseComponent::reregisterClock(TimeConverter& freq, Clock::HandlerBase* handler)
+BaseComponent::reregisterClock(TimeConverter freq, Clock::HandlerBase* handler)
 {
     return sim_->reregisterClock(freq, handler, CLOCKPRIORITY);
 }
@@ -290,7 +290,7 @@ BaseComponent::getLinkFromParentSharedPort(const std::string& port, std::vector<
 }
 
 Link*
-BaseComponent::configureLink_impl(const std::string& name, TimeConverter* time_base, Event::HandlerBase* handler)
+BaseComponent::configureLink_impl(const std::string& name, SimTime_t time_base, Event::HandlerBase* handler)
 {
     LinkMap* myLinks = my_info->getLinkMap();
 
@@ -376,11 +376,7 @@ BaseComponent::configureLink_impl(const std::string& name, TimeConverter* time_b
                 }
             }
         }
-
-        if ( nullptr != time_base )
-            tmp->setDefaultTimeBase(time_base);
-        else
-            tmp->setDefaultTimeBase(my_info->defaultTimeBase);
+        tmp->setDefaultTimeBase(time_base);
 #ifdef __SST_DEBUG_EVENT_TRACKING__
         tmp->setSendingComponentInfo(my_info->getName(), my_info->getType(), name);
 #endif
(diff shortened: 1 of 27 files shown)


src/sst/core/baseComponent.cc
```diff
@@ -168,6 +168,14 @@ BaseComponent::registerClock(const UnitAlgebra& freq, Clock::HandlerBase* handle
     return tc;
 }
 
+TimeConverter*
+BaseComponent::registerClock(TimeConverter& tc, Clock::HandlerBase* handler, bool regAll)
```
feldergast, line 172: I'm not entirely sure we need to pass by reference, but if we do, we should probably pass by const reference.

src/sst/core/baseComponent.cc
```diff
@@ -490,10 +524,16 @@ BaseComponent::getSimulationOutput() const
     return Simulation_impl::getSimulationOutput();
 }
 
+SimTime_t
+BaseComponent::getCurrentSimTime(TimeConverter& tc) const
+{
+    return tc.convertFromCoreTime(sim_->getCurrentSimCycle());
+}
+
 SimTime_t
```
feldergast, line 533: May want to move this definition to the header file so it can be inlined.

src/sst/core/timeConverter.h
```diff
@@ -30,6 +30,12 @@ class TimeConverter
     friend class TimeLord;
 
 public:
+    /**
+       Create a new TimeConverter object from a TimeConverter*
+       @param tc TimeConverter to initialize factor from
+     */
+    TimeConverter(TimeConverter* tc) : factor(tc->getFactor()) {}
```
feldergast, line 37: You could just initialize with tc->factor since they are the same class and can access private data members.

src/sst/core/timeConverter.h
```diff
@@ -53,17 +59,29 @@ class TimeConverter
      */
     UnitAlgebra getPeriod() const; // Implemented in timeLord.cc
 
+    /**
+     * TimeConverter* returned by the core should *never* be deleted by
+     * Elements. This was moved to public due to needing to support ObjectMaps.
+     */
+    ~TimeConverter() {}
+    TimeConverter() {} // Only needed to simplify serialization
```
feldergast, line 67: I think the comment is not true; now that people will be using TimeConverter directly, they will need a public default constructor.

feldergast (changes requested):

src/sst/core/baseComponent.cc
```diff
@@ -276,7 +290,7 @@ BaseComponent::getLinkFromParentSharedPort(const std::string& port, std::vector<
 }
 
 Link*
-BaseComponent::configureLink(const std::string& name, TimeConverter* time_base, Event::HandlerBase* handler)
+BaseComponent::configureLink_impl(const std::string& name, TimeConverter* time_base, Event::HandlerBase* handler)
```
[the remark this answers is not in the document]
gvoskuilen: The reason to pass by pointer was so that `configureLink(name, handler)` could then call this with `time_base = nullptr`. We could have it pass `defaultTimeBase` instead but it would have to construct a TimeConverter since `defaultTimeBase` is a `nullptr`. You'd also have to have the configureLink() that takes a TC pointer check if the pointer is null. Alternately, since we're moving that direction anyways, this function could take a SimTime_t instead of a TimeConverter and `configureLink` would convert.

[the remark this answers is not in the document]
feldergast: Since this isn't a public function, I like the idea of passing SimTime_t since that's what it gets converted to in the end.

src/sst/core/baseComponent.cc
gvoskuilen: I did not make this change because it would require `sim_` to be accessible in the header.

src/sst/core/timeConverter.h
gvoskuilen: Added a new comment

src/sst/core/baseComponent.cc
```diff
@@ -362,34 +376,52 @@ BaseComponent::configureLink(const std::string& name, TimeConverter* time_base,
...
         tmp->setSendingComponentInfo(my_info->getName(), my_info->getType(), name);
 #endif
     }
     return tmp;
 }
 
+Link*
+BaseComponent::configureLink(const std::string& name, TimeConverter* time_base, Event::HandlerBase* handler)
+{
+    // Lookup core-owned time_base in case it differs from the one passed in (unlikely but possible)
```
feldergast, line 390: Comment does not match code

src/sst/core/simulation.cc
```diff
@@ -1112,6 +1112,22 @@ Simulation_impl::registerClock(const UnitAlgebra& freq, Clock::HandlerBase* hand
     return registerClock(tcFreq, handler, priority);
 }
 
+TimeConverter*
+Simulation_impl::registerClock(TimeConverter& tc_freq, Clock::HandlerBase* handler, int priority)
```
feldergast, line 1116: Should be TimeConverter instead of TimeConverter& for this function and reregisterClock(), getNextClockCycle() and unregisterClock()

feldergast (approved):
Looks good.  Just a few things we should cleanup as we finalize things for release.

PR #1302, opened 2025-04-18
Fix enum warnings and make SerOption tests uniform

leekillough:
This fixes warnings about mixing `enum` and integers on GCC, and makes the `SerOption` tests uniform.

### DIFF
--- a/src/sst/core/serialization/impl/serialize_array.h
+++ b/src/sst/core/serialization/impl/serialize_array.h
@@ -85,7 +85,7 @@ struct serialize_impl_fixed_array
 {
     void operator()(OBJ_TYPE& ary, serializer& ser, ser_opt_t opt)
     {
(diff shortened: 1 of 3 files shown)


leekillough:
```
../../../../src/sst/core/serialization/impl/serialize_tuple.h:35:60: warning: enumerated and non-enumerated ession [-Wextra]
   35 |         ser_opt_t opt = (options & SerOption::as_ptr_elem) ? SerOption::as_ptr : 0;
      |                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
```

feldergast (approved):

feldergast:
Thanks, I thought I found all these but my compiler doesn't warn for this so it looks like I missed a few.

PR #1304, opened 2025-04-21
Explicit

leekillough:
Do not squash-merge this. It has commit ID for `.git-blame-ignore-revs` in it.


feldergast:
Closing.  We will do the .git-blame-ignore-revs update in a separate PR.

PR #1305, opened 2025-04-21
Override

leekillough:
Do not squash-merge this. It has commit info for `.git-blame-ignore-revs`.


feldergast:
Closing.  We will do the .git-blame-ignore-revs update in a separate PR.

PR #1308, opened 2025-04-22
Add script to check for missing #includes and automatically fix them; run them on the latest sources

leekillough:
The code is a Perl script with a built-in manpage. It requires GCC to work, because it parses GCC output and Clang does not produce the same messages. The script has safeguards in case GCC cannot be found, it cannot understand GCC's output, or something goes wrong with the script, such as it not knowing where in a file to insert the `#include`s.

Please try the script on other systems besides x86-64 Ubuntu 24.04 Linux, which I tested it on.


feldergast:
Thanks for putting this together.  At this point, we'll need to look at this after the release.

PR #1311, opened 2025-04-22
Add support for shutdown from interactive console

skuntz:
Experimental - not required for 15.0 release

Add support to interactiveConsole for an interactiveShutdown function that exits the interactive console and does a clean shutdown of the simulation. This is modeled after the signalShutdown function in simulation.cc and tested in the simpleDebug prototype. 
---

----

### DIFF
--- a/src/sst/core/interactiveConsole.cc
+++ b/src/sst/core/interactiveConsole.cc
@@ -111,4 +111,11 @@ InteractiveConsole::getComponentObjectMap()
     return Simulation_impl::getSimulation()->getComponentObjectMap();
 }
 
+void
+InteractiveConsole::simulationShutdown()
+{
+    Simulation_impl::getSimulation()->interactiveShutdown();
+}
+
+
 } // namespace SST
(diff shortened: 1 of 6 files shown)


src/sst/core/interactiveConsole.cc
```diff
@@ -111,4 +111,11 @@ InteractiveConsole::getComponentObjectMap()
     return Simulation_impl::getSimulation()->getComponentObjectMap();
 }
 
+void
+InteractiveConsole::simulationShutdown()
+{
+    Simulation_impl::getSimulation()->interactiveShutdown();
```
feldergast, line 117: I think I would just call endSimulation() here instead of introducing a new function.  endSimulation() has the added benefit of setting the simulation end time to the current time, though I'm not sure if that's strictly necessary if ending a simulation early.

As a side note, when we enable parallel debugging, we'll need to also notify all the other ranks/threads to shutdown. At that point, we may find it easiest to add a new shutdown function, but let's wait until then and reevaluate.

skuntz: Updated to use endSimulation().

feldergast (approved):

PR #1313, opened 2025-04-23
Serialize array bug

feldergast:
Fix ordering issue in call to sst_ser_obj and fix documentation on checkpoint-prefix.


leekillough:
LGTM. The reason this was not caught sooner as an argument mismatch error:
```c++
sst_ser_object(ser, static_cast<ELEM_T*>(data)[index], opt, nullptr);
```
... is because `ELEM_T` makes the _entire expression_ a [dependent expression](https://en.cppreference.com/w/cpp/language/dependent_name#Type-dependent_expressions), so it won't be checked for semantic errors until the template is instantiated and template parameter `ELEM_T` is set. Testing is limited (we need more), so this was not caught sooner.

gvoskuilen (approved):

PR #1316, opened 2025-04-24
Fix some warnings found in gcc 8.5-14.2

gvoskuilen:
Two types of compile warnings addressed: unused variables and missing override in functions marked final. 

While the missing override does not warn in all newer versions of gcc, it does warn in older ones that are still supported.

### DIFF
--- a/src/sst/core/action.h
+++ b/src/sst/core/action.h
@@ -29,7 +29,7 @@ public:
     Action() {}
     ~Action() {}
 
-    bool isAction() final { return true; }
+    bool isAction() override final { return true; }
 
 protected:
     /** Called to signal to the Simulation object to end the simulation */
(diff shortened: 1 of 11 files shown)


leekillough:
`final` without `virtual` implies `override`, so the newer compilers are correct not to warn about `override` being missing.

Here are the combinations of `virtual`, `override` and `final` (8 combinations of their each being present or not):

- `virtual`: Defines a function which is virtual from this point downwards in the inheritance hierarchy, possibly overriding a base class function, but not required to
- `virtual override`, `override`: Virtual function which must override a virtual base class function. `override` implies `virtual` so `virtual` is redundant but may used for documentary reasons
- `final`, `final override`, `virtual final override`: Defines a function which is final and must override a base class virtual function
- `virtual final`: Defines a virtual function which is final and _may_ override a base class virtual function, but isn't required to.
- (nothing): Defines a function which _may_ be virtual and _may_ override a base class virtual function, but isn't required to.

gvoskuilen:
Completely agree, and once we move past 8.5 we can remove the redundant statement. It doesn't seem to cause a problem to have the redundant `override final` in any of the compilers I tested on.

feldergast (approved):

PR #1319, opened 2025-04-28
add array serialization tests, fix bugs

leekillough:
This adds array serialization tests, and fixes two bugs:
1. `serializer::binary( buffer, size )` did not properly compute `SIZER` with `size * sizeof(ELEM_T)` -- it only used `size` for the `SIZER` computation of the array size. **This should be patched into SST 15 if possible.**
2. If a pointer to a fixed-size array, such as `int (*ary)[10]` was serialized, it was not properly allocated, leading to compilation errors in the template instantiation of `serialize_impl`.

The `serializeDeserialize()` function was modified slightly to take universal references, so that it can work with `array(ptr, size)` wrappers which are rvalue expressions but contain lvalue references inside.

### DIFF
--- a/src/sst/core/action.h
+++ b/src/sst/core/action.h
@@ -29,7 +29,7 @@ public:
     Action() {}
     ~Action() {}
 
-    bool isAction() final { return true; }
+    bool isAction() override final { return true; }
 
 protected:
     /** Called to signal to the Simulation object to end the simulation */
--- a/src/sst/core/baseComponent.cc
+++ b/src/sst/core/baseComponent.cc
@@ -457,7 +457,7 @@ Link*
 BaseComponent::configureSelfLink(const std::string& name, TimeConverter* time_base, Event::HandlerBase* handler)
 {
     addSelfLink(name);
-    return configureLink(name, time_base, handler);
+    return configureLink(name, *time_base, handler);
 }
 
 Link*
--- a/src/sst/core/clock.h
+++ b/src/sst/core/clock.h
@@ -66,7 +66,8 @@ public:
        handler will be left in the clock list.
      */
     template <typename classT, typename dataT = void>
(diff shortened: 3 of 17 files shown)


feldergast (dismissed):

feldergast:
Thanks.  We'll get this into the release.

leekillough:
Note that this won't work if the type is `void` -- I need to add `if constexpr` to handle that case, where I think we just assume an array of `char`.

leekillough:
Oh, wait, we already cast `void*` to `char*`:
```c++
    // For void*, we get sizeof(), which errors.
    // Create a wrapper that casts to char* and uses above
    template <typename Int>
    void binary(void*& buffer, Int& size)
    {
        binary(reinterpret_cast<char*&>(buffer), size);
    }
```

feldergast:
Yea, it was done that way before if constexpr was available.  We can leave it that way or use if constexpr if that's better.

leekillough:
I'd rather leave it that way for release.

`serializer.h` and `serialize.h` can probably be simplified with `if constexpr` and other things after release. For now I want to keep the changes simple.

The path here is executed for arithmetic/enum types. I took the short-circuit logic out temporarily (so that it serialized the array element by element), and it passed on that other path (the one which calls the `serialize_array_element<ELEM_T>` function).

gvoskuilen:
Failures are due to compile error on mac: 
`testElements/coreTest_Serialization.cc:226:18: error: constexpr variable 'array_size' must be initialized by a constant expression`

leekillough:
> Failures are due to compile error on mac: `testElements/coreTest_Serialization.cc:226:18: error: constexpr variable 'array_size' must be initialized by a constant expression`

Fixed.

feldergast (approved):

PR #1330, opened 2025-05-09
Add support for general handling of trivially serializable types

leekillough:
Template metaprogramming to determine if a type is trivially serializable - that it can be read and written as raw data without any special handling.

The criteria are:
* It is one of these types:
    - arithmetic (integral, floating-point)
    - enumeration (including `std::byte`)
    - pointer-to-member-object (not pointer or pointer-to-member-function) 
    - `std::complex<T>`, C99 `_Complex`
    - `std::bitset<N>`
    - trivially copyable, standard layout aggregate type with trivially serializable members and no `serialize_order()` method
 
An aggregate is a C-style array, `std::array`, or a class/struct/union with all public non-static data members and direct bases, no user-provided, inherited or explicit constructors (C++17), no user-declared or inherited constructors (C++20), and no virtual functions or virtual bases.

Pointers are not considered trivially serializable since they have specific addresses which are not portable across runs, and may require special tracking and allocation. But pointers to member objects are typed offsets within a class, and are portable across runs. Pointers to member functions point to a static data structure which is not stable from run to run and is able to handle all contingencies of overriden and virtual functions.

Note: If an aggregate is a union, only the first member will be considered active, and thus cannot be a pointer. Rather than ban unions entirely, pointers in unions are strongly discouraged and can cause unexpected results.

Note: `is_trivially_serializable<T>` should return `true` if `T` is trivially serializable, even if there exists a specialization for `serialize_impl<T>`. `is_trivially_serialzable<T>` is a trait for trivial serializability in general, not a `serialize_impl` specialization test condition. For example, it returns `true` for arrays of `int`, even if `serialize_impl<int[N]>` is specialized.

For mapping mode, this does not change the current behavior, which only implements mapping mode for arithmetic and enumeration types which are trivially serializable. If mapping mode is required for an aggregate class type, a `to_string()` method and custom `serialize_impl` specializations are required.

`std::complex`, C99 `_Complex`, `_Float16` and `__float128` are supported as well, although some of them may need to be `#ifdef`'ed to not be included on certain platforms.

Array serialization uses the `is_trivially_serializable<>` trait, and calls the primitive serialization routines to serialize/deserialize the whole array as one group of bytes, if the element type is trivially serializable.

This uses C++ metaprogramming and reflection methods to determine whether an aggregate type contains any members which are not trivially serializable, such as pointers.

### DIFF
--- a/src/sst/core/serialization/impl/serialize_utility.h
+++ b/src/sst/core/serialization/impl/serialize_utility.h
@@ -17,23 +17,203 @@
     "The header file sst/core/serialization/impl/serialize_utility.h should not be directly included as it is not part of the stable public API.  The file is included in sst/core/serialization/serialize.h"
 #endif
 
+#include <bitset>
+#include <complex>
+#include <cstddef>
 #include <type_traits>
+#include <utility>
 
 namespace SST::Core::Serialization {
 
-// Whether two names are the same template. Similar to std::is_same_v.
-template <template <typename...> class, template <typename...> class>
+/////////////////////////////////////////////////////////////////////////
+// Whether two names are the same template. Similar to std::is_same_v. //
+/////////////////////////////////////////////////////////////////////////
+template <template <class...> class, template <class...> class>
 constexpr bool is_same_template_v = false;
 
-template <template <typename...> class T>
+template <template <class...> class T>
 constexpr bool is_same_template_v<T, T> = true;
 
-// Whether a certain type is the same as a certain class template filled with arguments
-template <class, template <typename...> class>
+template <template <class...> class T1, template <class...> class T2>
+using is_same_template = std::bool_constant<is_same_template_v<T1, T2>>;
+
+///////////////////////////////////////////////////////////////////////////////////////
+// Whether a type is the same as a certain class template filled with type arguments //
+///////////////////////////////////////////////////////////////////////////////////////
+template <class, template <class...> class>
 constexpr bool is_same_type_template_v = false;
 
-template <template <typename...> class T1, typename... T1ARGS, template <typename...> class T2>
+template <template <class...> class T1, class... T1ARGS, template <class...> class T2>
 constexpr bool is_same_type_template_v<T1<T1ARGS...>, T2> = is_same_template_v<T1, T2>;
+
+template <class T, template <class...> class TT>
+using is_same_type_template = std::bool_constant<is_same_type_template_v<T, TT>>;
+
+//////////////////////////////////////////////////
+// Compute the number of fields in an aggregate //
+//////////////////////////////////////////////////
+namespace pvt_nfields {
+
+// glob is convertible to any other type while initializing an aggregate
+// The conversion function is only used in decltype() so it does not need to be defined
+struct glob
+{
+    template <class T>
+    operator T() const;
+};
+
+// Whether N fields can be used to initialize an aggregate, i.e., number of aggregate fields >= N
+// If the type is not an aggregate, always returns false, so that nfields<C> returns 0.
+template <class C, class, bool = std::is_aggregate_v<C>, class = C>
+constexpr bool nfields_ge_impl = false;
+
+// Try to initialize N fields in an aggregate, with glob() automatically converting to each field's
+// type. If initialization fails, it is not an error (SFINAE) and this specialization won't apply.
+template <class C, size_t... I>
+constexpr bool nfields_ge_impl<C, std::index_sequence<I...>, true, decltype(C { (I, glob())... })> = true;
+
+// Whether the number of fields in an aggregate is greater than or equal to N
+template <class C, size_t N>
+constexpr bool nfields_ge = nfields_ge_impl<C, std::make_index_sequence<N>>;
+
+// Binary search in [L,H) for number of fields in aggregate. Stops at L when M == L.
+template <class C, size_t L, size_t H, size_t M = L + (H - L) / 2, bool = M != L>
+constexpr size_t b_search = L;
+
+// Next search is in [L,M)
+template <class C, size_t L, size_t H, size_t M, bool>
+constexpr size_t b_search_next = b_search<C, L, M>;
+
+// Next search is in [M,H)
+template <class C, size_t L, size_t H, size_t M>
+constexpr size_t b_search_next<C, L, H, M, true> = b_search<C, M, H>;
+
+// Choose [L,M) or [M,H) depending on whether nfields >= M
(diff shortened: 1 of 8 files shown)


src/sst/core/serialization/impl/serialize_utility.h
```diff
@@ -17,24 +17,182 @@
...
+// Try to initialize N fields in an aggregate, with glob_ser() automatically converting to each field's
+// type. If conversion fails, it is not an error (SFINAE) and this specialization won't apply.
+template <class C, size_t... I>
+constexpr bool trivially_serializable_fields<C, std::index_sequence<I...>, decltype(C { (I, glob_ser())... })> = true;
+
+// has_no_ptr: Whether a type is an aggregate and contains no pointers, or is a non-pointer scalar
+
+// If it's not an aggregate, it needs to be integer, floating-point, enum or member pointer (not regular pointer)
+template <class C, bool = std::is_aggregate_v<C>>
+constexpr bool has_no_ptr = std::is_arithmetic_v<C> || std::is_enum_v<C> || std::is_member_pointer_v<C>;
```
feldergast, line 158: I've been trying to understand pointers to member, and it looks like pointer to member variable is just an offset and should be serializable, but a pointer to member function can change run to run and is not portable. std::is_member_pointer returns true for both of these.  Probably need to use std::is_member_object_pointer instead.

leekillough: We already exclude polymorphic classes generally, by requiring`std::is_standard_layout` and `std::is_aggregate`, so member function pointers in polymorphic classes are already excluded.

We would probably need to move the member function pointer test so that it is only applied to members of a class already verified to be non-polymorphic, rather than at the outermost level of the `is_trivially_serializable_v<>` test, such as moving it into the `glob_ser` class instead of the test above.

If the class is not polymorphic, how do member function pointers change from run to run? Does the member function pointer data format contain a field which can change from run to run even if the class is not polymorphic (no vtable, etc.)?

A member function pointer cannot itself depend on `this`, since it can be dereferenced through any instance of the class, e.g. `(this->*memfunc_ptr)()`.

leekillough: I added an `is_nonpolymorphic_member_function_pointer_v<T>` which returns `true` if a type is a cv-qualified pointer to member function of a nonpolymorphic class, and changed the test to `std::is_member_object_pointer_v<T> || is_nonpolymorphic_member_function_pointer_v<T>`.

But do you still think that member function pointers are non-portable across runs, even if the class is not polymorphic?

leekillough: I ran experiments with Clang and GCC on x86-64, and member function pointers pointing to virtual functions were actually simpler than member function pointers to non-virtual functions, because member function pointers to virtual functions simply have a vtable address offset based on which virtual function in the class is being pointed to (1st, 2nd, 3rd, etc.) ORed with `1`, while member function pointers to non-virtual functions have an even address to a static variable which changes from run to run.

So for non-virtual functions it points to a static data structure which is independent of the instance's `this` pointer, but which is not as simple as a fixed offset. In my tests, the address stored in the member function pointer was close to the address of a global variable I inserted in the code, so it's a data address instead of a code address or an offset. I think the member function pointer points to a hidden static data member that the compiler creates for the class which is independent of class instances. During invocation, the non-virtual function's implementation address is looked up in this static data structure and called, passing the `this` pointer and the rest of the arguments to it.

Since the member function pointer can pointer to either a virtual function or a non-virtual function independent of the member function pointer's type, the LSB bit of the address/offset is tested, and if it's `1`, the bit is cleared and the resulting offset is used to look up the vtable. If the LSB is `0`, it is a non-virtual function, and the static data structure is accessed through the address, and the function is called.

In the [Itanic ABI](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#member-function-pointers), which the x86-64 Unix C++ ABI is based on, it says that member function pointers are described by a pointer, and an adjustment used for handling inheritance and covariant functions.

So I have changed it to only consider member object pointers trivially serializable.

feldergast:
I've looked through the code at a high level several times, but haven't had time to delve deep or play around with the code, which I'll want to do before merging. Since it will still be a couple weeks before I have time to dig deeper into this optimization, I'll bring up some initial areas of potential concern now (though I'm not sure I should be concerned about these since I don't yet fully understand the code):

1 - The compile time tests should probably be moved to a .cc file, if possible.  I don't think we want to compile these for every file that includes serialization.

2 - Will moving the serialization for arithmetic types into the more complicated templating affect compile time?  We want to make sure we don't add too much compile time overhead to the base case of arithmetic operations. We already get some complaints about compile times and I don't want to make them worse (i.e. I'd prefer duplicated code over compile time overheads for the base case).

3 - Will this hide the default case check in serialize<T> that checks for non-polymorphic classes that have a serialize_order() function?  I shouldn't have to write a special serialize_impl class to get the serialize_order() function of these classes to be called.

4 - In general, except in the fast path (i.e. pointer tracking turned off), a class with a serialize_order() function should have that function called instead of the optimized trivially copyable path being used.

leekillough:
The code has been changed to consider a type as not trivially serializable if it contains a `serialize_order()` method. This is applied recursively, so if a `struct` contains a member which contains a `serialize_order()` method, the `struct` is not considered trivially serializable.

A test has been added which verifies that a `struct` containing a `serialize_order()` method is not trivially serializable, even if its members are otherwise trivially serializable.

leekillough:
This PR adds no significant time to compilation, and the trivial serialization templates, whether the ones for testing whether a type is trivially serializable, or the ones implementing `serialize_impl` for trivially serializable types, have a negligible effect of compilation time, as measured by `-ftime-trace`, since it does not show up in top profiles.

feldergast (approved):
Looks good

src/sst/core/serialization/impl/serialize_utility.h
```diff
@@ -17,24 +17,205 @@
...
+template <>
+inline constexpr bool is_trivially_serializable_v<long double _Complex, false> = true;
+#endif
+
+// Other floating-point types not covered by std::is_floating_point
+template <>
+inline constexpr bool is_trivially_serializable_v<_Float16, false> = true;
+
+template <>
+inline constexpr bool is_trivially_serializable_v<__float128, false> = true;
```
feldergast, line 208: Looks like _Float16 isn't available on the compilers on our Linux testers and _float128 isn't available on our Mac testers.  We'll either need to remove these are put checks around them.

PR #1333, opened 2025-05-12
core: remove __DATE__ / __TIME__ references from ELI

jmlapre:
They are rarely/never used and they prevent tools such as ccache from working correctly due to every compile returning different values.


feldergast (approved):

feldergast (approved):
Looks good.

PR #1337, opened 2025-05-15
Git rid of GCC false -Wmissing-noreturn warnings; remove unnecessary warning suppression code

leekillough:
This adds `DISABLE_WARN_MISSING_NORETURN`, to address GCC `-Wmissing-noreturn` false warnings introduced by recent PRs. It is only needed in `.cc` files which define virtual functions which do not return; it is not needed in `.h` files and there is no change needed for the virtual function declarations.

GCC (but not Clang) has a bug where `-Wmissing-noreturn` warnings are printed for virtual functions _defined outside of class_ which call `[[noreturn]]` functions but are not themselves marked `[[noreturn]]`. `[[noreturn]]` is a function attribute, not a type attribute, and is not inherited. A call to a virtual function with dynamic dispatch cannot be considered `[[noreturn]]` even if its declarations are marked `[[noreturn]]`, because it can be overriden with a function which is not `[[noreturn]]`, and a virtual function should not be warned about not being marked `[[noreturn]]` even if it calls a `[[noreturn]]` function, because `[[noreturn]]` adds no value to a virtual function and only serves to remove a false warning. There is discussion [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117337), [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61379) and [here](https://github.com/llvm/llvm-project/issues/12572).

This also removes some warning suppression methods which are now unnecessary:
1. `UNUSED()` macro can suppress warnings about unused variables and function parameters
2. It is not necessary to return dummy values from functions after `Output::fatal()` is called, since it is now marked `[[noreturn]]`


feldergast (approved):

leekillough:
GCC bug submitted:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120300

PR #1339, opened 2025-05-20
Generate JSON timing information at end of simulation

kpgriesser:
Added a new command line option `--timing-info-json=FILE` to write timing information into a JSON formatted file. Moved the code to print the timing information to `stdout` from main.cc and into a new file where that data is captured as key-value pairs and then can be printed as text and/or as JSON depending on the command line options. 

Added test case to verify well formed json is generated with the expected keys.

### DIFF
--- a/src/sst/core/config.cc
+++ b/src/sst/core/config.cc
@@ -415,7 +415,7 @@ Config::insertOptions()
         "Provide options to the python configuration script.  Additionally, any arguments provided after a final '-- ' "
         "will be appended to the model options (or used as the model options if --model-options was not specified).",
         model_options_, false, false, true);
-    DEF_FLAG_OPTVAL("print-timing-info", 0, "Print SST timing information", print_timing_, true);
+    DEF_FLAG_OPTVAL("print-timing-info", 0, "Print SST timing information", print_timing_, true, true, false);
     DEF_ARG(
         "timing-info-json", 0, "FILE", "Write SST timing information in JSON format", timing_json_, true, true, false);
     DEF_ARG("stop-at", 0, "TIME", "Set time at which simulation will end execution", stop_at_, true, true, false);
--- a/src/sst/core/config.cc
+++ b/src/sst/core/config.cc
@@ -320,7 +320,7 @@ Config::merge_checkpoint_options(Config& other)
         if ( option.annotations.size() <= r_index || !option.annotations[r_index] ) continue;
 
         // Now see if we need to copy over the option.  We copy it if
-        // it wasn't set on the command line, or it isne't allowed to
+        // it wasn't set on the command line, or it isn't allowed to
         // be set on the command line for a restart.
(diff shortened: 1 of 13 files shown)


kpgriesser:
I've changed the code to use the new options macros, moved the header to the sst_core_sources, and retested  combinations of print-timing-info and json-timing-info.  Ready for review.

src/sst/core/config.h
```diff
@@ -183,6 +183,11 @@ class Config : public ConfigShared, public SST::Core::Serialization::serializabl
     */
     SST_CONFIG_DECLARE_OPTION(bool, print_timing, false, &StandardConfigParsers::flag_default_true);
 
+    /**
+        Print SST timing information to JSON file
+    */
+    SST_CONFIG_DECLARE_OPTION(std::string, timing_json, "NONE", &StandardConfigParsers::from_string<std::string>);
```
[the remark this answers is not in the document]
feldergast: Okay, I'll mark as WIP until you push the changes.

src/sst/core/config.cc
```diff
@@ -416,6 +416,8 @@ Config::insertOptions()
         "will be appended to the model options (or used as the model options if --model-options was not specified).",
         model_options_, false, false, true);
     DEF_FLAG_OPTVAL("print-timing-info", 0, "Print SST timing information", print_timing_, true);
+    DEF_ARG(
+        "timing-info-json", 0, "FILE", "Write SST timing information in JSON format", timing_json_, true, true, false);
```
feldergast, line 420: We had this option as not carrying over to a restart (second annotation).  I'm okay if we change that, but the --print-timing-info and --timing-info-json should match for whichever way we choose.

src/sst/core/main.cc
```diff
@@ -1265,62 +1266,39 @@ main(int argc, char* argv[])
...
-        g_output.output("  Current global TimeVortex depth: %" PRIu64 " entries\n", global_current_tv_depth);
-        g_output.output("  Max TimeVortex depth:            %" PRIu64 " entries\n", global_max_tv_depth);
-        g_output.output(
-            "  Max Sync data size:              %s\n", global_max_sync_data_size_ua.toStringBestSI().c_str());
-        g_output.output("  Global Sync data size:           %s\n", global_sync_data_size_ua.toStringBestSI().c_str());
-        g_output.output("------------------------------------------------------------\n");
-        g_output.output("\n");
-        g_output.output("\n");
+    if ( myRank.rank == 0 && (cfg.verbose() || cfg.print_timing() || cfg.timing_json() != "") ) {
+        std::unique_ptr<TimingOutput> timingOutput =
```
feldergast, line 1278: Is there a reason to use std::unique_ptr here instead of just creating an instance of the object in this scope?  Are you just wanting to avoid allocating the TimingOutput object on the stack?

kpgriesser:
requested changes pushed

PR #1344, opened 2025-05-25
serialize std::variant

leekillough:
This adds serialization of `std::variant`.


leekillough:
```c++

template <typename... Types>
template <size_t... INDEX>
constexpr void (*serialize_impl<std::variant<Types...>>::set_index<std::index_sequence<INDEX...>>[])(
    std::variant<Types...>&) { [](std::variant<Types...>& obj) { obj.template emplace<INDEX>(); }... };
```
###  In case the code is too hard to understand, here's a dissection:
```c++
template <typename... Types>
```
The outer template parameter list.
```c++	
template <size_t... INDEX>
```
The inner template parameter list.
```c++
constexpr
```
The `constexpr` applies to the `set_index` variable template being defined here, which also means that, as a `constexpr static` member variable of `serialize_impl<std::variant<Types...>>`, `set_index` is `inline`, and so it can be defined in a header file without causing ODR violations (multiply-defined symbols).
```c++
 void (*                                                                                [])( std::variant<Types...>&)
        serialize_impl<std::variant<Types...>>::set_index<std::index_sequence<INDEX...>> 
```
A partial specialization of the `set_index` member variable template of `serialize_impl<std::variant<Types...>` which matches `std::index_sequence<INDEX...>`, which is a list of `0, 1, 2, ..., sizeof...(Types)-1`. 

This specialization is declared as an unspecified-size array of pointers to functions taking `std::variant<Types...>&` and returning `void`.
```c++
 { 
   [](std::variant<Types...>& obj) { obj.template emplace<INDEX>(); }
   ... 
 };
```
An initialization list for the unspecified-size array of function pointers, which gets expanded for each element of the `INDEX` list of subscripts.

Each element of the list is a lambda expression with no captures which takes a `std::variant<Types...>&` and (implicitly `auto`) returns `void`. Because there are no captures, the lambda expression is implicitly convertible to the function pointer type of the array elements.

The lambda function calls the `emplace()` method on the `obj` argument with the template argument `INDEX`, which is the index corresponding to the current function pointer array element in the initialization list.

Passing the `emplace()` method a template argument of `INDEX` causes the `Types...` type corresponding to `INDEX` to become the variant's new type.

The `emplace()` argument list is empty, so it default-initializes the new variant type.

feldergast (changes requested):
Thanks for the comments. I mostly understand what's going on and am willing to take the rest on faith. Can you add those comments to the appropriate places in the code so that we have it in the future when someone needs to understand the code.

PR #1353, opened 2025-06-07
Clean up serializer class

leekillough:
This PR cleans up the `serializer` class:

- The current mode is controlled by a `std::variant<pvt::ser_sizer, pvt::ser_packer, pvt::ser_unpacker, pvt::ser_mapper> ser_` member, each variant of which holds all data necessary for the current mode.

- The variants have indices of `SIZER`, `PACK`, `UNPACK`, and `MAP`, and the current mode is the one returned by `ser_.index()`.

- The serializer data for a mode is accessed like `std::get<UNPACK>(ser_)`, where `ser_` is the `std::variant` member.

- Since their names are shorter, and they already existed, the accessors `sizer()`, `packer()`, `unpacker()` and `mapper()` are used everywhere, and they call `std:get<MODE>(ser_)`. What was previously `sizer_` is now `sizer()`.

- If a method is called which is not valid for the current mode, a `std::bad_variant_access` exception is thrown. This
error checking is automatically provided by `std::get<MODE>(std::variant)`.

- All pointer tracking data structures, such as `ser_pointer_set` and `ser_pointer_map`, have been moved to `pvt::sizer`, `pvt::ser_packer`, `pvt::ser_unpacker`, and `pvt:ser_mapper`, so that they are private to each mode, and are automatically initialized/destroyed when the mode changes, without requiring explicit initialization or destruction/clearing in the `serializer` class. This was motivated by `std::shared_ptr` tracking requiring additional tracking containers, and requiring different ones for different serialization modes.

- ~~`set_mode()` is marked deprecated, because it was not very useful, was not used anywhere in Core or Elements, and in the new code, we cannot change the mode without changing and initializing the new variant. `set_mode()` will still change the mode, but the variant of the new mode will be default-initialized, which means `pvt::ser_packer` will have a `max_size` of 0, etc. There was no way in the old interface to change `max_size` except `start_packing()`, and if you changed modes, some of the settings could be uninitialized or have bad data from the old mode. Now switching modes is clean, and data for each mode is clearly separated.~~

- ~~`reset()` is not used anywhere either, and is marked deprecated.~~

- ~~If it is not important that we keep `set_mode()` and `reset()`, I vote to remove them immediately.~~

- `start_sizing()`, `start_packing()`, `start_unpacking()`, and `start_mapping()` change the active mode, and
initialize the mode's data structures by calling the variant's constructor.

- The `.init()` methods of the `pvt::` serializers have been replaced by constructors. For example,

```c++
    void start_unpacking(char* buffer, size_t size) { ser_.emplace<UNPACK>(buffer, size); }
```

In this example, `start_unpacking()` changes the mode to `UNPACK`, automatically calling the destructors of any data in the previous mode, and calls `pvt::ser_unpacker`'s constructor with `buffer` and `size` arguments, instead of
separately calling an `.init()` method.

- `pvt::ser_packer` and `pvt::ser_unpacker`, which are derived from `ser_buffer_accessor`, use C++11 inherited constructors so that the `pvt::ser_packer(buffer, size)` and `pvt::ser_unpacker(buffer, size)` constructors are automatically created, and pass their arguments to the inherited `ser_buffer_accessor` constructor.

- Non-templated serialization functions which aren't very short were moved from `serializer.h` to `serializer.cc`.

- To avoid having to overload `sizer()`, `packer()`, `unpacker()` and `mapper()` with `const` versions in addition to their non-const versions, `const` has been removed from the `size()` method, which was the only method which used `const` on `*this`. Since almost all uses of `size()` will be `ser.size()` where `ser` is a non-`const` reference to a `serializer` (all cases in Core and Elements so far), this will not cause a problem. Only if you had `const serializer& ser` and needed to call `ser.size()` would a `const` method of `size()` be needed. If this is not acceptable, we can either duplicate the accessors with `const` versions, or we can change `size()` to not use the accessors but to use `std::get<>(ser_)` directly. I like the accessors as being a single place where the variants are extracted, and they replace names like `sizer_` with `sizer()`, `packer_` with `packer()`, etc.

- `check_pointer_pack()` was split into `check_pointer_pack()` and `check_pointer_sizer()`, and the `std::set` was moved from `serializer` into `pvt::ser_sizer` and `pvt::ser_packer`, so that `SIZE` and `PACK` modes have their own pointer tracking sets, and the right `check_pointer_*` for the current mode is called.

- The maps used in mapping mode and unpacking mode were moved to `pvt::ser_mapper` and `pvt::ser_unpacker`, and although the map has the same declaration in both, the way the map is used is totally different in unpacking and mapping mode. Best to separate them.

- The new code is 40+ lines shorter and much cleaner. Adding serialization of `std::shared_ptr` and `std::unique_ptr` will be simpler since now the tracking data is separately in each mode's `pvt::` class, and is cleanly initialized/destroyed when switching modes.

Edit 6/10:

- `EMPTY` has been added as the default serializer mode. 
- A `std::monostate` type has been added as the default `EMPTY` type of the `std::variant`.
- Any attempt to access `sizer()`, `packer()`, `unpacker()` or `mapper()` during a different mode than `SIZER`, `PACK` ,`UNPACK` or `MAP`, respectively, causes a runtime error.
- The `set_mode()` and `reset()` methods have been removed.
- A `finalize()` method has been added which, right now, simply sets the mode to `EMPTY` and destroys any state object of the old mode. Later it might be changed to do more, such as considering the current mode and doing something special for it, such as finalization of mapping mode.
- In past discussions, a comment has been made that switching to `SIZER` mode is sometimes used to clear the old state; however, `SIZER` mode has its own state (a `std::set` which is empty on construction). An explicit `EMPTY` mode separate from other modes is preferable.
- To prevent warnings about `EMPTY` not being listed in `switch` cases, `EMPTY` has been defined outside of the `enum`.
- All `pvt::ser_packer` functions defined in `serializer.cc` have been moved to a new file `impl/packer.cc`.
- All `pvt::ser_unpacker` functions defined in `serializer.cc` have been moved to a new file `impl/unpacker.cc`
- The `ser_buf_accessor` has been made much simpler, with a single method `buf_next(size)` which returns the buffer address and advances it by `size`, making sure that it does not overflow. This simplifies the multiple methods which were used before, such as special methods for `std::string`.


leekillough:
`UNSTARTED` is the name of the default mode of the serializer, and `finalize()` changes the mode back to `UNSTARTED`, destroying any tracking object which holds state for the current mode.

For `std::shared_ptr`, `finalize()` will need to be called after all unpacking is done, so that all of the copies of `std::shared_ptr` owners stored in `pvt::ser_unpacker` are destroyed at the end of deserialization, and the reference counts of the `std::shared_ptr` will be the same as they were when they were serialized. Otherwise the reference counts will be `+1` because of an extra copy in `pvt::ser_unpacker`. The copy is needed throughout the entire deserialization sequence so that `std::shared_ptr` and `std::weak_ptr` can be deserialized with the same shared ownership relationships that they had during serialization.

PR #1359, opened 2025-06-16
Statistic parameter documentation

gvoskuilen:
Enable inheriting params in stat ELI
Add param documentation to stat types and outputs


feldergast (approved):
There are a few typos in string that I'll fix in a PR I'm currently working on. Otherwise, looks good.

PR #1362, opened 2025-06-18
Checkpoint config object

feldergast:
Checkpoint Config object and merge with new Config on restart.  Merge some regular and restart paths based on having the merged Config early on in the execution.

### DIFF
--- a/src/sst/core/params.cc
+++ b/src/sst/core/params.cc
@@ -279,25 +279,67 @@ Params::getParamName(uint32_t id)
 void
 Params::serialize_order(SST::Core::Serialization::serializer& ser)
 {
-    SST_SER(my_data);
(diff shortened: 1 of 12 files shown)


src/sst/core/params.cc
```diff
@@ -279,25 +279,67 @@ Params::getParamName(uint32_t id)
...
+            break;
+        case SST::Core::Serialization::serializer::MAP:
+            // This function not called in mapping mode
+            break;
+        }
+    }
+    else {
+        // Used for checkpointing, where there shouldn't be many
+        // params objects being used.  We do it this way because the
+        // keyMap and associated data structures are guaranteed to be
```
gvoskuilen, line 313: not guaranteed

PR #1365, opened 2025-06-25
Serialize aggregate class types

leekillough:
This is a relatively minor PR in terms of its impact.

It allows for the automatic serialization of aggregate classes which are not trivially serializable and which do not have a `serialize_order()` method.

Structured bindings are used to extract the all-public members of an aggregate class, and the members are serialized individually.

This is not likely to fill many gaps in serialization, since it does not apply to types which are trivially serializable; it does not apply if a class has private or protected data members; it does not apply to classes with virtual methods or virtual bases; it does not apply if a class has a `serialize_order()` method.


leekillough:
The biggest gain that this provides is automatic serialization of C-style `struct`s with pointer members.

feldergast (approved):

Issue #1350, opened 2025-06-02
Reduce compilation time by reducing the time spent instantiating ELI templates


leekillough:
Attached is the profile generated by Clang `-ftime-trace` and analyzed by [ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer).

```
**** Templates that took longest to instantiate:
 11986 ms: SST::ELI::InstantiateBuilderInfo<SST::Profile::ComponentCodeSegmentProfileTool, SST::Profile::ComponentCodeSegmentProfileToolCount>::isLoaded (45 times, a
vg 266 ms)
 11981 ms: SST::ELI::ElementsInfo<SST::Profile::ComponentCodeSegmentProfileTool>::add<SST::Profile::ComponentCodeSegmentProfileToolCount> (45 times, avg 266 ms)
 11976 ms: SST::Profile::ComponentCodeSegmentProfileTool::addDerivedInfo<SST::Profile::ComponentCodeSegmentProfileToolCount> (45 times, avg 266 ms)
  7785 ms: SST::ELI::InstantiateBuilderInfo<SST::Statistics::Statistic<void>, SST::Statistics::NullStatistic<void>>::isLoaded (79 times, avg 98 ms)
  7776 ms: SST::ELI::ElementsInfo<SST::Statistics::Statistic<void>>::add<SST::Statistics::NullStatistic<void>> (79 times, avg 98 ms)
  7766 ms: SST::Statistics::Statistic<void>::addDerivedInfo<SST::Statistics::NullStatistic<void>> (79 times, avg 98 ms)
  7757 ms: SST::ELI::BuilderInfoImpl<SST::ELI::ProvidesInterface, SST::ELI::ProvidesParams, SST::ELI::ProvidesDefaultInfo, void>::BuilderInfoImpl<SST::Statistics::Nu
llStatistic<void> *> (79 times, avg 98 ms)
  7687 ms: SST::ELI::BuilderInfoImpl<SST::ELI::ProvidesDefaultInfo, SST::ELI::ProvidesInterface, SST::ELI::ProvidesParams, void>::BuilderInfoImpl<SST::Profile::Compo
nentCodeSegmentProfileToolCount *> (45 times, avg 170 ms)
```

Most of time is spent instantiating templates in ELI.

[...]

PR #1366, opened 2025-06-28
Relax serialize_impl rules to allow polymorphism, make has_serialize_order trait global to the serializer namespace

leekillough:
Do not forbid polymorphic classes from using the default `serialize_impl` implementation

- Some mapping mode tests tried to serialize polymorphic classes
- Mapping mode is more likely to be a use-case where polymorphic classes will need to be supported
- Since a `serialize_order()` method must be defined for a class if `serialize_impl` is not specialized for it, that `serialize_order()` method should know whether it needs to handle derived classes and how to handle them
- The `serialize_order()` method can itself be virtual to handle different derived types, or it can use `typeid` and/or `dynamic_cast`
- A class can be marked `final` if deriving from it should be prohibited, rather than prohibiting all polymorphic classes which define `serialize_order()` from the default `serialize_impl`

Make the `has_serialize_order` and `has_serialize_order_v` traits global to the `SST::Core::Serialization` namespace so that there is one consistent definition instead of multiple definitions (one in `serialize_impl` and one in `pvt_trivial`).

Make `LinkMap`'s `serialize_order()` method public so that it can be detected by `has_serialize_order` defined outside of `serialize_impl`, instead of making `serialize_order()` private to `LinkMap` and `serialize_impl` a friend class of `LinkMap`.

Include `serialize_utility.h` at the top of `serialize.h` so that `has_serialize_order` and `has_serialize_order_v` can be used in `serialize.h`, and so that all other files in `impl/` do not need to `#include` `serialize_utility.h`. There is no slowdown in compile-time from just `#include`'ing the template definitions; the slowdown occurs during instantiation, and right now ELI, not the serializer, is where the [bulk of template instantiation compile time](https://github.com/sstsimulator/sst-core/issues/1350) is spent.

Greatly simplify the pointer serialization when pointer tracking is disabled.

Change `sst_ser_object()` implementation to define type `T` locally as the universal reference `TREF&&` with references removed, for code brevity and readability.

Improve the `serialize_impl` `static_assert` wording about missing `serialize_order()` to be more accurate and precise.

Re-wrap some comments and deprecation warnings so that they take up fewer lines but stay within the 120 width limit.


feldergast:
Generally, this looks good, but I don't think removing the requirement for the class to be non-polymorphic is safe.  Generally classes are polymorphic because they will be referred to by their base classes, and while the class writer could likely make the SIZE and PACK modes work correctly, UNPACK won't work if the class was serialized from a pointer to the base class.  The reason is that the serializer needs to know what the actual class is in order to create it before calling serialize_order() on it. If you serialized it by a pointer to the base class, then it will just create the base class and call its serialize_order() function and all the data from the derived class will be left out.  The other effect of that case is that the serialization stream will now contain extra data and the next thing to serialize will get the wrong data.

For polymorphic classes, the objects need to inherit from serializable and call the ImplementSerializable macro.  You can add this to the class hierarchy at any time through multiple inheritance.  The serializable class infrastructure handles the tracking of the actual class being serialized so that it can construct the correct class on UNPACK.

While the user could potentially make all this work by casting to the final class before serialization, I'd rather not allow it to avoid issue with the more common use case, which won't work.

leekillough:
> Generally, this looks good, but I don't think removing the requirement for the class to be non-polymorphic is safe. Generally classes are polymorphic because they will be referred to by their base classes, and while the class writer could likely make the SIZE and PACK modes work correctly, UNPACK won't work if the class was serialized from a pointer to the base class. The reason is that the serializer needs to know what the actual class is in order to create it before calling serialize_order() on it. If you serialized it by a pointer to the base class, then it will just create the base class and call its serialize_order() function and all the data from the derived class will be left out. The other effect of that case is that the serialization stream will now contain extra data and the next thing to serialize will get the wrong data.

That's a problem with inheritance generally, not just polymorphic classes. Maybe using base class pointers to derived classes is more likely if the class is polymorphic, but the problem remains with serializing non-polymorphic base classes which are derived from. There's no way to test whether a non-`final` class has been derived from.

Maybe requiring `std::is_final` is a better test than `!std::is_polymorphic`, but then you could say that it's too strict and breaks compatibility for non-polymorphic classes which aren't marked `final`. A compromise could be `std::is_final || !std::is_polymorphic`.

In `std::shared_ptr` there is a similar problem with _aliasing_ `std::shared_ptr` -- a `std::shared_ptr<T>` can share _ownership_ with a `std::shared_ptr<U>` even when `T` and `U` are independent and one is not necessarily a base of the other.

The solution there is to use a wrapper which serializes a "parent" `std::shared_ptr` in addition to the `std::shared_ptr` being serialized, such as `SST_SER( SST::Core::Serialization::shared_ptr( ptr, parent_ptr ) )`, passing both `ptr` and its owning `parent_ptr`. When `ptr` is deserialized, it deserializes the `parent_ptr`, creating a parent object by allocating memory and deserializing it. The `ptr` is then set to alias the parent but share ownership with it. Fortunately, during sizing/packing serialization, you can verify that `ptr` and `parent_ptr` share the same ownership, and flag it with an error if they do not.

> For polymorphic classes, the objects need to inherit from serializable and call the ImplementSerializable macro. You can add this to the class hierarchy at any time through multiple inheritance. The serializable class infrastructure handles the tracking of the actual class being serialized so that it can construct the correct class on UNPACK.
> 
> While the user could potentially make all this work by casting to the final class before serialization, I'd rather not allow it to avoid issue with the more common use case, which won't work.

This problem was encountered during mapping mode when trying to write a `serialize_order()` for a polymorphic class. If we make an exception to allow polymorphic classes for mapping mode, then the test for polymorphism becomes a runtime error instead of a compile-time error. But then we are splitting apart the uses of serialization, using `serialize_order()` for mapping mode in a way which won't work for other modes.

I can add the test for polymorphism back, but I think it should not apply if the class is `final`. The question remains how to handle mapping mode for polymorphic classes.

feldergast (dismissed):

feldergast:
This is failing on GCC 8.5.  The has_serialize_order attribute doesn't like it when the function is protected.

```
../../../../sst-core/src/sst/core/serialization/impl/serialize_utility.h:64:48: error: ‘virtual void SST::Shared::SharedObjectData::serialize_order(SST::Core::Serialization::serializer&)’ is protected within this context
 using has_serialize_order = std::bool_constant<has_serialize_order_v<T>>;
                                                ^~~~~~~~~~~~~~~~~~~~~~~
In file included from ../../../../sst-core/src/sst/core/main.cc:41:
../../../../sst-core/src/sst/core/shared/sharedObject.h:214:10: note: declared protected here
     void serialize_order(SST::Core::Serialization::serializer& ser) override { SST_SER(name); }
          ^~~~~~~~~~~~~~~
 In file included from ../../../../sst-core/src/sst/core/serialization/serialize.h:21,
                 from ../../../../sst-core/src/sst/core/serialization/serializable.h:16,
                 from ../../../../sst-core/src/sst/core/mempool.h:15,
                 from ../../../../sst-core/src/sst/core/activity.h:15,
                 from ../../../../sst-core/src/sst/core/action.h:15,
                 from ../../../../sst-core/src/sst/core/clock.h:15,
                 from ../../../../sst-core/src/sst/core/baseComponent.h:15,
                 from ../../../../sst-core/src/sst/core/baseComponent.cc:14:
```

This passes on our Mac test platform.

Not sure if there's an easy way to fix it, but I'd rather not have to make serialize_order always public, when it doesn't need to be.  has_serialize_order should just be false if the function isn't public.

leekillough:
I will have to investigate this further.

On my GCC compiler, `has_serialize_order_v()` returns `false` if `serialize_order()` is not public, and SFINAE makes it not an error if it is not public. I get no compilation error with my version of GCC.

I seem to recall that `decltype()` used to ignore access specifiers in early versions of C++11, but later it was required to consider them. In SFINAE contexts, an access error should not cause a compilation error, but simply make the function overload or template specialization not apply.

I don't like the general pattern of making class member functions `private` and then granting friendship to only certain classes/functions. There are other techniques such as ADL (argument dependent lookup) which can allow access to private or hidden internals. I don't have the time to elaborate on it now.

The `has_serialize_order` trait defined at namespace scope cannot access private or protected `serialize_order()` methods of classes. It will need to be placed inside a class if it is going to be granted friendship by other classes. The global version will need to use that class's version if we don't want duplicated definitions.

If the old code did not have a compilation error but the new code does (I will need to investigate the cause), it could be because the old code accessed a protected `serizalize_order()` method through friendship granted to `serialize_impl`, something which cannot be granted to namespace-scope variable template definitions.

feldergast:
The ImplementSerializable() macro does add friendship for the serialize and/or serialize_impl templates.  I think that was originally modeled off of how the boost serialization macros worked.

feldergast:
> 
> On my GCC compiler, `has_serialize_order_v()` returns `false` if `serialize_order()` is not public, and SFINAE makes it not an error if it is not public. I get no compilation error with my version of GCC.
> 

This is how I would expect things to work.  This is likely a "bug" in older versions of GCC.

leekillough:
See:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116639

I have pushed a fix which tests for `serialize_order()` member function not by testing `decltype(std::declval<T>().serialize_order(std::declval<serializer&>()))` but by testing `decltype(&T::serialize_order)` .

feldergast:
Still seeing issues with older versions of GCC when serialize_order() is not public:

```
../../../src/sst/core/serialization/impl/serialize_utility.h:60:62: error: ‘void SST::SSTHandlerBase<returnT, argT>::serialize_order(SST::Core::Serialization::serializer&) [with returnT = bool; argT = long unsigned int]’ is protected within this context
 constexpr bool has_serialize_order_v<T, std::void_t<decltype(&T::serialize_order)>> =
                                                              ^~
In file included from ../../../src/sst/core/clock.h:16,
                 from ../../../src/sst/core/baseComponent.h:15,
                 from baseComponent.cc:14:
../../../src/sst/core/ssthandler.h:201:10: note: declared protected here
     void serialize_order(SST::Core::Serialization::serializer& ser) override
          ^~~~~~~~~~~~~~~
```

leekillough:
I tried to delegate it to `serialize_impl`, putting `has_serialize_order_v` inside of `serialize_impl` to allow friendship access, but I found that this led to weird template recursion -- in order to test whether `serialize_order()` can be called from inside `serialize_impl`, you first have to instantiate `serialize_impl`, which leads to `has_serialize_order_v` being tested to choose which `serialize_impl` to instantiate, etc., leading to infinite template recursion and other errors.

`has_serialize_order_v<T>` -> `serialize_impl<T>` -> `has_serialize_order_v<T>` -> `serialize_impl<T>` etc.

Bug reports/discussions (it has been a perpetual problem with GCC):

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116639
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96204
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90880
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61806
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=59002
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=51213
https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1170
https://stackoverflow.com/questions/39758335/private-member-access-in-template-substitution-and-sfinae

I've been trying to think of workarounds, but cannot think of any for the general case. Before C++11, access control was ignored in SFINAE, which means that a member could be private/protected and the template parameter substitution could still succeed, but the code would probably fail later. In C++11 and later, access control was made a part of SFINAE, and bad access was considered a substitution failure (but not an error), so `has_serialize_order_v<T>` will return `false` if `serialize_order()` is private/protected in `T`.

A possible workaround is to duplicate `has_serialize_order_v` inside `serialize_impl` to get rid of this particular error, i.e. preserving compatibility with legacy SST code, but the namespace-scope `has_serialize_order_v` will still get an access error if it is tried with the examples, such as when `is_trivially_serializable_v<T>` tests if a type contains `serialize_order()` to exclude it from trivial serialization. That particular test is short-circuited out with other tests, which is probably why it avoids getting the access error before this PR. I'll add `#if` code to enable this workaround for older versions of GCC which fail this test:
```c++
#include <utility>

class c { void serialize_order() {} };

template<class C, class = void>
constexpr bool has_serialize_order_v = false;

template<class C>
constexpr bool has_serialize_order_v<C, decltype(std::declval<C>().serialize_order())> = true;

static_assert( !has_serialize_order_v<c>, "access");
```
I'll push the changes soon.

feldergast (dismissed):

leekillough:
Does it still get the error? I wonder if it can be mitigated by making `serialize_order()` method public. (I don't like the "pattern" of using all-private classes and only granting friendship between classes.)

I would need to know the version of GCC, the source file and line which instantiates the template which causes the error, the template instantiation chain which leads to the access error (`In X`, `instantiated by Y`, `required by Z`, etc.). A full error log would be helpful.

The latest version of this PR keeps `has_serialize_order_v` defined inside of `serialize_impl` as well as outside of it for older versions of GCC, so it should not cause access errors in code which relied on friendship granted to `serialize_impl` when it tests `has_serialize_order_v` inside of `serialize_impl`. But if it tests it outside of friendship-granted `serialize_impl` by using the namespace-scope definition of `has_serialize_order_v`/`has_serialize_order`, and `serialize_order()` is not public, then it will either correctly evaluate `false` or it will get the GCC access error bug.

The shadow definition of `has_serialize_order_v` inside of `serialize_impl` is enabled by a complicated `#if` expression which might not being enabled in your build. Through experimentation on [Godbolt](https://godbolt.org/z/vdvKq8d9v), I determined that GCC 12.1 fixed the access error during template substitution with the testcase shown above, but this may not adequate:
```c++
#if defined(__GNUC__) && __GNUC__ < 12 && !defined(__llvm__) && !defined(__INTEL_COMPILER)
```
(The LLVM and Intel compiler tests are because those compilers also define `__GNUC__`.)

I would need the compiler and its version, and the result of `g++ -E` when fed this source:
```c++
__GNUC__
__GNUC_MINOR__
__GNUC_PATCHLEVEL__
__llvm__
__INTEL_COMPILER
```
(If any of the macros are undefined, `g++ -E` will not replace them, and so they will appear unchanged in the output.)

leekillough:
I have added a workaround:

If a class type `T` is derived from `serializable_base`, it is assumed to have a `serialize_order()` method even if it it private or protected in `T`.

`serialize_order()` is a `public` virtual function in `serializable_base`, but many intermediate derived classes mark it as `private` or `protected` to prevent serialization of these intermediate classes through `serialize_impl`. In final derived classes, it is `public` again.

This workaround is conditional based on `#if` to detect buggy GCC compilers prior to version 12.

If this workaround is disabled because the compiler does not have the SFINAE bug, `has_serialize_order_v<T>` returns `false` if the `serialize_order()` method is private or protected in `T`.

With this workaround, if `T` is derived from `serializable_base`, `has_serialize_order_v<T>` returns `true` for the purposes of exclusion in `is_trivially_serializable` and the `static_assert` message in the `serialize_impl` default definition, but when the default `serialize_impl` tries to call it, it will still get a compile-time access error saying that it is `private` or `protected`. 

So the only loss is a less thorough diagnostic error message when the compiler has the SFINAE bug with access specifiers.

feldergast (approved):

Issue #1061, opened 2024-03-20
Show diff for failing tests


berquist:
Maybe first ~20 lines of diff if it's long.

PR #1377, opened 2025-07-08
Print diff on failure for tests that generate them

berquist:
Part of https://github.com/sstsimulator/sst-core/issues/1061.

I think a better implementation, and the one I started with, is to have the core functions return the diff result lines rather than a success flag.  The problem with doing this is that the current functions return true if there is no diff (success), but truthiness of a Python collection is true if it holds things (there is a diff) and false if empty (there is no diff).  This means that every single `testing_compare*` usage needs to have its logic flipped and is a breaking change for users.

Instead I saw in `testsuite_default_UnitAlgebra.py` that the diff lines are read from a file and printed before the respective assertion if it is going to fail, so I copied that idea.

Attached is the original implementation patch for posterity.

---

```patch
diff --git src/sst/core/testingframework/sst_unittest_support.py src/sst/core/testingframework/sst_unittest_support.py
index 4bc0d584..332e5398 100644
--- src/sst/core/testingframework/sst_unittest_support.py
+++ src/sst/core/testingframework/sst_unittest_support.py
@@ -1426,7 +1426,7 @@ def testing_compare_filtered_diff(
     sort: bool = False,
     filters: Union[LineFilter, List[LineFilter]] = list(),
     do_statistics_comparison: bool = False,
-) -> bool:
+) -> List[str]:
     """Filter, optionally sort and then compare 2 files for a difference.

         Args:
@@ -1439,8 +1439,10 @@ def testing_compare_filtered_diff(
               in order, but will break early if any filter returns None

         Returns:
-            (bool) True if the 2 sorted files match
-
+            (list[str]) The contents of a possible diff.  If the list is empty,
+              the files were identical after applying the optional sort and filters.
+              Evaluating the output as a boolean will give True if there were any
+              differences and False if there were none.
     """

     check_param_type("test_name", test_name, str)
@@ -1452,32 +1454,34 @@ def testing_compare_filtered_diff(

     if not os.path.isfile(outfile):
         log_error("Cannot diff files: Out File {0} does not exist".format(outfile))
-        return False
+        return []

     if not os.path.isfile(reffile):
         log_error("Cannot diff files: Ref File {0} does not exist".format(reffile))
-        return False
+        return []

     # Read in the output and reference files, and optionally sorting them
     out_lines = _read_and_filter(outfile, filters, sort, False)
     ref_lines = _read_and_filter(reffile, filters, sort, True)

-    # Get the diff between the files
-    diff = difflib.unified_diff(out_lines,ref_lines,outfile,reffile,n=1)
+    # Get the diff between the files.  We exhaust the diff generator
+    # immediately in order to always work with the list of diff lines.
+    # Because of truthiness
+    # (https://docs.python.org/3/library/stdtypes.html#truth-value-testing),
+    # an empty list (no differences) will evaluate as false and one with
+    # elements (differences between files) will evaluate as true.
+    diff = list(difflib.unified_diff(out_lines, ref_lines, outfile, reffile, n=1))

-    # Write the diff to a file and count the number of lines of output
-    # to determine if they matched or not (the output of the diff is a
-    # generator, so the only way to see if it's empty is to start to
-    # write the data).
+    # Write the diff to a file; it can later be recalled using the
+    # `testing_get_diff_data` function.
     diff_file = "{1}/{0}_diff_file".format(test_name, test_output_get_tmp_dir())
-    count = 0
     with open(diff_file, "w") as fp:
-        for line in diff:
-            fp.write(line)
-            count += 1
+        fp.write("\n".join(diff))
+
+    # Files are the same if there were no lines of diff output (the list is
+    # empty).
+    return diff

-    # Files are the same if there were no lines of diff output
-    return count == 0

 ###

```

### DIFF
--- a/tests/testsuite_default_Checkpoint.py
+++ b/tests/testsuite_default_Checkpoint.py
@@ -170,6 +170,9 @@ class testcase_Checkpoint(SSTTestCase):
             StartsWithFilter("WARNING: No components are assigned") ]
 
         cmp_result = testing_compare_filtered_diff("NeedSomethingHere", outfile_cpt, reffile, True, filters_cpt)
+        if not cmp_result:
+            diffdata = testing_get_diff_data(testtype)
+            log_failure(diffdata)
         self.assertTrue(cmp_result, "Output from original checkpoint run {0} did not match reference file {1}".format(outfile_cpt, reffile))
 
         prefix_rst = teststr + "_rst"
@@ -198,6 +201,9 @@ class testcase_Checkpoint(SSTTestCase):
             StartsWithFilter("WARNING: No components are assigned") ]
 
         cmp_result = testing_compare_filtered_diff("PortModule", outfile_rst, reffile, True, filters_rst)
+        if not cmp_result:
+            diffdata = testing_get_diff_data(testtype)
+            log_failure(diffdata)
         self.assertTrue(cmp_result, "Output/Compare file {0} does not match Reference File {1}".format(outfile_rst, reffile))
 
         ## Optionally run from restart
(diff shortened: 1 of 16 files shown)


berquist:
`master` branch...

Issue #1376, opened 2025-07-08
Increase minimum required Python version from 3.6 to 3.9


berquist:
As part of the 15.0.0 internal release plan, we decided to perform a minimum required Python version bump.  However, in the public announcement at http://sst-simulator.org/sstannouncements/2025/05/09/SST-v15.0.0-Released/, this was not included.  Between that and not making the code modifications in the build system, it was mostly forgotten about.

Making this change doesn't require elements changes, since it picks up whatever Python is in the config file during compile time (for example, https://github.com/sstsimulator/sst-elements/blob/2120920858b871f1d14c719b393c2675149ff27a/config/sst_check_numpy.m4#L10).

This change consists of two required parts:
1. Update the minimum required version in the build system
2. Ensure the CI jobs are updated

I've marked this as major since it is technically breaking and was never announced.

berquist:
I forgot to add, a not-required part would be updating all the Python code to add language features from 3.7, 3.8, and 3.9.  It's not required because no breaking changes in the language means none of the existing code will break, only that it can be modernized in some places.  This would happen later.

berquist:
It actually was "announced" on http://sst-simulator.org/SSTPages/SSTElementReleaseMatrix/.

jleidel:
TCL test infrastructure has been updated to reflect the changes

jwilso:
@berquist @gvoskuilen Making this future, progress can be assumed when TOSS repository access is acquired.

gvoskuilen:
This is complete, minimum Python version is now 3.9.

PR #1384, opened 2025-07-11
Enforce minimum Python version of 3.9

berquist:
Part of #1376.  There should be follow-ups to:
- Modernize the Python code in the test framework to use 3.9 language constructs
- Remove any pre-3.9-specific CPython code from the source

This PR will fail if there are any jobs using pre-3.9 versions.

### DIFF
--- a/config/sst_check_python.m4
+++ b/config/sst_check_python.m4
@@ -16,8 +16,8 @@ dnl check if user provided a specific python-config
 dnl search python3-config
   AS_IF([test $PYTHON_CONFIG_EXE = "no"],
     [AS_IF([test -n "$with_python"],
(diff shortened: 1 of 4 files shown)


feldergast (approved):

feldergast:
Converting this to draft for now so we don't accidentally merge.  At least one of our nightly tests is still using Python 3.6.

berquist:
Random failure?
```
======================================================================
FAIL: test_prospero_binary_withtimingdram_using_TAR_traces (testsuite_default_prospero.testcase_prospero)
----------------------------------------------------------------------
  File "testsuite_default_prospero.py", line 20, in setUp
    self._create_prospero_PIN_trace_files()
  File "testsuite_default_prospero.py", line 230, in _create_prospero_PIN_trace_files
    self.assertTrue(rtn.result() == 0, "array.c failed to compile")
AssertionError: False is not true : array.c failed to compile
```

PR #1395, opened 2025-07-26
Simplify serialization of std::variant by using std::array with bounds check

leekillough:
This simplifies the `std::variant` serialization code
* Use `std::array` instead of C-style array of function pointers for `set_index`
* Defines `set_index` as a `struct` inside of the `serialize_impl` class
* Use `.at()` method for runtime array bounds checking
* Remove `// clang-format off` / `// clang-format on` region mark since the new code is formatted well with `clang-format`


feldergast (dismissed):

feldergast:
This PR is failing with a missing symbol in the coreTestElement library:


dlopen(/Users/sstbuilder/jenkins/workspace/workspace/SST__AutotestGen2_NewFW_OSX-14-XC15-ARM2_OMPI-4.1.6_PY3.10_sst-elements/702_2025-08-14_09-54-51/local/sst-core/lib/sstcore/libcoreTestElement.so, 0x0009): symbol not found in flat namespace '__ZN3SST4Core13Serialization14serialize_implINSt3__17variantIJNS3_6vectorIiNS3_9allocatorIiEEEEdNS3_12basic_stringIcNS3_11char_traitsIcEENS6_IcEEEEEEEvE9set_indexINS3_16integer_sequenceImJLm0ELm1ELm2EEEEEE'
 
 c++filt on the symbol name gives this:


SST::Core::Serialization::serialize_impl<std::__1::variant<std::__1::vector<int, std::__1::allocator<int>>, double, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>>, void>::set_index<std::__1::integer_sequence<unsigned long, 0ul, 1ul, 2ul>>

leekillough:
The error may be caused by a mismatch in dynamic libraries and headers triggered by the complicated nested template definition.

I have worked around it by making `set_index` a nested class inside of `serialize_impl` with a static `array` member. I have moved the definition of `set_index` to the inside of `serialize_impl`.

Rather than using `static_assert` with always-`false` conditions depending on a template parameter, I will get a compilation error if the `std::index_sequence` partial specialization does not match and there is no `array` member.

leekillough:
This is another one which has been sitting out there.

feldergast (approved):

PR #1396, opened 2025-07-28
Serialize std::unique_ptr

leekillough:
Serialize `std::unique_ptr`. Wrappers can be used to support unbounded arrays with runtime size, and custom deleters.


leekillough:
Design notes: [Smart Pointer Serialization.pptx](https://github.com/user-attachments/files/21819429/Smart.Pointer.Serialization.pptx)

Issue #359, opened 2018-08-15
Log Histogram


afrodri:
Histogram statistics are useful, but often require an inordinately large number of bins. There are cases where fixed linear bin widths (e.g. 1-10,11-20,21-30,...) are not necessary and logarithmic bins could be used (e.g. 0-1,1-2,3-4,5-8,9-16,17-32). A new logHistogram statistic type could address this.

jjwilke:
Did anything happen with this?

Do we want a new LogHistogram object? Or an option to the existing histogram to do things logarithmically?

This might be as easy as virtualizing the `addData_impl` function to pass through log values to the original `addData_impl` in the regular histogram.

leekillough:
Related: https://github.com/sstsimulator/sst-core/pull/1410

PR #1409, opened 2025-08-20
loghistogram #359

ct-clmsn:
initial import of `LogBinHistogram` class implementation for #359


feldergast (changes requested):
You have at least one merge from the master branch into this branch, which is not allowed.  It is also preferred that you rebase against level instead of merging, especially over so long of a period. Could you rebase this branch on current devel and then do a force push to clean things up.

PR #1410, opened 2025-08-20
LogHistogram

ct-clmsn:
initial import of LogBinHistogram for #359

### DIFF
--- a/src/sst/core/statapi/statlghistogram.h
+++ b/src/sst/core/statapi/statlghistogram.h
@@ -83,7 +83,7 @@ public:
         m_itemsBinnedCount = 0;
         this->setCollectionCount(0);
 
-        for(std::size_t i = 0; i < m_numBins; ++i) {
+        for( std::size_t i = 0; i < m_numBins; ++i ) {
             m_binsMap.insert(std::pair<BinDataType, CountType>(i, static_cast<CountType>(0)));
         }
     }
@@ -142,8 +142,8 @@ protected:
         // To support signed and unsigned values along with floating point types,
         // the calculation to find the bin_start value must be done in floating point
         // then converted to BinDataType
-        double const log2_value = std::log2(static_cast<double>(value));
-        double const bin_floor_value = std::floor(log2_value); // Find the floor of the value
+        double const      log2_value = std::log2(static_cast<double>(value));
+        double const      bin_floor_value = std::floor(log2_value); // Find the floor of the value
         BinDataType const bin_start = static_cast<BinDataType>(bin_floor_value);
         //      printf("DEBUG: value = %d, junk1 = %f, calc2 = %f, calc3 = %f : bin_start = %d, item count = %ld, \n",
         //      value, calc1, calc2, calc3, bin_start, getStatCollectionCount());
@@ -179,8 +179,8 @@ private:
     */
     CountType getBinCountByBinStart(BinDataType const binStartValue)
     {
-        double const log2_value = std::log2(static_cast<double>(binStartValue));
-        double const bin_floor_value = std::floor(log2_value); // Find the floor of the value
+        double const      log2_value = std::log2(static_cast<double>(binStartValue));
+        double const      bin_floor_value = std::floor(log2_value); // Find the floor of the value
         BinDataType const bin_start = static_cast<BinDataType>(bin_floor_value);
 
         // Find the Bin Start Value in the Bin Map
@@ -211,7 +211,7 @@ private:
         // Compute the max value; the histogram is indexed by the log2 floor of the bin
         // return pow2 of 1 + the number of total bins
         // 
-        return static_cast<BinDataType>(std::pow(2, m_numBins+1)); // (m_binWidth * m_numBins) + m_minValue - 1;
+        return static_cast<BinDataType>(std::pow(2, m_numBins+1));
     }
 
     /**
--- a/src/sst/core/statapi/statlghistogram.h
+++ b/src/sst/core/statapi/statlghistogram.h
@@ -1,0 +1,382 @@
+// Copyright 2009-2025 NTESS. Under the terms
+// of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Copyright (c) 2009-2025, NTESS
+// All rights reserved.
+//
+// This file is part of the SST software package. For license
+// information, see the LICENSE file in the top level directory of the
+// distribution.
+
+#ifndef SST_CORE_STATAPI_STATLGHISTOGRAM_H
+#define SST_CORE_STATAPI_STATLGHISTOGRAM_H
+
+#include "sst/core/sst_types.h"
+#include "sst/core/statapi/statbase.h"
+#include "sst/core/statapi/statoutput.h"
+#include "sst/core/warnmacros.h"
+
+namespace SST::Statistics {
+
+// NOTE: When calling base class members in classes derived from
+//       a templated base class.  The user must use "this->" in
+//       order to call base class members (to avoid a compiler
+//       error) because they are "nondependant named" and the
+//       templated base class is a "dependant named".  The
+//       compiler will not look in dependant named base classes
+//       when looking up independent names.
+// See: http://www.parashift.com/c++-faq-lite/nondependent-name-lookup-members.html
+
+/**
+    \class HistogramStatistic
+    Holder of data grouped into pre-determined width bins.
+    \tparam BinDataType is the type of the data held in each bin (i.e. what data type described the width of the bin)
+*/
+#define CountType   uint64_t
+#define NumBinsType uint32_t
+
+template <class BinDataType>
+class LogBinHistogramStatistic : public Statistic<BinDataType>
+{
+
+public:
+    SST_ELI_DECLARE_STATISTIC_TEMPLATE_DERIVED(
+        LogBinHistogramStatistic,
+        BinDataType,
+        "sst",
+        "LogBinHistogramStatistic",
+        SST_ELI_ELEMENT_VERSION(1, 0, 0),
+        "Track distribution of statistic across bins",
+        "SST::Statistic<T>")
+
+    SST_ELI_DOCUMENT_PARAMS(
+        {"minvalue", "The minimum data value to include in the histogram.", "0"},
+        {"numbins",  "The number of histogram bins.", "100"},
+        {"dumpbinsonoutput", "Whether to output the data range of each bin as well as its value.", "true"},
+        {"includeoutofbounds", "Whether to keep track of data that falls below or above the histogram bins in separate out-of-bounds bins.", "true"})
+
+
+    LogBinHistogramStatistic(
+        BaseComponent* comp, const std::string& statName, const std::string& statSubId, Params& statParams) :
+        Statistic<BinDataType>(comp, statName, statSubId, statParams)
+    {
+        // Identify what keys are Allowed in the parameters
+        std::vector<std::string> allowedKeySet;
+        allowedKeySet.push_back("minvalue");
+        allowedKeySet.push_back("numbins");
+        allowedKeySet.push_back("dumpbinsonoutput");
+        allowedKeySet.push_back("includeoutofbounds");
+        statParams.pushAllowedKeys(allowedKeySet);
+
+        // Process the Parameters
+        m_minValue           = statParams.find<BinDataType>("minvalue", 0);
+        m_numBins            = statParams.find<NumBinsType>("numbins", 100);
+        m_dumpBinsOnOutput   = statParams.find<bool>("dumpbinsonoutput", true);
+        m_includeOutOfBounds = statParams.find<bool>("includeoutofbounds", true);
+
+        // Initialize other properties
+        m_totalSummed      = 0;
+        m_totalSummedSqr   = 0;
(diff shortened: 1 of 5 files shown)


src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+//       when looking up independent names.
+// See: http://www.parashift.com/c++-faq-lite/nondependent-name-lookup-members.html
+
+/**
+    \class HistogramStatistic
+    Holder of data grouped into pre-determined width bins.
+    \tparam BinDataType is the type of the data held in each bin (i.e. what data type described the width of the bin)
+*/
+#define CountType   uint64_t
+#define NumBinsType uint32_t
```
leekillough, line 37: Unless macros are really needed, I would suggest `using`:
```
using CountType   = uint64_t;
using NumBinsType = uint32_t;

File: src/sst/core/statapi/statlghistogram.h [hunk 2]
```diff
@@ -0,0 +1,382 @@
...
+namespace SST::Statistics {
+
+// NOTE: When calling base class members in classes derived from
+//       a templated base class.  The user must use "this->" in
+//       order to call base class members (to avoid a compiler
+//       error) because they are "nondependant named" and the
+//       templated base class is a "dependant named".  The
+//       compiler will not look in dependant named base classes
+//       when looking up independent names.
+// See: http://www.parashift.com/c++-faq-lite/nondependent-name-lookup-members.html
```
leekillough, line 29: A simple one-line comment is better than a long explanation:
```
this->setCollectionCount(0);  // look in dependent base class Statistic<BinDataType>
```

src/sst/core/statapi/statlghistogram.h
ct-clmsn: Type aliasing is also my preference. The class implementation was largely lifted from the existing Histogram class - wasn't sure if modernizing to `using` keyword over `#define` macros was preferable. Will make those changes tomorrow.

src/sst/core/statapi/statlghistogram.h
ct-clmsn: yep, this was a byproduct of the existing Histogram lift

src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+    LogBinHistogramStatistic(
+        BaseComponent* comp, const std::string& statName, const std::string& statSubId, Params& statParams) :
+        Statistic<BinDataType>(comp, statName, statSubId, statParams)
+    {
+        // Identify what keys are Allowed in the parameters
+        std::vector<std::string> allowedKeySet;
+        allowedKeySet.push_back("minvalue");
+        allowedKeySet.push_back("numbins");
+        allowedKeySet.push_back("dumpbinsonoutput");
+        allowedKeySet.push_back("includeoutofbounds");
```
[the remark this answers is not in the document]
ct-clmsn: using initializer lists is the preference? cool, those changes can happen tomorrow.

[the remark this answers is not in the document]
feldergast: The coding standard doesn't specify a preference for this case, but I think going with initializer lists makes sense.  We may want to go back and update some of the older code at some point, as well.

src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+
+        // Was the bin found?
+        if ( bin_itr == m_binsMap.end() ) {
+            // No, Create the bin and set a value of 1 to it
+            m_binsMap.insert(std::pair<BinDataType, CountType>(bin_start, static_cast<CountType>(N)));
+        }
+        else {
+            // Yes, Increment the specific bin's count
+            bin_itr->second += static_cast<CountType>(N);
+        }
```
leekillough, line 161:
```c++
        HistoMapItr_t bin_itr = m_binsMap.find(bin_start);

        // Was the bin found?
        if ( bin_itr == m_binsMap.end() ) {
            // No, Create the bin and set a value of 1 to it
            m_binsMap.insert(std::pair<BinDataType, CountType>(bin_start, static_cast<CountType>(N)));
        }
        else {
            // Yes, Increment the specific bin's count
            bin_itr->second += static_cast<CountType>(N);
        }
```
... can be replaced with one line:

```c++
m_binsMap[bin_start] += static_cast<CountType>(N); // Increment the specific bin's count
```
The `std::map` `operator[]` creates a value-initialized new entry if one does not exist, and returns a reference to the entry. _Value-initialized_ means `0` for arithmetic types.

src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+
+private:
+    /** Count how many bins are active in this histogram */
+    NumBinsType getActiveBinCount() { return m_binsMap.size(); }
+
+    /** Count how many bins are available */
+    NumBinsType getNumBins() { return m_numBins; }
+
+    /** Get the width of a bin in this histogram */
+    NumBinsType getBinWidth() { return m_binWidth; }
```
leekillough, line 174: Getters should usually be `const`.

i.e.

```c++
    /** Count how many bins are active in this histogram */
    NumBinsType getActiveBinCount() const { return m_binsMap.size(); }

    /** Count how many bins are available */
    NumBinsType getNumBins() const { return m_numBins; }

    /** Get the width of a bin in this histogram */
    NumBinsType getBinWidth() const { return m_binWidth; }
```
Also maybe `getBinCountByBinStart` which follows.

src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+        Sum up every item presented for storage in the histogram
+        \return The sum of all values added into the histogram
+    */
+    BinDataType getValuesSummed() { return m_totalSummed; }
+
+    /**
+    Sum up every squared value entered into the Histogram.
+    \return The sum of all values added after squaring into the Histogram
+    */
+    BinDataType getValuesSquaredSummed() { return m_totalSummedSqr; }
```
leekillough, line 247: Same comment as above for `const` getters.

src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+            m_Fields.push_back(statOutput->registerField<CountType>("NumOutOfBounds-MinValue"));
+            m_Fields.push_back(statOutput->registerField<CountType>("NumOutOfBounds-MaxValue"));
+        }
+
+        // Do we also need to dump the bin counts on output
+        if ( true == m_dumpBinsOnOutput ) {
+            BinDataType binLL;
+            BinDataType binUL;
+
+            for ( uint32_t y = 0; y < getNumBins(); y++ ) {
```
leekillough, line 283:
```c++
for ( NumBinsType y = 0; y < getNumBins(); y++ ) {
```
`getNumBins()` returns `NumBinsType` so for consistency.

src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+    // values such as variance.
+    BinDataType m_totalSummedSqr;
+
+    // A map of the the bin starts to the bin counts
+    HistoMap_t m_binsMap;
+
+    // Support
+    std::vector<StatisticOutput::fieldHandle_t> m_Fields;
+    bool                                        m_dumpBinsOnOutput;
+    bool                                        m_includeOutOfBounds;
```
leekillough, line 375: For general defensive programming, all class members which are not defined by default initialization (integral types including `bool` and `enum`, floating-point, regular pointers) should be default-initialized to a value such as `0` or `nullptr`, so that they are initialized even if the constructor does not initialize them. The easiest way to do this is to put `{ }` braces after the member declarator, which initializes them to `0`, `false`, `0.0`, `nullptr`, etc. (that's what Rev does), but you can use equals if that looks better. That way they're always initialized to something, and you don't chase bugs with random garbage because you forgot to initialize them in constructors. It also sometimes makes it easier because then you don't have to repeat the names in constructors and make sure you got every name, and the initialization is clear at the point of definition. There is no penalty if you initialize them in constructors, because then the constructor initializers override any default initializers.

src/sst/core/statapi/statlghistogram.h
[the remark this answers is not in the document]
leekillough: > Feel free to modernize.

> The coding standard doesn't specify a preference for this case, but I think going with initializer lists makes sense. We may want to go back and update some of the older code at some point, as well.

I have not reviewed other SST stats code before. This is just my objective review of this PR. I am not implying anything about old code which was cloned, and I understand if old code was cloned instead of designing and writing this PR's code from scratch, etc.

src/sst/core/statapi/statlghistogram.h
```diff
@@ -0,0 +1,382 @@
...
+    /**
+        Get the largest possible value represented by this histogram (i.e. the highest value in any of items bins
+       rounded above to the size of the bin)
+    */
+    BinDataType getBinsMaxValue()
+    {
+        // Compute the max value; the histogram is indexed by the log2 floor of the bin
+        // return pow2 of 1 + the number of total bins
+        // 
+        return static_cast<BinDataType>(std::pow(2, m_numBins + 1));
```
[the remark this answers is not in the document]
leekillough: `ldexp()` can be used for general floating-point results of arbitrary magnitude. Much better than `pow()`.

src/sst/core/statapi/statlghistogram.h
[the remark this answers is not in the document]
ct-clmsn: > The coding standard doesn't specify a preference for this case, but I think going with initializer lists makes sense. We may want to go back and update some of the older code at some point, as well.

Yep, a lot of this code was migrated directly from the histogram class. I kept lots of it to avoid a coding standard conflict.

ct-clmsn:
@feldergast I'm not sure what the linting check is unhappy with - could use a second glance.

berquist:
I had to approve your workflow to run, but you can see it here: https://github.com/sstsimulator/sst-core/actions/runs/17130304179/job/48593800032?pr=1410

ct-clmsn:
@berquist those lines look equivalent ... it's kind of throwing me off, is this a spacing issue?

feldergast:
If you have pre-commit set up on your system, you can just run "pre-commit run --all" and it should do the proper formatting as part of those checks.  If not, getting that working may be easier than getting clang-format20 on your system.

berquist:
It is hard to read from the diff, because most don't display it well, but if you do what Scott says, you'll see that it's trailing whitespace.  You can also see that "trim trailing whitespace" failed before running clang-format.

feldergast:
Looks like core has diverged too far.  This PR fails with missing type declarations.  Will need to be rebased, fix and repushed.

leekillough:
I have created https://github.com/leekillough/sst-core/tree/loghistogram which is merged with `devel` and squashed with `git reset --soft` but I get compilation errors which indicate possibly unfinished work, or dependence on removed features like `StatisticBase::StatMode_t` and `StatisticBase::STAT_MODE_COUNT`:
```
In file included from ../../../../src/sst/core/statapi/statbase.cc:20:
../../../../src/sst/core/statapi/statlghistogram.h:308:30: error: ‘SST::Statistics::StatisticBase::StatMode_t’ has not been declared
  308 |     bool isStatModeSupported(StatisticBase::StatMode_t mode) const override
      |                              ^~~~~~~~~~~~~
../../../../src/sst/core/statapi/statlghistogram.h: In member function ‘bool SST::Statistics::LogBinHistogramStatistic<BinDataType>::isStatModeSupported(int) const’:
../../../../src/sst/core/statapi/statlghistogram.h:311:29: error: ‘STAT_MODE_COUNT’ is not a member of ‘SST::Statistics::StatisticBase’
  311 |         case StatisticBase::STAT_MODE_COUNT:
      |                             ^~~~~~~~~~~~~~~
../../../../src/sst/core/statapi/statlghistogram.h:312:29: error: ‘STAT_MODE_PERIODIC’ is not a member of ‘SST::Statistics::StatisticBase’
  312 |         case StatisticBase::STAT_MODE_PERIODIC:
      |                             ^~~~~~~~~~~~~~~~~~
../../../../src/sst/core/statapi/statlghistogram.h:313:29: error: ‘STAT_MODE_DUMP_AT_END’ is not a member of ‘SST::Statistics::StatisticBase’
  313 |         case StatisticBase::STAT_MODE_DUMP_AT_END:
      |                             ^~~~~~~~~~~~~~~~~~~~~
../../../../src/sst/core/statapi/statlghistogram.h: In instantiation of ‘class SST::Statistics::LogBinHistogramStatistic<int>’:
../../../../src/sst/core/statapi/statbase.cc:247:1:   required from here
../../../../src/sst/core/statapi/statlghistogram.h:308:10: error: ‘bool SST::Statistics::LogBinHistogramStatistic<BinDataType>::isStatModeSupported(int) const [with BinDataType = int]’ marked ‘override’, but does not override
  308 |     bool isStatModeSupported(StatisticBase::StatMode_t mode) const override
      |          ^~~~~~~~~~~~~~~~~~~
```
If this feature is still desired, someone else will need to spend some time bringing the code up-to-date.

jleidel:
@gvoskuilen any guidance here to port the code to the latest stats codebase?

PR #1435, opened 2025-10-07
Fix a missing #include <vector>

leekillough:
In some build environments, `#include <vector>` needs to be added to `testElements/coreTest_OverheadMeasure.h`.

@kpgriesser


feldergast:
Fixed in #1434

Issue #1446, opened 2025-10-16
Compiler error (when building a component) in SST serialization header when using CCE


stonea:
Depending on the compiler I use, sometimes I run into the following error when I try and build a third party SST component that makes use of serialization:

```
In file included from /home/users/stonea/sst/sst-core/install/include/sst/core/serialization/serialize.h:436:0,
                 from /home/users/stonea/sst/sst-core/install/include/sst/core/serialization/serializable.h:16,
                 from /home/users/stonea/sst/sst-core/install/include/sst/core/mempool.h:15,
                 from /home/users/stonea/sst/sst-core/install/include/sst/core/activity.h:15,
                 from /home/users/stonea/sst/sst-core/install/include/sst/core/event.h:15,
                 from /home/users/stonea/sst/sst-core/install/include/sst/core/interfaces/stringEvent.h:15,
                 from cell.cpp:2:
/home/users/stonea/sst/sst-core/install/include/sst/core/serialization/impl/serialize_variant.h: In lambda function:
/home/users/stonea/sst/sst-core/install/include/sst/core/serialization/impl/serialize_variant.h:108:80: error: parameter packs not expanded with ‘...’:
                 [](std::variant<Types...>& obj) { obj.template emplace<INDEX>(); }...
                                                                                ^
/home/users/stonea/sst/sst-core/install/include/sst/core/serialization/impl/serialize_variant.h:108:80: note:         ‘INDEX’
/home/users/stonea/sst/sst-core/install/include/sst/core/serialization/impl/serialize_variant.h: At global scope:
/home/users/stonea/sst/sst-core/install/include/sst/core/serialization/impl/serialize_variant.h:108:83: error: expansion pattern ‘<lambda>’ contains no argument packs
                 [](std::variant<Types...>& obj) { obj.template emplace<INDEX>(); }...
```

This is referring to a template that was added to SST in <https://github.com/sstsimulator/sst-core/pull/1344>.

For some reason it seems like it's expecting us to unpack the parameters within the lambda instead of outside of it.

I'm not strong at template metaprogramming, so I tried prompting an LLM tool to "refactor so the thing the parameter pack expression (...) is being applied to is factored out into a separate template" (thinking that doing so would disambiguate where ... is being applied.  The tool suggested the following:

```
template <typename Variant, size_t Index>
constexpr void set_variant_index(Variant& obj) {
    obj.template emplace<Index>();
}

template <typename... Types>
template <size_t... INDEX>
[...]

PR #1447, opened 2025-10-18
Work around bug in some compilers which do not expand lambda expressions across parameter packs correctly

leekillough:
This should fix https://github.com/sstsimulator/sst-core/issues/1446.

It is probably related to [this historic bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47226) and [its fix](https://gcc.gnu.org/legacy-ml/gcc-patches/2017-08/msg01678.html).  This bug probably still exists in some toolchains like CCE.

Rather than creating lambda expressions which contain a template pack parameter (in this case, `INDEX`), and then expanding the lambdas across all instances of the pack parameter, a templated function which takes the pack parameter is called with the pack parameter for each instance. This avoids the compiler from having to generate lambda expressions across parameter packs.

There is no serialization performance impact for this change, since it still results in an array of function pointers.

@stonea


stonea:
I've confirmed this removes the error message I was seeing with Cray Clang 19 on my end.

PR #1452, opened 2025-10-21
Interactive Console Major Update

kpgriesser:
Major Changes:

- config.h, config.cc

  Add support for interactive realtime action, sst.rt.interactive (#1266) 
  Add checkpoint-enable flag to enable checkpointing in interactive console
  Make sst.interactive.simpledebug the default for interactive console 
  Only use --interactive-console to specify a custom IC File 
  replay option ( --replay )

- from_string.h:

  Full precision printing of float types. Scientific notation when value is greated than 10**6 or less than 10**-6

- impl/interactive/Makefile.inc

  Included cmdLineEditor files for build

- impl/interactive/cmdLineEditor.h, impl/interactive/cmdLineEditor.cc

  In-line command editing support

- impl/interactive/simpleDebug.h, impl/interactive/simpleDebug.cc

  Shutdown support (#1311) Watchpoint support 
  Add checkpoint-enabled flag to enable in interactive console
  Make sst.interactive.simpledebug the default of interactive console
  Only use --interactive-console to specify custom IC
  Allow for empty command and other minor whitespace annoyances
  Console command logging and replay. Support commented lines for replay scripts
  Consolidated commands into a command funtion table. 
  Standardized specifying commands, help text, and command groups.
  Provided auto-completion support for commands and object names.
  Various nullptr checks and fixes to avoid segmentation faults.

- interactiveConsole.h

  Add support for shutdown (#1311)

- serialization/objectMap.cc, serialization/objectMap.h

  Support for trace buffers and watchpoint actions.
  Circular buffer and trigger recording Add ability to check/sample before and after both clock and event handlers.
  Watchpoint support for adding set var val action.
  Extended trigger tests functionality: support for logic operators (e.g. v1 < 10 && v2 > 5)  and comparisons between two variables
  Support for trigger comparison between two variables.
  All set string to handle spaces. watchpoint verbosity control Various bug fixes for command line editing.
  Fixes for nullptr checks and address sanitizer issues causing seg faults.

- simulation.cc

  Interactive console realtime action (#1266).
 Addition of checkpoint-enable flag to enable checkpoint in IC. 
 Make sst.interactive.simpledebug the default for IC replay option for IC

- watchPoint.h, watchPoint.cc

  Circular buffer, trigger record. 
  Add ability to check/sample before and after both clock and event handlers.
  Add watchpoint action class and add set var val action.
  Add support for trigger comparison between two variables Update watchpoint handling.
  Allow set string to handle spaces
  Add shutdown as trigger action.
  Add watchpoint index to watchpoint class.
  Add checkpoint-enable flag to enable checkpointing in interactive console.
  Makesst.interactive.simpledebug the default for interactive console.
  Only use --interactive-console to specify a custom IC.
  watchpoint verbosity control.

### DIFF
--- a/src/sst/core/Makefile.am
+++ b/src/sst/core/Makefile.am
@@ -78,6 +78,7 @@ nobase_dist_sst_HEADERS = \
 	math/sqrt.h \
 	uninitializedQueue.h \
 	unitAlgebra.h \
+	watchPoint.h \
 	eli/elementinfo.h \
 	eli/elibase.h \
 	eli/attributeInfo.h \


kpgriesser:
@gvoskuilen  I pushed an update to include watchPoint.h in the distribution. This is to support work exploring an external tool using the interactive console.

src/sst/core/watchPoint.cc
```diff
@@ -9,6 +9,7 @@
 // information, see the LICENSE file in the top level directory of the
 // distribution.
 
+//clang-format off
```
feldergast: Why is clang-format turned of here?  They seem to be in the order specified in the .clang-format file.

PR #1455, opened 2025-10-22
Add statistics to port modules

gvoskuilen:


berquist:
The cmake-format error is not helpful, but the fact that pre-commit passes means it's a problem with https://github.com/puneetmatharu/cmake-format-lint-action specifically.  Possible solutions are updating the version of the action or removing it altogether, since it is supposed to be handled by https://github.com/sstsimulator/sst-core/blob/29e833899e55acc37e54103ee6e418714ca3f817/.pre-commit-config.yaml#L21-L30.

gvoskuilen:
I think the pre-commit check should be sufficient for this and then we remove one extra dependency that could (did) break.

Issue #1439, opened 2025-10-09
Add `./scripts/test-includes.pl` to CI


berquist:
> We should see if there is a practical way to add `./scripts/test-includes.pl` as a regular part of CI. When run without arguments, it tests for missing `#include`s and returns a nonzero exit code if there are any missing, as well as describe which ones are missing. If run with `--fix`, it automatically fixes them in your working tree (you still need to `git add` and `git commit` the changes).

> _Originally posted by @leekillough in https://github.com/sstsimulator/sst-core/issues/1423#issuecomment-3325046237_

We should add this as a local hook under pre-commit.

berquist:
Using https://github.com/sstsimulator/sst-core/pull/1442/commits/ecd37cb2dea2ff3f6f72be57ac6886bcde5ea04c from #1442 on itself, `time ./scripts/test-includes.pl --fix` results are:
- On Arch with GCC 15.2.1 on a Threadripper 2950X, `Missing #includes found and fixed in 41 files. ./scripts/test-includes.pl --fix  65.98s user 89.10s system 101% cpu 2:32.51 total`.
- On Rocky 8.6, the Memoize module is missing.  When adding to CI, it is possible the bare Perl package from a distro will not provide this, in which case it needs to be installed from the distro package manager or CPAN.

Even if we assume a time lower bound of one minute, this is too slow for a pre-commit hook, and it will be frustrating for people using pre-commit.[^1]  pre-commit also isn't currently capable of automatically managing Perl: the Perl install on `PATH` will have to have the correct modules already installed.  This check will have to go in regular GitHub Actions CI.  Running the script locally outside of pre-commit for those interested is easy.

[^1]: I know some pre-commit hooks are already slow right now at `pre-commit run -a  25.37s user 24.03s system 579% cpu 8.531 total`, which I'm looking to reduce or at least not increase.

leekillough:
Please see the latest version at https://github.com/sstsimulator/sst-core/pull/1442

It runs in under 15 seconds on SST core on a recent machine. Elements, maybe one minute.

leekillough:
I could add a Perl `require vx.y.z` but I am not using the very latest. I don't know about MacOS but any recent Linux distribution should support it.

leekillough:
Even if it is too slow for `pre-commit` interactive use, it should be acceptable for CI tests.

leekillough:
[...]

PR #1457, opened 2025-10-24
adding missing header file to appease test-includes script

jleidel:
Fixes a missing header in the `checkpointableInfo.h` header file.  passes all tests, include the test-headers.pl script.

### DIFF
--- a/src/sst/core/eli/checkpointableInfo.h
+++ b/src/sst/core/eli/checkpointableInfo.h
@@ -18,8 +18,8 @@
 #include <cstdint>
 #include <iostream>
 #include <string>
-#include <vector>
 #include <type_traits>
+#include <vector>
 
 namespace SST::ELI {
 


leekillough:
We need to figure out why the script did not automatically detect and fix this when it was run before. It might not be detecting the missing header on some platforms for some reason.

If this was due to a recent PR merge, we need to ask people to be run it manually before every PR until we get it integrated into `pre-commit` or the CI tests.

leekillough:
Looks like it  was introduced with https://github.com/sstsimulator/sst-core/pull/1451/files#diff-f2ce33e134432cb9242c180d19a3630aefccb435a47b2153cc44c8c5950fb4cfR1

@feldergast @gvoskuilen , Please run `scripts/test-includes.pl` before every PR until we get it [integrated into CI](https://github.com/sstsimulator/sst-core/issues/1439).

jleidel:
> We need to figure out why the script did not automatically detect and fix this when it was run before. It might not be detecting the missing header on some platforms for some reason.
> 
> If this was due to a recent PR merge, we need to ask people to be run it manually before every PR until we get it integrated into `pre-commit` or the CI tests.

On the TCL CI systems, we run the script, but don't implicitly fix the issues.

leekillough:
> On the TCL CI systems, we run the script, but don't implicitly fix the issues.

See https://github.com/sstsimulator/sst-core/pull/1458. GCC versions prior to 10 do not find the missing `#include <type_traits>` for `std::void_t`. The feature to find missing `#include`s requires GCC version 8 or higher. The changes in https://github.com/sstsimulator/sst-core/pull/1458 make it work on MacOS and Linux, and it prints more precise error messages if a suitable version cannot be found. If the GCC version works for some `#include` files like `<vector>`, a warning message is printed that some missing `#include` files might not be found if the version is below 10.

PR #1458, opened 2025-10-24
Refine test-includes.pl to print more precise error messages about compiler requirements; work with more versions of GCC

leekillough:
This  makes `scripts/test-includes.pl` work with more versions of GCC, and to print more precise error messages if the GCC compiler cannot be found or does not produce the expected warning outputs.

GCC version 8 is the minimum version required when this `#include` warning was added; however, some missing `#include`s such as https://github.com/sstsimulator/sst-core/pull/1457 are not found until GCC version 10 (version 8 finds a missing `#include <vector>` for `std::vector` but does not find a missing `#include <type_traits>` for `std::void_t` until version 10 -- I looked in the Bugzilla database, but I could not find a confirmed fix).

It has also been tested on MacOS. It finds `g++-<nnn>` if they exist, since `g++` is usually aliased to `clang`. More helpful error messages are printed about installing GCC on MacOS if a suitable version cannot be found.

@berquist

### DIFF
--- a/scripts/test-includes.pl
+++ b/scripts/test-includes.pl
@@ -1,5 +1,6 @@
 #!/usr/bin/perl
 
+use v5.10.0;
 use strict;
 use warnings qw(FATAL all);
 use Cwd qw(realpath);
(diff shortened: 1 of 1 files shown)


leekillough:
I tested all GCC versions on [godbolt.org](https://godbolt.org), but none of them warn about missing `#include <algorithm>` when `std::copy_if` is used. So apparently this GCC missing `#include` warning only handles a subset of names defined in system headers.

It is better than nothing, but sometimes a missing `#include` can only be found with a build error on certain platforms.

feldergast (approved):

berquist:
After this change I'm now getting
```
$ ../sst-core/scripts/test-includes.pl --fix
g++ (Spack GCC) 14.3.0
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Argument "14.3.0\n" isn't numeric in int at ../sst-core/scripts/test-includes.pl line 44, <$comp> line 5.
```

leekillough:
> After this change I'm now getting
> 
> ```
> $ ../sst-core/scripts/test-includes.pl --fix
> g++ (Spack GCC) 14.3.0
> Copyright (C) 2024 Free Software Foundation, Inc.
> This is free software; see the source for copying conditions.  There is NO
> warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
> 
> Argument "14.3.0\n" isn't numeric in int at ../sst-core/scripts/test-includes.pl line 44, <$comp> line 5.
> ```
https://github.com/sstsimulator/sst-core/pull/1466

PR #1464, opened 2025-10-28
Add cast to remove errors on some platforms

leekillough:
@feldergast @gvoskuilen 

This PR is experimental. Please try it and see if it removes errors building on some platforms.

It uses a cast which normally should not be necessary.


leekillough:
Since `std::shared_ptr` and `std::weak_ptr` serialization is disabled for 15.1, these build errors on certain platforms need to be investigated and possibly worked around.

PR #1465, opened 2025-10-28
Fix more missing headers

leekillough:


feldergast:
We will run the header file check right after all other changes have stabilized and before we branch for the release.

leekillough:
It was not just the result of running the script -- I removed redundant `#include`s by moving them to `.h` files instead of `.cc` files.

TCL's CI runs `scripts/test-includes.pl` before it builds and runs SST tests. So it's a little more pedantic. I'm submitting PRs to address any issues we find. I'm sorry if my submitting a lot of PRs slows down the process. If you want me to submit all of my changes in one big PR, I can, but my instinct and habits are to submit separate PRs for every little change, so that they are isolated.

leekillough:
This should be reopened and merged. It's not simply the result of running `scripts/test-includes.pl --fix`.

PR #1468, opened 2025-10-28
Various release fixes

gvoskuilen:
* Add PortModules to sst-info output
* Update list of python version-specific binaries to search
* Fix issue in sst test framework
* Disable compilation of shared_ptr serialization - confirmed that this fixes a compile error on older Mac platforms


leekillough:
If I could access an OSX 14 machine, I might be able to fix the build error.

feldergast (approved):

PR #1471, opened 2025-10-30
Fixes for TimeVortex.

feldergast:
- StopAction that detects empty event queue now added to TimeVortex on restarts
- Changed all if's dependent on template parameters to if constexpr in TV implemenations


kpgriesser:
I ran this branch through my testing and they are now all passing! Nice catch!

PR #1472, opened 2025-10-31
Use SST_SER() and SST_SER_NAME() macros instead of calling sst_ser_object()

leekillough:
**Sorry for this last-minute fix, but it's critical for debugging**.

Fix mapping serialization errors which affect debugging, causing `nullptr` to be passed for object map `name`:
- Use `SST_SER()` and `SST_SER_NAME()` macros instead of calling `sst_ser_object()`
  - We might want to make `sst_ser_object()` more private in later releases
- Remove default arguments for `sst_ser_object()` to reduce the chances of surprise errors, such as `std::tuple` serialization calling `sst_ser_object()` with a default `name = nullptr`, even in mapping mode
  - `sst_ser_object()` should only be called by `SST_SER()` and `SST_SER_NAME()`, and should never receive a `nullptr` `name`.

### DIFF
--- a/src/sst/core/serialization/impl/serialize_array.h
+++ b/src/sst/core/serialization/impl/serialize_array.h
@@ -56,7 +56,7 @@ template <typename ELEM_T>
 void
 serialize_array_element(serializer& ser, void* data, ser_opt_t opt, size_t index)
 {
-    sst_ser_object(ser, static_cast<ELEM_T*>(data)[index], opt, nullptr);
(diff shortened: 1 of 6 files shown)


kpgriesser:
Some late testing on mapping data types showed a few cases throwing exceptions causing the debugger to crash. We want to avoid having a user with a fully checkpoint able design find the debugger crashes as soon as he enters a component in the interactive console. 

Workaround is for user to using the no_map option in the serialization macro.

feldergast (approved):

PR #1485, opened 2025-11-17
Reenable serialization of std::shared_ptr, making std::weak_ptr array serialization contingent on SST_SERIALIZE_WEAK_PTR_ARRAY macro

leekillough:
Reenable serialization of `std::shared_ptr`, making `std::weak_ptr` array serialization contingent on a `SST_SERIALIZE_WEAK_PTR_ARRAY` macro.

Many older versions of C++ libraries do not support `std::weak_ptr` correctly, especially `std::weak_ptr` with arrays. This is probably due to much less knowledge and testing of `std::weak_ptr`, as well as [an oversight in the C++17 standard](https://cplusplus.github.io/LWG/issue3001). 

In this PR we enable serialization of `std::weak_ptr` arrays with:

```
#if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 150000
#define SST_SERIALIZE_WEAK_PTR_ARRAY 0
#else
#define SST_SERIALIZE_WEAK_PTR_ARRAY 1
#endif
```
[This libc++ bug fix](https://reviews.llvm.org/D112092) was released in version 15.

[This libstdc++ bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=120561) has a workaround (see the comments in the code changes).

I do not have a MacOS 14 machine to test with.

### DIFF
--- a/src/sst/core/serialization/impl/serialize_shared_ptr.h
+++ b/src/sst/core/serialization/impl/serialize_shared_ptr.h
@@ -15,7 +15,7 @@
 // libc++ versions before 15 define std::weak_ptr<T>::element_type as T instead of std::remove_extent_t<T>, so we must
 // disable serialization of std::weak_ptr arrays.
 // https://reviews.llvm.org/D112092  https://cplusplus.github.io/LWG/issue3001
-#if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 15000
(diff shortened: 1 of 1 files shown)


leekillough:
The [fix is here](https://github.com/llvm/llvm-project/commit/065ac30026d506172994310bd5badd65120f7d74), so I'll try to find the version macro information.

PR #1491, opened 2025-11-24
Reorganize ConfigGraph files

feldergast:
Moved ConfigGraph files to sst/core/model and split configGraph.{h,cc} into multiple smaller files.

Moved the cfgoutput directory into the model directory.

There is only reorganized code in this PR, not new features/code.


src/sst/core/model/Makefile.inc
gvoskuilen: Why are some files (i.e., configComponent.h) listed here while others (cfgouput dir) in the main Makefile.am? It would be cleaner to list all files in `model/` here.

PR #1492, opened 2025-11-25
Add ObjectMultimapCmp functor to sort ObjectMap entries numerically when possible

leekillough:
Add `ObjectMultimapCmp` functor to sort `ObjectMap` entries numerically when possible

This sorts `ObjectMap` keys numerically when possible, falling back to lexicographic comparison if the keys are not numeric.

Add `ObjectMultimap` as the type of the `{name, ObjectMap *}` map sorted according to `ObjectMultimapCmp`.

It also makes `EmptyVars` a static local variable in a function, to get around the static initialization order fiasco.

**Bug fix:** Avoid looking up empty string variable name when setting `CHANGED` watchpoints.

**Bug fix:** Catch exceptions when setting values

Use `catch ( const std::exception& e )` instead of `catch ( std::exception& e )` when catching exceptions

### DIFF
--- a/src/sst/core/serialization/objectMapDeferred.h
+++ b/src/sst/core/serialization/objectMapDeferred.h
@@ -52,7 +52,7 @@ public:
        ObjectMap's child variables. pair.first is the name of the
        variable in the context of this object.
      */
-    const std::multimap<std::string, ObjectMap*>& getVariables() const final override { return obj_->getVariables(); }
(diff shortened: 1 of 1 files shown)


leekillough:
This is also causing test failures which need to be resolved

gvoskuilen:
@leekillough Can you resolve the merge conflicts on this PR?

gvoskuilen (dismissed):

leekillough:
> ?

Sorry, I was out Thursday

PR #1497, opened 2025-12-02
Work around compiler bugs which suggest override when final is used without virtual

leekillough:
Workaround spurious `-Wsuggest-override` compiler warnings on methods marked `final` without `virtual`.

Compilers should not be emitting a warning about missing `override` if `final` is specified without `virtual`, but some do: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=78010

The [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rh-override) suggest that only one of `virtual`, `override` or `final` be specified.

The workaround is to use `final override` instead of just `final`, to suppress warnings about missing `override`.

Virtual methods in derived classes which depend on private member variables, can be marked `final override`.

`isFundamental()` should be marked `final override`, since every class which derives from a fundamental class must stay fundamental.

`getType()` should use `override` but not `final`, since it does not depend on private member variables, and I have written classes like `ObjectMapFundamentalReference` which override `getType()`.

### DIFF
--- a/src/sst/core/serialization/objectMap.h
+++ b/src/sst/core/serialization/objectMap.h
@@ -21,6 +21,7 @@
 #include <cstdio>
 #include <iostream>
 #include <map>
+#include <memory>
 #include <ostream>
 #include <string>
 #include <type_traits>


src/sst/core/timeLord.cc
```diff
@@ -263,7 +263,7 @@ class ObjectMapFundamental<TimeConverter> : public ObjectMap
         setReadOnly(true);
     }
 
-    std::string getType() const final { return demangle_name(typeid(TimeConverter).name()); }
+    std::string getType() const override { return demangle_name(typeid(TimeConverter).name()); }
```
gvoskuilen: Was the removal of `final` intended?

gvoskuilen (approved):

PR #1499, opened 2025-12-09
Add ObjectMap support for SharedObject containers

leekillough:
- Make `initialized` a member variable of `SharedArray`, `SharedMap`, and `SharedSet`, so that its address is persistent and `ObjectMap` can point to it
- Serialize the `published` and `initialized` members in `map_read_only` mode
- If `initialized==true`, serialize the underlying `data` container in mapping mode (currently it's mutable instead of `map_read_only`, but we don't enforce `map_read_only` except on fundamental types)
- Make the `mtx` and `update_mtx` mutexes `static inline` members of their classes, to avoid having to define them outside of the class
- Same for `static SharedObjectDataManager manager;`
- ~~In the update function, acquire both mutexes in a deadlock-free way with C++17 `std::scoped_lock`:~~
- Shorten the names of identifiers, by removing unnecessary namespace prefixes, or by adding `using namespace SST::Core::Serialization` to the `SST::Shared` namespace.
- Add some more missing `#include`s, although my script did not detect them
- Remove some now-unnecessary `#include`s. I did not visually see any dependency on the `#include`, it still builds, and my script does not detect any problems.
- Remove `// clang-format off` around the `#include`s of `simpleDebug.h` so that they use canonical ordering
- Remove a duplicate `#include "simpleDebug.h` from `simpleDebug.cc`

### DIFF
--- a/src/sst/core/shared/sharedObject.cc
+++ b/src/sst/core/shared/sharedObject.cc
@@ -45,8 +45,7 @@ getNumRanks()
 void
 SharedObjectDataManager::updateState(bool finalize)
 {
-    // Wait for both mtx and update_mtx to be locked before updating state
-    std::scoped_lock lock(mtx, update_mtx);
+    std::lock_guard<std::mutex> lock(update_mtx);
 
 #ifdef SST_CONFIG_HAVE_MPI
     // Exchange data between ranks


kpgriesser:
As another data point: We are running address sanitizers on multiple platforms. The previously failing tests affected by this PR are now passing. The remaining failures appear to be unrelated to any of these changes. In other words, green light from our side for merging

PR #1501, opened 2025-12-10
Clean up serialization mapper to use much less code

leekillough:
- This moves all of the `mapper.cc` code into `mapper.h`. The functions are small enough that they can be inlined.
- The `next_item_read_only` `ser_mapper` variable, and the `setNextObjectReadOnly()` function, were unnecessarily complicated
   - The `next_item_read_only` variable had to be cleared after every mapper function, whether it was used or not
   - It is simpler to call `obj_map->setReadOnly()` on the `ObjectMap` object before it is passed to a `ser_mapper` function
- The `indent` `ser_mapper` variable was never used, and as the `ObjectMap` code has progressed, it does not seem necessary anymore
- The `sst_ser` default constructor was not necessary, since `serializer::start_mapping()` always passes it a single constructor argument
- Change the tests for `SerOption::map_read_only` to use the `SerOption::is_set()` function instead of hard-coded `&` test

The new code is around 80 lines smaller than the old code.


feldergast (dismissed):

feldergast:
Looks like this PR needs to be rebased against current devel and retested.  It won't let me resolve the conflicts using the web interface.

leekillough:
I have merged it with `devel`, resolving conflicts, and then `git reset --soft`, `git commit`, and `git push --force`, so there should be no conflicts now.

PR #1507, opened 2025-12-16
Interactive Console Updates

skuntz:
This PR provides a set of updates to the interactive debug console, object mapping support, and bug fixes.

- Ensure signal command line flags are processed before the checkpoint directory is created and have it fail immediately if the --checkpoint-prefix PREFIX is not valid. 
- Add support for user defined commands in the interactive console, consisting of a named sequence of commands and supporting nested definitions 
- Single-rank, multithreaded interactive debug console support. In multithreaded simulations, interactive console is invoked from the syncManager, starting in thread 0. The user can navigate the object map, set watchpoints, etc. within a thread. The “info” command provides the rank, thread ID, and process ID for threads and the “thread” command switches control between threads. 

----

### DIFF
--- a/src/sst/core/checkpointAction.cc
+++ b/src/sst/core/checkpointAction.cc
@@ -331,6 +331,14 @@ initializeCheckpointInfrastructure(Config* cfg, bool rt_can_ckpt, int myRank)
     std::string checkpoint_dir_name = "";
 
     if ( myRank == 0 ) {
+        // Verify prefix format (no '/')
+        const std::string prefix   = cfg->checkpoint_prefix();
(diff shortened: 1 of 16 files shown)


leekillough:
I expect a lot of merge conflicts with https://github.com/sstsimulator/sst-core/pull/1517. If it is merged soon enough, I'll help resolve the conflicts.

kpgriesser:
Let's postpone this PR until after #1517 is merged. Also will refresh the code with the latest work for multi-threaded debugging.

skuntz:
Replaced by PR #1535

PR #1510, opened 2025-12-17
Throw exceptions for unsupported mapping mode serializations

leekillough:
~~Throw exceptions for unsupported mapping mode serializations~~

Use a reference in timeVortex to avoid a copy

Simplify a `std::vector` declaration

### DIFF
--- a/src/sst/core/objectSerialization.h
+++ b/src/sst/core/objectSerialization.h
@@ -29,8 +29,7 @@ serialize(dataType& data)
 
     size_t size = ser.size();
 
-    std::vector<char> buffer;
-    buffer.resize(size);
+    std::vector<char> buffer(size);
 
     ser.start_packing(buffer.data(), size);
     SST_SER(data);
--- a/src/sst/core/timeVortex.cc
+++ b/src/sst/core/timeVortex.cc
@@ -24,7 +24,7 @@ namespace TV::pvt {
 void
 pack_timevortex(TimeVortex*& s, SST::Core::Serialization::serializer& ser)
 {
-    std::string type = Simulation_impl::getSimulation()->timeVortexType;
+    std::string& type = Simulation_impl::getSimulation()->timeVortexType;
     SST_SER(type);
     s->serialize_order(ser);
 }


feldergast:
The intent for mapping mode is that things that don't support mapping would just silently not map anything.  We don't want exceptions during mapping.  This model allows you to serialize things for checkpointing that don't work with mapping, but not have to check modes just for the purpose of not serializing something in mapping mode.

leekillough:
I am not adding any extra serialization mode tests which were not already there (there was already a switch case or an in-else which tested the mode; I'm simply making it croak if you try to map something which is not mappable).

If mapping mode is in effect and you try to serialize an unsupported type, I would want a hard error instead of just silently continuing. It should reduce the time spent debugging problems caused by mapping mode being selected where it is not supported and it just silently continues, causing other problems to be detected later.

The change to use a reference instead of string copy should be used for two reasons: mapping mode (preventing dangling pointers to local variables), and faster speed (no allocation and copying of a string).

kpgriesser:
Perhaps we could consider a runtime option to enable these fatal exceptions for testing purposes

leekillough:
I have removed the `throw` statements for the unhandled `serializer::MAP` `switch` cases.

One of them, in `link.cc`, threw an exception in mapping mode which was caught but not handled well, causing the code to behave badly, so somehow it needs to work in mapping mode even if Link mapping does nothing. The old code says:
```c++
    } break;
    case serializer::MAP:
        // TODO: Implement Link mapping mode
        break;
    }
```
So Link mapping mode needs to be implemented, but I don't have the time to do it.

The rest of this PR can be merged as a minor style fix / copy avoidance.

feldergast:
> Perhaps we could consider a runtime option to enable these fatal exceptions for testing purposes

Deciding not to map anything during mapping mode is not an error (by deliberate decision when designing the feature).  Somethings don't have a rational way to present themselves in the ObjectMap hierarchy.  Others don't have anything useful to contribute to debugging. You don't want to have to force the person writing the serialize_order functions to know this and mark them as no_map.

If we did turn on fatal exceptions in mapping mode, pretty much every test that used mapping mode would just fatal.  Catching exceptions isn't particularly useful either, since you can't really recover the part of the ObjectMap that was skipped during mapping because the exception jumped you to the handler code and left a bunch of the serialization code unexecuted.  It's just best to happily continue on, even if nothing was done.

PR #1515, opened 2025-12-19
Add warnings for trivially serializable types which cannot automatically have an ObjectMap created for them

leekillough:
Add warnings for trivially serializable types which cannot automatically have an `ObjectMap` created for them

This is motivated by: https://github.com/tactcomplabs/sst-ext-tests/issues/79

The problem:

For classes/unions/odd scalar types which are trivially serializable but do not define their own serialization methods, there is no compilation error, and the current code is faster for checkpointing, but no `ObjectMap` is created, so the debugger appears defective in not listing certain objects.

If trivial serialization was totally removed, then both checkpointing and mapping would require the user to define their own serialization method. Users who do not care about mapping would need to define serialization methods when they do not need to currently, and users who care about mapping would _still_ need to define serialization methods.

The solution:

If a type is trivially serializable but the user has not defined their own serialization method, and has not defined methods to convert their type to/from `std::string`, a warning message is displayed once for each such type when the verbose level >= 2.

In the debugger, it appears like this:

```
>  sst --verbose=2 --interactive-start=0 dbgsst15.py
Entering interactive mode at time 0
Interactive start at 0
> cd cp0
 SST Core: Warning: Trivially serializable union type SSTDEBUG::DbgSST15::ag_union_t does not automatically have an ObjectMap created for it.
To create an ObjectMap for SSTDEBUG::DbgSST15::ag_union_t, use one of these methods:
1. Define a serialize_impl<SSTDEBUG::DbgSST15::ag_union_t> specialization.
2. Add a constructor taking a std::string and a conversion operator to std::string to SSTDEBUG::DbgSST15::ag_union_t.

 SST Core: Warning: Trivially serializable union type SSTDEBUG::DbgSST15::ag_union_struct_t does not automatically have an ObjectMap created for it.
To create an ObjectMap for SSTDEBUG::DbgSST15::ag_union_struct_t, use one of these methods:
1. Define a serialize_impl<SSTDEBUG::DbgSST15::ag_union_struct_t> specialization.
2. Add a constructor taking a std::string and a conversion operator to std::string to SSTDEBUG::DbgSST15::ag_union_struct_t.
```
A compile-time warning could be generated instead of a runtime warning at `verboseLevel >= 2`, but it would need to use non-standard compiler extensions or clever tricks to produce specific warning messages, since C++17 has no standard way to print compile-time warnings except for [`#warning`](https://en.cppreference.com/w/cpp/preprocessor/error) at the preprocessor level, which cannot handle this problem.

Also, this runtime warning is more helpful than a compile-time warning, since it occurs every time the simulation is run, at exactly the time that the unsupported `ObjectMap` needs to be created but isn't. Also, compile-time warnings are often ignored, since they are often false-positives.

Even though the warnings are printed at runtime, the code to print them is conditionally enabled by `if constexpr`, and there is no performance impact if the warnings are not generated.

Previously: `ObjectMap` was silently not created for types which are trivially serializable but which do not define their own serialization methods, and do not have known ways of converting them to `ObjectMapFundamental`. Variables of these types would simply not appear in the debugger.

After this change: A warning message showing the type and how to fix it to get it added to `ObjectMap`, is printed once for each type when `verboseLevel >= 2`, at the point where `ObjectMap` needs to be created.

If it is not too intrusive, I can lower the `verboseLevel` threshold to `1`. I already only print the warning once for each type, and only when that type is considered for `ObjectMap` creation.

@kpgriesser @skuntz @jleidel

### DIFF
--- a/src/sst/core/serialization/impl/ser_shared_ptr_tracker.cc
+++ b/src/sst/core/serialization/impl/ser_shared_ptr_tracker.cc
@@ -54,7 +54,7 @@ ser_shared_ptr_unpacker::get_shared_ptr_owner(size_t tag)
     }
 
     Output::getDefaultObject().fatal(
-        __LINE__, __FILE__, __func__, 1, "Serialization Error: std::shared_ptr ownership tag is out of order");
+        CALL_INFO, 1, "Serialization Error: std::shared_ptr ownership tag is out of order");
 }
 
 } // namespace SST::Core::Serialization::pvt
--- a/src/sst/core/serialization/impl/serialize_shared_ptr.h
+++ b/src/sst/core/serialization/impl/serialize_shared_ptr.h
@@ -142,7 +142,7 @@ pack_shared_ptr_address(
         const void* parent_addr = parent.get();
 
         if ( !parent_addr )
-            Output::getDefaultObject().fatal(__LINE__, __FILE__, __func__, 1,
+            Output::getDefaultObject().fatal(CALL_INFO, 1,
                 "Serialization Error: Serialized std::%s has a non-null stored pointer, but the parent specified by %s "
                 "has a null stored pointer.\nThere is no way to know how this raw pointer should be serialized.\n",
                 ptr_string<PTR_TEMPLATE>, wrapper_string<PTR_TEMPLATE, PARENT_TYPE>);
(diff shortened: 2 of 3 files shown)


kpgriesser:
Seeing some new warnings on  MacOS
```
Apple clang version 17.0.0 (clang-1700.6.3.2)
Target: arm64-apple-darwin25.2.0
```


```
  In file included from ../../../../src/sst/core/serialization/serialize.h:462:
../../../../src/sst/core/serialization/impl/serialize_trivial.h:61:29: warning: variable 'mode' set but not used [-Wunused-but-set-variable]
   61 |         switch ( const auto mode = ser.mode() ) {
      |  
```

leekillough:
@kpgriesser Those are false positives in the MacOS compiler which say that a variable is not used even though it is used on some `if constexpr` paths. Later compilers do not produce this false positive warning. It can be suppressed by using the `UNUSED()` macro.

feldergast (approved):

PR #1517, opened 2025-12-24
Rebuild ObjectMap at the beginning of every interactive session; add ObjectMap refresh() method

leekillough:
Fixes: https://github.com/tactcomplabs/sst-ext-tests/issues/77

The `ObjectMap` needs to be rebuilt at the beginning of every interactive session, or it will be stale.

Use the `name_stack` which was already a part of `SimpleDebugger` but wasn't used. There was already a comment which described the missing functionality of this PR:
```
    // This is the stack of where we are in the class hierarchy.  This
    // is needed because when we advance time, we'll need to delete
    // any ObjectMap because they could change during execution.
    // After running, this will allow us to recreate the working
    // directory as far as we can.
    std::vector<std::string> name_stack;
```
(I changed it from `std::vector` to `std::deque`.)

At the end of an interactive session, traverse up `obj_`'s directory tree in `save_name_stack()`, saving names in `name_stack`, and then clear `obj_` (I call `obj_->decRefCount()` but I'm not sure this is right). I didn't use `getFullName()`, but simply used `selectParent()` until it was null, adding names to `name_stack`. This also calls the `deactivate()` callbacks.

Make all of the `cmd_*` functions and `dispatch_cmd()` return `bool` instead of `void` so that we can tell whether they succeeded or not.

@kpgriesser 

At the beginning of an interactive session, create the `ObjectMap` and cd to the `name_stack` position in `cd_name_stack()`. As soon as a `cmd_cd` command fails, we stop descending further down.

-----

Add a new `refresh()` method to refresh an `ObjectMap` by re-serializing its data.

This is not used right now, but it may be used in the future if an `ObjectMap` does not reflect changes in the program, such as if a STL container invalidates references to its elements when it is modified in the program.

This `refresh()` method does nothing for Fundamental types.

For `ObjectMapWithChildren`, it runs recursively on each of its children. 

For `ObjectMapContainer`, it re-serializes the children in the `ObjectMapContainer` by re-serializing the whole container, creating new children.

For `ObjectMapArray` (unbounded arrays `T[]` with a runtime `size` parameter), it re-serializes the variable-sized array with the `size` parameter. To prevent compilation errors, `ObjectMapArray` has been moved to `serialize_array.h`, and some parts of `serialize_array.h` have been moved around so that definitions appear before uses.

`ObjectMapBoundedArray` has been removed as unnecessary, since `ObjectMapContainer<T[N]>` works sufficiently.

The `SerializeObjectMap()` has been changed to not need any `name` argument, and to just use a temporary `_proxy_object_` variable which it extracts after the object is mapped inside of an `ObjectMapClass`. `SerializeObjectMap()` has been changed to use a universal/forwarding reference, to allow `array()` and other proxy classes to be used as rvalues.

The mapper's `map_primitive()` and `map_container()` have been consolidated into `map_object()`, since they performed exactly the same operation with the same function signature, and `map_container()` was not used.

I tested the `refresh()` method by disabling the re-creation of the `ObjectMap` and of descending to the `name_stack` at interactive startup, and instead running `refresh()` on the `obj_` which will still be at the same directory position that it was at before the simulation was restarted. This worked to correctly observe the changes in `std::queue` by calling `obj_->refresh()`, which only refreshed the `pwd` that it was at.

### DIFF
--- a/src/sst/core/serialization/impl/serialize_shared_ptr.h
+++ b/src/sst/core/serialization/impl/serialize_shared_ptr.h
@@ -173,7 +173,7 @@ pack_shared_ptr_address(
 // Pack the parent owner of a shared pointer, which is done the first time an ownership tag is seen
 template <class PARENT_TYPE>
 void
-pack_shared_ptr_parent(const std::shared_ptr<PARENT_TYPE>& parent, size_t* size, serializer& ser, ser_opt_t opt)
+pack_shared_ptr_parent(const std::shared_ptr<PARENT_TYPE>& parent, size_t* size, serializer& ser, ser_opt_t UNUSED(opt))
 {
     // OWNER_TYPE is PARENT_TYPE with cv-qualifiers removed so that it can be serialized
     using OWNER_TYPE = std::remove_cv_t<PARENT_TYPE>;
@@ -215,7 +215,8 @@ pack_shared_ptr_parent(const std::shared_ptr<PARENT_TYPE>& parent, size_t* size,
 // Unpack a shared pointer owner, finding or creating a std::shared_ptr owner with a particular tag
 template <class PARENT_TYPE>
 const std::shared_ptr<void>&
-unpack_shared_ptr_owner(size_t tag, std::shared_ptr<PARENT_TYPE>& parent, size_t* size, serializer& ser, ser_opt_t opt)
+unpack_shared_ptr_owner(
+    size_t tag, std::shared_ptr<PARENT_TYPE>& parent, size_t* size, serializer& ser, ser_opt_t UNUSED(opt))
 {
     // OWNER_TYPE is PARENT_TYPE with cv-qualifiers removed so that it can be deserialized
     using OWNER_TYPE = std::remove_cv_t<PARENT_TYPE>;
--- a/src/sst/core/serialization/objectMap.h
+++ b/src/sst/core/serialization/objectMap.h
@@ -1315,8 +1315,8 @@ public:
         return new ObjectMapComparison_impl<T, REF>(name, addr_, op, value);
     }
 
-    ObjectMapComparison* getComparisonVar(
-        const std::string& name, ObjectMapComparison::Op op, const std::string& name2, ObjectMap* var2) const override
+    ObjectMapComparison* getComparisonVar(const std::string& name, ObjectMapComparison::Op UNUSED(op),
+        const std::string& name2, ObjectMap* var2) const override
     {
         // Ensure var2 is fundamental type
         if ( !var2->isFundamental() ) {


feldergast:
The original prototype had this functionality.  I wonder when it got removed.

leekillough:
> The original prototype had this functionality. I wonder when it got removed.

The original mention of `name_stack` in `simpleDebug.h` is in commit 2b5da60f9fd6d8b2666aa84410a497447f3d3a83, presumably the commit of the original prototype, but `name_stack` was not used in `simpleDebug.cc`.

I looked at the entire commit history of `simpleDebug.cc`, but it never used `name_stack` until this PR.

It has only checked if `obj_` is `nullptr` when starting interactive mode, and it creates `obj_` if it was `nullptr`, meaning that only the first interactive session creates the `ObjectMap`. I demonstrated this effect in the bug report above, by starting the interactive mode 1ns before the simulation ends, and the `ObjectMap` is correct and reflects all of the changes that the program made.

I have concerns about watchpoints having dangling pointers if containers are modified during runs. `ObjectMap` and its tree never becomes stale as long as it is alive, because it uses `std::multimap` which never invalidates pointers/references/iterators to non-deleted nodes, but `ObjectMapFundamental` and other leaf nodes may point to dangling pointers if they point to container nodes which are modified during runs. The `ObjectMapContainer` holding the container may need to "refresh" itself before any watchpoints are looked at, because pointers to the containers' nodes might become invalidated if the container is modified during the program running.

leekillough:
> I have concerns about watchpoints having dangling pointers if containers are modified during runs. `ObjectMap` and its tree never becomes stale as long as it is alive, because it uses `std::multimap` which never invalidates pointers/references/iterators to non-deleted nodes, but `ObjectMapFundamental` and other leaf nodes may point to dangling pointers if they point to container nodes which are modified during runs. The `ObjectMapContainer` holding the container may need to "refresh" itself before any watchpoints are looked at, because pointers to the containers' nodes might become invalidated if the container is modified during the program running.

I have started a prototype `refresh()` method which should be able to address these concerns. I tested it on the `std::queue` example test case, and even if `ObjectMap` is not totally rebuilt when the interactive session resumes, `obj_->refresh()` is sufficient to make it correctly display the data modified during the program run.

kpgriesser:
I've done some more testing and the previously failing cases for the break into interactive mode are all now passing. For the sake of time, we could approve this PR to fix the interactive console and work on the watchpoint separately.

Regarding the watchpoints:
```I have concerns about watchpoints having dangling pointers if containers are modified during runs. ```
and
```This refresh() method does nothing for Fundamental types```

With the code today, can we assert that fundamental types not within a std::container class are `watchable`? 
- If no, can we conceive a test case that would break it? Also, we have to discuss performance implications of refreshing the object map every sample.
- If yes, could we consider restricting watchpoints to non-container types for faster performance. That is, the user would get a message indicating either that the type cannot be watched... or, warn that this watchpoint could have a bigger impact on performance.

Again, I would support merging this PR without regard to watchpoints. Alternatively, add the refresh for the watchpoints and we can evaluate the performance and decide how to optimize or restrict usage if possible.

**Comment by leekillough** [comment]
> With the code today, can we assert that fundamental types not within a std::container class are watchable?

Yes, as long as they do not go out of scope or get reallocated to a new address, etc. In `ObjectMap` parlance, "Fundamental" types represented by `ObjectMapFundamental` are types which do not have any `ObjectMap` children.

`refresh()` is a `virtual` method which exists in the `ObjectMap` base class as a function which does nothing but simply return.

As I said above, `refresh()` refreshes each of its children in `ObjectMapWithChildren`, and it rebuilds the entire `ObjectMap` for `ObjectMapContainer` and its children. For other `ObjectMap`-derived classes such as `ObjectMapFundamental`, it does nothing. This is using the C++ Virtual Function Table (VTable) method of programming, rather than specific runtime tests for what kind of `ObjectMap` something is.

> Also, we have to discuss performance implications of refreshing the object map every sample.

I would not be too concerned about performance right now -- we mainly want functionality. There are other places in SST which are better targets if we are worried about performance. Only after a profiling reveals the true hotspots, would we know exactly where to go first.

> Again, I would support merging this PR without regard to watchpoints. Alternatively, add the refresh for the watchpoints and we can evaluate the performance and decide how to optimize or restrict usage if possible.

I have tested `refresh()` with the `std::queue` example. But my original fix of, in a sense, "refreshing" the entire `ObjectMap` tree at the beginning of each interactive console, is still important, and this is faster than running `refresh()` at all different levels of the tree as we `cd` into different levels (I hope later that full pathnames with `..` and `/` are supported with `cd`, `ls` etc.).

I have added `refresh()` and tested it, so this PR should go in first, and then you can try to deal with the watchpoint issue. Without scanning (`refresh()`ing) the entire STL container, there's no way to tell whether it was changed by the program. There's no "dirty bit" or flag which says "I've been modified". There's also no guarantee that a STL container's elements won't change address during a program run.

**Comment by leekillough** [comment]
More notes:

Any C++ STL container (like `std::vector`) can invalidate iterators/references/pointers to its elements if it is modified by the program.

If watchpoints are set, then every `ObjectMapContainer` class from the top of the `ObjectMap` hierarchy to the object being watched, must be refreshed. Calling `refresh()` on the top of the hierarchy can accomplish this, because `ObjectMapWithChildren::refresh()` calls `refresh()` on each of its children, and `ObjectMapContainer` (and `ObjectMapArray`) have specific `refresh()` methods which update their children according to the state of the container (or array).

As a performance optimization, rather than refreshing the entire `ObjectMap` from the root, only the `ObjectMapContainer` objects along the path from the root to the watched object need to be refreshed. This prevents refreshing subtrees which are not along the path from the root to the watched object. There is an `isContainer()` method which tells whether a general `ObjectMap` is an `ObjectMapContainer` (or derived thereof). `ObjectMapWithChildren` and `ObjectMapClass` objects which are not `ObjectMapContainer` do not have this flag set, so they do not need to be refreshed.

To work in general, during checking for changed values, watchpoints must descend the full path name from the root to the watched object, because it's possible that some parts of the path name have changed or gone away, and there might be `ObjectMapContainer` objects along the path from the root to the watched object, and they will need to be refreshed before they can be descended into. 

Pseudocode:
```c++
node = component_root;
if ( node->isContainer() ) node->refresh();  // Refresh, but only if it's a container
for (const auto& name : name_stack) {
   node = node->findVariable(name);
   if ( !node ) break;  // abort watching, since the watched variable's path name is no longer valid
   if ( node->isContainer() ) node->refresh();  // Refresh, but only if it's a container
}
```
Here, `name_stack` is similar to `name_stack` in `simpleDebug.cc`, but other methods of storing/retrieving and navigating path names is possible.

PR #1518, opened 2025-12-29
Add clang-tidy script; make the least-impact changes found by clang-tidy

leekillough:
This adds a `clang-tidy.sh` script which runs `clang-tidy` with a modest default set of checks, a more aggressive `--picky` set of checks, and a `--checks` option for specifying specific tests.

`--fix` can be used to invoke the `clang-tidy` automatic fixer, and it modifies files in the working tree similar to how `test-includes.pl --fix` automatically adds `#include`s. 

This `clang-tidy.sh` is not intended for CI runs, but for periodic manual runs to detect potential issues and to modernize the SST code to modern C++ idioms.

I recommend running it with no arguments at first, inspecting the output file, and if one of the suggestions is acceptable, running it with the `--fix --checks <check_name>` with the check name listed in brackets near the suggestion ( `[<check_name>]`), running tests, and merging the changes if they are acceptable. Some of the automatic fixes might require massaging the code after it is edited, because it is not always perfect.

I found one `va_arg` leak (`va_end()` was not called), and I found a reversed name string comparison. I also saw an `if ( strcmp(...) )` which `clang-tidy` warned about because the `int` result was implicitly converted to `bool`, but I only added `!= 0` since I am not sure whether the comparison was reversed or not.

A self-assignment check had to be added to `decimal_fixedpoint`. `clang-tidy` detected that self-assignment wasn't accounted for.

There were warnings about `strcpy()` being insecure, but in two such cases the `strcpy()` was preceded by a `malloc()` or `new[]` and both the allocation and `strcpy()` could be replaced with `strdup()`.

I will add comments below each of these changes.

The changes are minimal, to avoid disruptions, but the script is robust enough to be used in the future for `clang-tidy` research into possible issues, as well as ideas for modernizing the code.

### DIFF
--- a/scripts/clang-tidy.sh
+++ b/scripts/clang-tidy.sh
@@ -7,6 +7,9 @@ if ! root=$(git rev-parse --show-toplevel); then
     exit 2
 fi
 
+[[ -n ${MAKEFLAGS:-} ]] && echo "Warning: MAKEFLAGS is ignored" >&2
+unset MAKEFLAGS  # We do not want parallel builds when running bear
+
 cd "${root}"
 
 rebuild=0
@@ -58,8 +61,9 @@ EOF
         exit 2
     fi
     cd build
-    rm -rf clang-tidy.out compile_commands.json src
+    rm -rf clang-tidy.out compile_commands.json
     make -j clean
+    rm -rf src
     bear -- make install
     cd ..
 fi


src/sst/core/configGraphOutput.cc
```diff
@@ -17,15 +17,12 @@
...
 ConfigGraphOutput::ConfigGraphOutput(const char* path)
 {
     SST::Util::Filesystem filesystem = Simulation_impl::filesystem;
     outputFile                       = filesystem.fopen(path, "wt");
 }
 
-
-} // namespace Core
-} // namespace SST
+} // namespace SST::Core
```
leekillough: There were still a few places where nested namespace names were not used. `clang-tidy` was able to fix these automatically, although it sometimes left the old namespace closing comment around, so I had to manually remove them.

src/sst/core/elemLoader.cc
```diff
@@ -110,15 +110,13 @@ static std::vector<std::string>
...
-    char* p    = nullptr;
+    char*                    pathCopy = strdup(searchPaths.c_str());
+    char*                    brkb     = nullptr;
+    char*                    p        = nullptr;
     for ( p = strtok_r(pathCopy, ":", &brkb); p; p = strtok_r(nullptr, ":", &brkb) ) {
         paths.push_back(p);
     }
-
-    delete[] pathCopy;
+    free(pathCopy);
```
leekillough: `clang-tidy` complained about `strcpy()` being inherently insecure. Since here we are simply allocating and copying a string, `strdup()` is simpler, and removes the warning.

src/sst/core/impl/partitioners/simplepart.cc
```diff
@@ -16,9 +16,9 @@
 #include "sst/core/model/configGraph.h"
 #include "sst/core/warnmacros.h"
 
+#include <cstdlib>
```
leekillough: There are numerous places where C-style headers were replaced with C++ headers ( `<stdlib.h>` -> `<cstdlib>` etc.).

src/sst/core/serialization/serializable_base.h
```diff
@@ -208,9 +208,9 @@ public:
     {                                                                                     \
         return SST::Core::Serialization::serializable_builder_impl<obj>::static_cls_id(); \
     }                                                                                     \
-    static obj* construct_deserialize_stub()                                              \
+    static obj* construct_deserialize_stub() /* NOLINT(bugprone-macro-parentheses) */     \
     {                                                                                     \
-        return new obj;                                                                   \
+        return new obj; /* NOLINT(bugprone-macro-parentheses) */                          \
```
leekillough: I got `clang-tidy` warnings about macro arguments requiring parentheses, but with inline comments, I suppressed most, if not all, of them. They cannot be used here because they change the meaning of the code.

src/sst/core/simulation_impl.h
```diff
@@ -79,11 +79,6 @@ class StatisticOutput;
 class StatisticProcessingEngine;
 } // namespace Statistics
 
-namespace Serialization {
-class ObjectMap;
-} // namespace Serialization
-
-
```
leekillough: `ObjectMap` is declared in `SST::Serialization` here instead of `SST::Core::Serialization`, so it's an unused declaration which `clang-tidy` warned about.

src/sst/core/sst_types.h
```diff
@@ -43,13 +43,13 @@ static constexpr StatisticId_t STATALL_ID = std::numeric_limits<StatisticId_t>::
...
-#define SUBCOMPONENT_ID_CREATE(compId, sCompId) ((((uint64_t)sCompId) << COMPONENT_ID_BITS) | compId)
+#define SUBCOMPONENT_ID_CREATE(compId, sCompId) ((((uint64_t)(sCompId)) << COMPONENT_ID_BITS) | (compId))
 #define CONFIG_COMPONENT_ID_BITS                (COMPONENT_ID_BITS + SUBCOMPONENT_ID_BITS)
 #define CONFIG_COMPONENT_ID_MASK(x)             ((x) & 0xFFFFFFFFFFFFULL)
-#define STATISTIC_ID_CREATE(compId, statId)     ((((uint64_t)statId) << CONFIG_COMPONENT_ID_BITS) | compId)
+#define STATISTIC_ID_CREATE(compId, statId)     ((((uint64_t)(statId)) << CONFIG_COMPONENT_ID_BITS) | (compId))
 #define COMPDEFINED_SUBCOMPONENT_ID_MASK(x)     ((x) >> 63)
 #define COMPDEFINED_SUBCOMPONENT_ID_CREATE(compId, sCompId) \
-    ((((uint64_t)sCompId) << COMPONENT_ID_BITS) | compId | 0x8000000000000000ULL)
+    ((((uint64_t)(sCompId)) << COMPONENT_ID_BITS) | (compId) | 0x8000000000000000ULL)
```
leekillough: Here is one place where macro arguments can be parenthesized safely to get rid of `clang-tidy` warnings.

scripts/clang-tidy.sh
```diff
@@ -0,0 +1,96 @@
...
+
+cd build
+output=$(realpath clang-tidy.out)
+
+echo "Using clang-tidy --checks=${clang_tidy_checks}"
+
+perl -ne 'print "$1\0" if /"file":\s*"([^"]+)"/ && !m:/external/:' compile_commands.json | \
+    xargs -0 -P "${njobs}" -I {} timeout 30 sh -c 'echo "Checking {}"; clang-tidy '"${clang_tidy_fix}"' --quiet --format-style=file -p . --system-headers=0 --checks="'"${clang_tidy_checks}"'" "{}"' | tee "${output}" || true
+
+echo "Results are in ${output}"
```
leekillough: The `clang-tidy.sh` script has `clang-tidy` checks optimized for SST Core, by not needlessly listing too many warnings which are false positives, but allowing `--picky` to increase the number of warnings, while still being informative and not too much noise.

leekillough:
The biggest changes that `clang-tidy` found, but which were not included in this PR, are:

1. Use range-based `for` loops (`for (auto& i : container)` instead of `for (auto it = container.begin(); it != container.end(); ++it)`)
2. Make trivial constructors and destructors use `= default` instead of empty bodies `{ }`.
3. Pass arguments like `std::string` by value, and then `std::move()` them, instead of passing them by `const` reference and then copying them. This allows rvalues to be _moved_ instead of _copied_ when passed to functions, avoiding unnecessary memory allocations and memory copies.
4. Pass non-copied read-only arguments by `const` reference instead of by value (the reverse of `#3`).
5. Use references instead of copied values for range-based `for` loop variables or local variables.
6. Use `return {...}` brace syntax instead of `return type(...)` where `type` is the return type. Unless `type`'s constructor is `explicit`, braces `{ }` can be used to construct a `type` return value with list initialization, without repeating `type`.
7. Use C++17 `std::scoped_lock` instead of `std::lock_guard` (and sometimes `std::unique_lock`). With class template argument deduction, it is as simple as `std::scoped_lock lock(mutex);` or `std::scoped_lock lock(mutex1, mutex2, mutex3);`
8. Use faster `find()` methods using character literals instead of length-1 strings

Most of these can be automatically fixed with `--fix`. Most of them create much more code changes and are at a higher risk of breaking code if they are not done exactly right, so they were left out of this PR. Or they are purely cosmetic, with no performance or correctness impact.

src/sst/core/sstregistertool.cc
```diff
@@ -129,10 +129,12 @@ sstRegister(char* argv[])
...
     FILE* cfgFile = fopen(cfgPath, "r+");
-    populateEnvironmentConfig(cfgFile, database, true);
+    if ( cfgFile ) {
+        populateEnvironmentConfig(cfgFile, database, true);
 
-    database->getGroupByName(groupName)->setValue(key, value);
-    fclose(cfgFile);
+        database->getGroupByName(groupName)->setValue(key, value);
+        fclose(cfgFile);
+    }
```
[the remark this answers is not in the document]
kpgriesser: Double check line 131 and 139 for use of cfgFile ( two fopens )

scripts/clang-tidy.sh
```diff
@@ -0,0 +1,97 @@
...
+    clang_tidy_checks="bugprone-*,performance-*,modernize-*,readability-redundant-*,-modernize-use-trailing-return-type,-modernize-avoid-c-arrays,-modernize-use-override,-modernize-use-nodiscard,-modernize-macro-to-enum,-performance-enum-size,-modernize-use-auto,-bugprone-easily-swappable-parameters,-bugprone-reserved-identifier,google-explicit-constructor,misc-static-assert,misc-header-include-cycle,misc-definitions-in-headers,-modernize-return-braced-init-list,-clang-analyzer-cplusplus.NewDeleteLeaks"
+
+    if [[ ${picky} -eq 0 ]]; then
+        # This disables more checks when --picky is not used.
+        clang_tidy_checks="${clang_tidy_checks},-readability-redundant-inline-specifier,-modernize-use-equals-default,-performance-avoid-endl,-modernize-pass-by-value,-modernize-use-default-member-init,-readability-redundant-string-init,-modernize-use-equals-delete,-performance-unnecessary-copy-initialization,-readability-redundant-access-specifiers,-modernize-use-bool-literals,-readability-redundant-member-init,-performance-unnecessary-value-param,-bugprone-narrowing-conversions"
+    fi
+fi
```
berquist: Is there any way we can get these into configuration files?  It would make it convenient for people who do have clang-tidy set up in their editor to be getting the same reports as from running the script.

scripts/clang-tidy.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+if ! root=$(git rev-parse --show-toplevel); then
+    echo "$0 needs to be run in a SST Core Git working tree" >&2
```
berquist: Consider
```suggestion
    echo "$0 needs to be run in a SST Git working tree" >&2
```
as well as the later line, since we may also want to run this on elements eventually, and there's nothing here that functionally prevents doing so.

berquist (changes requested):
This review is for the script only, and not the actual clang-tidy checks used.

scripts/clang-tidy.sh
```diff
@@ -0,0 +1,97 @@
...
+https://github.com/rizsotto/Bear
+EOF
+        exit 2
+    fi
+    if [[ ! -e build/config.status ]]; then
+        echo "The build subdirectory needs to be created, autogen.sh run, and configure run." >&2
+        exit 2
+    fi
+    cd build
+    rm -rf clang-tidy.out compile_commands.json src
```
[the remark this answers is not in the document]
leekillough: `rm` with the `-f` (force) flag should not give errors on missing files, even on BSD or other POSIX variants. So removing `-r src` and adding `|| true` are unnecessary.

The error shown above is in a `cd src` command, not `rm -rf src`. Maybe `make clean` runs `cd src`. In that case, `rm -rf src` should come _after_ `make -j clean`.

scripts/clang-tidy.sh
```diff
@@ -0,0 +1,97 @@
...
+    make -j clean
```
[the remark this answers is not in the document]
leekillough: I did not know whether `bear` would work safely with parallel builds, so I left out `-j` for the `bear make` build, but I added it for the `make -j clean`. I would not rely on environment variables for _force_ `make` to build everything in parallel, because then some tools might not work.

It finishes the build in around 30 minutes without parallelism, which is not too slow for the purpose of this script.

scripts/clang-tidy.sh
leekillough: There is a `--checks` flag to enable specific checks. If specific `--checks` are specified, I have to disable the clang-analyze checks (or whatever they are called) because they are enabled by default. But one of them is required for one of the other checks, so I did not disable it. It causes false positive warnings about code which is never executed.

feldergast (approved):

PR #1520, opened 2025-12-30
Modernize loops to use C++11 range-based-for

leekillough:

This was generated with `scripts/clang-tidy.sh --fix --checks modernize-loop-convert`


feldergast:
simulation.cc:581:26: error: non-const lvalue reference to type 'unsigned long' cannot bind to a value of unrelated type 'unsigned long long'
11:41:47     for ( unsigned long& interThreadLatencie : interThreadLatencies ) {

Looks like this needs to be SimTime_t instead of unsigned long

feldergast:
Closing due to the codebase moving too far out of sync

PR #1523, opened 2025-12-31
Usually do not use std::endl since it is slower than \n

leekillough:

This was generated by `scripts/clang-tidy.sh --checks performance-avoid-endl --fix` with some massaging.

The interactive console code is exempted from this, since it requires `std::endl` to flush output.


feldergast:
None of the changes I saw are in a timing sensitive path and many of them are used in error messages where we want everything to be flushed along the way.  Going to close this one.

PR #1527, opened 2025-12-31
Pass std::string and RankInfo by value and use std::move() to move it, reducing allocations and copies

leekillough:

This was obtained by running `scripts/clang-tidy.sh --checks modernize-pass-by-value --fix`.


feldergast:
Closing.  None of these are in performance critical paths.

PR #1528, opened 2025-12-31
Use more efficient string concatenation

leekillough:

This gets rid of warnings generated by `scripts/clang-tidy.sh --checks performance-inefficient-string-concatenation`.

`clang-tidy` can detect, but not auto-fix these inefficient string concatenations. I did it by hand, to get rid of the warnings. It is best to append (`+=`) to a `std::string` variable and then `std::move()` it, than it is to create a large string concatenation expression, because it does not need to allocate and deallocate memory for each temporary string expression, and it only needs to append to the same string variable repeatedly.


feldergast:
Closing due to code moving too far out of sync

Issue #1531, opened 2026-01-07
Python upgrade causing some sst-core tests to fail on SLES systems


kpgriesser:
This issue was exposed while updating a SLES system to support a more current version of Python per 
`1b2b6c4e (HEAD, sandia/devel, sles_fix) Enforce minimum Python version of 3.9 (#1384)`. 
 It is NOT an issue with that commit, however. To prove this we went to the previous commit which passes on the older version of Python. 
`2d7a4959 Add warnings for trivially serializable types which cannot automatically have an ObjectMap created for them (#1515)`

The original environment included:
```
export CC=gcc-15
export CXX=g++-15
export PATH=$SST_CORE_HOME/bin:/usr/lib64/mpi/gcc/openmpi4/bin:$PATH
export LD_LIBRARY_PATH=/usr/lib64/mpi/gcc/openmpi4/lib64
export OMPI_CC=gcc-15
export OMPI_CXX=g++-15

$ python --version
Python 3.6.15
```
We configure using 
`../configure --prefix=$SST_CORE_HOME`
, compile, run, and everything passes.

Now after cleaning this build (including removing ./test/__pycache__) and starting a fresh shell we run with this configuration:

`../configure --prefix=$SST_CORE_HOME --with-python=/usr/bin/python3.11-config`

Note that the default Python installation still exists but we are assuming we should be able to override it with the above configuration option.

All the tests pass except for this small subset.
```
=== TEST RESULTS BREAKDOWN ===========================================

testsuite_default_Checkpoint.testcase_Checkpoint
 - FAILED  : test_Checkpoint_SharedObject_array (testsuite_default_Checkpoint.testcase_Checkpoint.test_Checkpoint_SharedObject_array)
 - FAILED  : test_Checkpoint_SharedObject_bool_array (testsuite_default_Checkpoint.testcase_Checkpoint.test_Checkpoint_SharedObject_bool_array)
 - FAILED  : test_Checkpoint_SharedObject_bool_array_n2one (testsuite_default_Checkpoint.testcase_Checkpoint.test_Checkpoint_SharedObject_bool_array_n2one)
 - FAILED  : test_Checkpoint_SharedObject_map (testsuite_default_Checkpoint.testcase_Checkpoint.test_Checkpoint_SharedObject_map)
 - FAILED  : test_Checkpoint_SharedObject_set (testsuite_default_Checkpoint.testcase_Checkpoint.test_Checkpoint_SharedObject_set)
```

[...]

PR #1533, opened 2026-01-08
Remove stale code in PyModel 

kpgriesser:
This is an empirical fix for #1531 which removing code which generates warnings in certain tests when running on SLES systems using Python11.   The fix has been verified across TCL platforms.  Not knowing the history of this code, we're assuming this code is no longer needed.


kpgriesser:
Made the changes requested in #1531 
Tests are passing

PR #1535, opened 2026-01-08
Interactive Console Updates for Multi-thread/Single-Rank Debug Support

kpgriesser:
This PR replaces #1507 to track the interactive console feature branch development.
We decided that it would be more efficient to refer to the live feature branch rather than a snapshot (which has since gotten out of date). 

To recap: This PR provides a set of updates to the interactive debug console, object mapping support, and bug fixes.

- Ensure signal command line flags are processed before the checkpoint directory is created and have it fail immediately if the --checkpoint-prefix PREFIX is not valid.
- Add support for user defined commands in the interactive console, consisting of a named sequence of commands and supporting nested definitions
- Single-rank, multithreaded interactive debug console support. In multithreaded simulations, interactive console is invoked from the syncManager, starting in thread 0. - - The user can navigate the object map, set watchpoints, etc. within a thread. The “info” command provides the rank, thread ID, and process ID for threads and the “thread” command switches control between threads.


kpgriesser:
Replacing this PR with one using a new branch

Issue #1534, opened 2026-01-08
Intermittent sst-core test failures on very fast systems


kpgriesser:
Our newer macOS/clang20+ platforms are significantly faster than our other test platforms and occasionally fail for these two tests:

```
sst-test-core -e test_RealTime_SIGALRM_multiaction
sst-test-core -e test_RealTime_heartbeat
```

While currently this is a minor annoyance it could become a bigger problem as more tests are added that follow the same design pattern.  We haven't found a way to make this repeatable. We could spend time trying to instrument the code to capture a failure but decided it's not a priority at this point.

Maybe Gwen can figure it out through inspection :)

kpgriesser:
I was able to reproduce a failure by running the following script for about 20 minutes:

```
#!/bin/bash -x

while true; do
  sst-test-core -f -w "*_R*" | tee log
  grep -q FAILED log
  if [ $? -eq 0 ]; then
   echo FAILED
   exit 1
  else
   echo PASSED
  fi
done
```

Note:  I'm running suites starting with `*_R*`. I wasn't able to reproduce the failure by running the individual test repeatedly.

The detailed test command is 
`sst  --exit-after=6s --sigalrm=sst.rt.heartbeat(interval=1s);sst.rt.status.core(interval=2s) /Users/dev/kgriesser/work/sst-core/tests/test_RealTime.py`

A side by side diff follows:
```
#  right side passing, left side failing 
$ diff  -y -W 200 passes/sst_test_outputs/run_data/test_RealTime_SIGALRM_multiaction.out failures/sst_test_outputs/run_data/test_RealTime_SIGALRM_multiaction.out > test_RealTime_SIGALRM_multiaction.notes

        CurrentSimCycle:  23810000								   |		CurrentSimCycle:  10619000
	CurrentSimCycle:  46904000								   |		CurrentSimCycle:  19421000
	CurrentSimCycle:  70545000								   |	# Simulation Heartbeat: Simulated Time 5.347 us (Real CPU time since last period 1.00965 seconds
# Simulation Heartbeat: Simulated Time 11.447 us (Real CPU time since last period 1.00478 second   <	
	Max mempool usage:               11.6654 MB								Max mempool usage:               11.6654 MB
	Global mempool usage:            11.6654 MB								Global mempool usage:            11.6654 MB
	Global active activities         5 activities								Global active activities         5 activities
	Max TimeVortex depth:            102 entries								Max TimeVortex depth:            102 entries
[...]

PR #1536, opened 2026-01-12
extend test times for tests counting heartbeats

kpgriesser:
Refer to issue  #1534 for more data.

For some TCL systems we are seeing simulation rates varying from ~3.8us/s to ~14us/s (on the same system). 
These 2 tests have been failing intermittently:
 `test_RealTime_SIGALRM_multiaction`
 `test_RealTime_heartbeat`

They both expect 5 heartbeats in 6 seconds (wall clock). Assuming that the  `--exit-after` option includes the start-up phases, we believe that not enough time is left for execution. By increasing this option from 6s to 30s, we should have sufficient margin to reduce (or eliminate) the failures for the target system.

If, on the other hand, the `--exit-after` is purely timing the execution phase, then there may be a more serious bug that needs attention. 

The original failure took 20 minutes of repeated simulations to reproduce.  So far, ongoing testing with the increased times has not produced any errors since 10:50MST (currently 11:36MST).

### DIFF
--- a/tests/test_RealTime.py
+++ b/tests/test_RealTime.py
@@ -1,9 +1,10 @@
 import sst
 import sys
 
-# Note: test WILL NOT terminate on its own
+# PR #1536 All tests not exiting on their own should use --exit-after
 if len(sys.argv) == 2 and sys.argv[1] == "stop":
-    sst.setProgramOption("stop-at", "25us")
+    print("WARNING[test_RealTime.py:6] ignoring 'stop' argument");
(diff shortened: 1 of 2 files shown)


kpgriesser:
'test_RealTime_SIGALRM_multiaction' fails repeatedly in the CI testing but I don't know how to debug it on whatever system it failed on. So I pushed a (hopefully) less intrusive change that backs off the timeout to 15s instead of 30.   Perhaps there is some other nascent nefarious problem being exposed here

PR #1537, opened 2026-01-12
TimeConverter updates

feldergast:
Allow TimeConverters to be directly instantiated and simplify APIs that use timebases by allowing implicit conversion of string and UnitAlgebra to TimeConverter instead of having multiple copies of the same function with different parameters.  Also cleaned up the basecomponent and timelord source files.

### DIFF
--- a/src/sst/core/baseComponent.h
+++ b/src/sst/core/baseComponent.h
@@ -229,6 +229,8 @@ public:
 
     /**
        Return the simulated time since the simulation began in the default timebase
+
+       @return Current simulation time in units of the default timebase for the object
     */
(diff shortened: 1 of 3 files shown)


src/sst/core/baseComponent.h
```diff
@@ -438,8 +577,7 @@ class BaseComponent : public SST::Core::Serialization::serializable_base
 
 
     /**
-       Handles the profile points, default time base, handler tracking
-       and checkpointing.
+       Handles the profile points, default time base, handler tracking and checkpointing.
```
gvoskuilen: `@param`

gvoskuilen (changes requested):

PR #1538, opened 2026-01-13
Provide CTRL-D functionality to quit if used when line is empty

kpgriesser:
This fixes #32 

We should now support the following behavior with CTRL-D

```
       end-of-file (usually C-d)
              The  character indicating end-of-file as set, for example, by ``stty''.  If this character is
              read when there are no characters on the line, and point is at the  beginning  of  the  line,
              Readline interprets it as the end of input and returns EOF.
       delete-char (C-d)
              Delete  the  character  at point.  If this function is bound to the same character as the tty
              EOF character, as C-d commonly is, see above for the effects.
```

@skuntz I don't have a way to add a test for this. Appreciate if you can kick the tires on it.

As an aside, I'm wondering if we want to add another  prompt for 'are you sure you want to quit' or 'are you sure you want to shutdown'. That's a separate issue though

### DIFF
--- a/src/sst/core/impl/interactive/simpleDebug.cc
+++ b/src/sst/core/impl/interactive/simpleDebug.cc
@@ -776,7 +776,7 @@ SimpleDebugger::cmd_print(std::vector<std::string>& tokens)
     }
 
     // See if have a -r or not
-    int         recurse = 0;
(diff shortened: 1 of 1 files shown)


kpgriesser:
Did not intend to create this in this repository. closing

Issue #1540, opened 2026-01-13
Changing event delivery sort criteria based on the link the event was sent on


feldergast:
**Please describe the new features with any relevant use cases.**

The order of Event delivery depends on 4 criteria:

1. Delivery time
2. Delivery priority (defaults to 50)
3. An order tag (described below)
4. Order of insertion into the queue (first in, first out if all other criteria match)

Currently, for links connected in the input file, the order tag is based on the alphabetized link names; alphabetically earlier link names are delivered before later ones.  This means that Event delivery was dependent on user provided information.  It also required the core to sort all the link names. The order tag for SelfLinks is set to std::numeric_limits<uint32_t>::max() for all SelfLinks.  This means that Events with identical delivery time and priority are delivered in the order they were inserted into the queue.

We are planning to change the order tag to be based on the order that configureLink() is called in the Component tree.  This has two advantages.  First, ordering of Events with identical delivery time and priority within a Component will now be under control of the model writer and will be consistent no matter how the links are named.  Second, we no longer need to sort the list of link names to assign the order tag, which can save a lot of execution time for simulations with large numbers of links.  SelfLinks will continue to all be given the same order tag, so delivery order will still depend on insertion order.

The side effect of these changes will be that the output of a given simulation may be different between the sorting order used today and the new sorting order.

**Describe the proposed solution you plan to implement**

A new function will be added to BaseComponent to get the order tag for a link when configureLink() is called.  The top level Component for the Component tree will have a counter that is incremented each time a new order is needed.  Additionally, the sorting of the Link names will be removed.

**Testing plan**
Testing will be done as it is today, but it is likely that a number of test reference files will need to be changed along with the code changes.

PR #1549, opened 2026-01-26
Event order update

feldergast:
Changing how the event ordering based on the links they were sent on.  Currently it is based on the alphabetized link name, this PR moves control of ordering to the model developer by basing it on the order in which configureLink() is called.  This PR leaves ordering for self links the same (they are all given the same order tag, so events with the same deliver time and priority will deliver in the order they were sent).

This is the implementation for issue #1540 

NOTE: Since this PR changes some of the outputs of test reference files, we will need to force push the reference file update on Elements and then send this PR for testing.


kpgriesser:
FYI - checked that this branch is passing on TCL CI pipeline as well

PR #1552, opened 2026-01-30
Add support for paging of output and max line length in interactive mode

donofrio:
This addresses Issue https://github.com/tactcomplabs/sst-core/issues/42 (and a little bit of https://github.com/tactcomplabs/sst-core/issues/44 ) by adding functionality for the debugger to pause long running (multiple screens, or long lines) outputs. Currently if the output of the ls command goes beyond 50 lines the user is greeted with the message:

`--Type <RET> for more, q to quit, c to continue without paging--`

In addition, any lines longer than 160 characters are truncated with ...

This is accomplished by adding a new stream class `DebugStream` - declared as a new member named `dout` in `SimpleDebugger` It is used as a drop in replacement for `std::cout` and has an overloaded `<<` operator. The key difference is you must call `dout << dreset` after each use to reset the stream state. No additional buffers are needed. `DebugStream` may implement other stream modifiers that are passed in as function pointers.

Along with this PR is a slight change to the CaptCrunch_Interactive test - an independent PR for that is also posted in the sst-ext-tests repo.

### DIFF
--- a/src/sst/core/impl/interactive/Makefile.inc
+++ b/src/sst/core/impl/interactive/Makefile.inc
@@ -9,5 +9,4 @@ sst_core_sources += \
 	impl/interactive/cmdLineEditor.cc \
 	impl/interactive/cmdLineEditor.h \
 	impl/interactive/debugStream.h  \
-	impl/interactive/debugStream.cc 
-	
+	impl/interactive/debugStream.cc


kpgriesser:
@gvoskuilen When you get a chance, could you give this a try with your memhiearchy debug case and let us know what you think?

kpgriesser:
Is there anything blocking this merge? We have test changes behind it.

feldergast:
I thought autodetection of screen size was going to be added.  If that's not the case, I can send it through testing.

kpgriesser:
Now that you mention it, we had a few items in our meeting notes which I should have captured here.
- add boilerplate
- add comment that this feature is only used in the context of the interactive console
- Look at sst output formatted to get the width of the console lines.

Ok, let's hold off on this until we close these

donofrio:
Updated with requested bullets above

PR #1560, opened 2026-02-17
Fix the bad MPI return codes

jmlapre:
Sometimes MPI commands fail with odd error codes.  If an error is expected, we should treat it as a zero or non-zero return code.


src/sst/core/testingframework/sst_unittest.py
```diff
@@ -368,6 +368,8 @@ def run_sst(
         err_str = "SST Timed-Out ({0} secs) while running {1}".format(timeout_sec, oscmd)
         self.assertFalse(rtn.timeout(), err_str)
         if expected_rc:
+            # convert it to a boolean like C: !!rtn
+            rtn._run_status = not not rtn._run_status
```
gvoskuilen: What about the assertion on line 374? It will still compare `rtn._run_status` (bool) and `expected_rc` (int) for equality which will fail if `expected_rc != 1`.

gvoskuilen (approved):

PR #1564, opened 2026-03-02
Update and fix JSON parsing to pass core tests when translated to JSON

gvoskuilen:
This updates the JSON streaming parser to (a) fix some known bugs and (b) pass most of the SST core testsuite including support for statistic enables, port modules, arbitrarily nested subcomponents, etc. 

Specific updates include:
* Moved definition of `STATALLFLAG` to sstmodel.h since it is used only for configuration
* Added `statistics_options` sections to components, subcomponents, portmodules to allow global statistic configuration for those elements without needing to explicitly list statistics
* JSON output previously stopped after 2 levels of nested subcomponents; it now supports arbitrary nesting
* States in streaming parser are stored as a single state var rather than a combo of a state + bools
* Port modules are supported in both JSON output and input
* Added snakecase versions of camelcase keywords so that all keywords in json can be specified in snakecase (usability improvement). Did not remove camelcase versions.
* Added more error checking to JSON parsing, particularly for unexpected tokens.

### DIFF
--- a/src/sst/core/model/json/jsonmodel.cc
+++ b/src/sst/core/model/json/jsonmodel.cc
@@ -32,7 +32,7 @@ SSTConfigSaxHandler::findComponentIdByName(const std::string& name, bool& succes
 
     if ( name.length() == 0 ) {
         error_str_ = "Error: Name given to findComponentIdByName is null";
-        success  = false;
+        success    = false;
         return id;
     }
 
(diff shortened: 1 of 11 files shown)


src/sst/core/model/json/jsonmodel.cc
```diff
@@ -22,597 +22,1174 @@ DISABLE_WARN_STRICT_ALIASING
 
 using namespace SST;
 using namespace SST::Core;
+using namespace SST::Core;
```
feldergast: Duplicate using statement

feldergast (dismissed):

feldergast (approved):

PR #1573, opened 2026-03-18
Add `categorize` decorator for controlling when tests run

### LINKED ISSUE #1541: Nightly tests don't run for release test jobs
5d4426272 (#1290) introduced functionality for reducing the testing time in PRs with the Autotester by selectively setting tests to only run during nightly testing.

https://github.com/sstsimulator/sst-core/blob/20f93198bdca337d02d2117fc6ad102d2db1e6bf/src/sst/core/testingframework/sst_unittest_support.py#L82-L87

An example of its usage is https://github.com/sstsimulator/sst-elements/blob/3e4b9308866aeab46c3a79d0b3d48bc042fa3b44/src/sst/elements/prospero/tests/testsuite_default_prospero.py#L56:

```python
    @unittest.skipIf(not testing_check_is_nightly(), "test_prospero_text_using_PIN_traces only runs on Nightly builds.")
    def test_prospero_text_using_PIN_traces(self):
        ...
```

The problem with this is that we want release tests (tests are scheduled to run weekly on the versions tagged for release) to also run all of these tests.  `BUILD_TAG` is one of the [environment variables Jenkins automatically creates for every job](https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables) and looks like `jenkins-SST__15.1.0_beta_NewFW_Ubuntu-22.04_OMPI-4.1.4_PY3.10_Mainline_MR-2-75` for a release job.

None of the core tests are tagged with this, only elements, but because this involves the test framework and will require changes in this repo, the issue is created here.

This functionality will be reworked from an environment variable to a command-line argument so that it is clearer on how to control which tests are chosen or excluded[^1].  Rather than the binary choice of "skip unless nightly", the logic should be changed to control which subset of tests are run.  We think there are three current subsets which are not necessarily disjoint:
- tests that should run for PRs
- tests that should run nightly and for releases
- tests that should run weekly

We should investigate if [scenarios](https://github.com/sstsimulator/sst-core/tree/20f93198bdca337d02d2117fc6ad102d2db1e6bf/src/sst/core/testingframework#test-scenarios) can be used.  It looks like specifying a scenario means that test is skipped, but we are looking for something closer to [pytest markers](https://docs.pytest.org/en/stable/example/markers.html#registering-markers).

In the meantime, I will see if it is possible to override `BUILD_TAG` in the Jenkins shell script section to trigger release jobs to run the nightly tests.

[^1]: `BUILD_TAG` may still be used inside of bamboo in order to determine what the testing command-line is.

berquist:
Closes #1541


berquist:
This will remain draft until updates to SQE are made and confirmed working.  This PR isn't strictly dependent on those changes: testing wouldn't be broken, but nightly tests would be unintentionally skipped.

Since the nightly tests are no longer derived from an automatically-set environment variable in Jenkins, the `--categorize nightly` needs to be passed to `sst-test-{core,elements}`.  In order to not change the `bamboo.sh` command line, it will probably need to read from `BUILD_TAG`, which is how the test framework currently implements the nightly check.

berquist:
> This will remain draft until updates to SQE are made and confirmed working. This PR isn't strictly dependent on those changes: testing wouldn't be broken, but nightly tests would be unintentionally skipped.

I see bamboo reads from `SST_TEST_FRAMEWORKS_PARAMS`, so I'll try that, and only changes to Jenkins instead of bamboo might be necessary.

berquist:
I've confirmed via AnyBranch that:
- With this PR, `export SST_TEST_FRAMEWORKS_PARAMS="--categories nightly"` does run all the nightly tests, without the decorator, because the logic of `testing_check_is_nightly` was updated to use the same category logic as the decorator.
- Without this PR, adding that shell export causes an unrecognized argument error from argparse.  This means that the nightly Jenkins jobs can't be updated for these changes until after the PR is merged.

src/sst/core/testingframework/sst_unittest_support.py
```diff
@@ -81,10 +85,7 @@ def __str__(self) -> str:
 
 def testing_check_is_nightly() -> bool:
     """If Nightly is in the BUILD_TAG it's very likely a Nightly build."""
```
gvoskuilen: Comment isn't accurate anymore, remove.

gvoskuilen (approved):

PR #1579, opened 2026-03-26
Remove deprecated functions using TimeConverter* and rework internals…

feldergast:
… to only use TimeConverter


jmlapre (approved):
Everything looks pretty good to me.

gvoskuilen (approved):

PR #1583, opened 2026-04-01
Fix Python 3.14 compatibility: segfaults in GC and module loader

deanchester:
## Summary

SST core crashes with SIGSEGV on Python 3.14 due to three independent bugs in the Python/C API integration. All 10 Merlin tests now pass with Python 3.14.2.

### 1. `PyGC_Collect()` crash during shutdown (Pattern 2)

Python 3.14 rewrote the cycle GC to be incremental (two generations instead of three). Calling `PyGC_Collect()` during shutdown triggers a segfault because `type_clear` may have already cleared type MROs and internal references, causing the collector to dereference freed memory. See CPython issues [#130380](https://github.com/python/cpython/issues/130380), [#133932](https://github.com/python/cpython/issues/133932), [#135115](https://github.com/python/cpython/issues/135115).

**Fix:** Skip the explicit `PyGC_Collect()` call on Python >= 3.14. `Py_Finalize()` already calls it internally.

### 2. `mlExecModule()` returns borrowed reference as new reference (Pattern 1)

`mlExecModule()` returned `args` (a borrowed reference) instead of `Py_None`. The C API contract requires returned values to be new references. This caused a refcount underflow leading to use-after-free, manifesting as SIGSEGV in `BaseException_clear` during error display after any exception when SST element modules (e.g., merlin) were loaded.

**Fix:** Return `Py_None` per PEP 451 import protocol.

### 3. Reference leak and error handling in `mlCreateModule()`

- `PyObject_GetAttrString` result (`nameobj`) was never `Py_DECREF`'d — memory leak.
- Returned `nullptr` without setting a Python exception when module name didn't start with `"sst."` — C API violation causing `SystemError`.

**Fix:** Copy string, decref `nameobj`, return `Py_None` for non-SST modules.

### 4. Additional hardening

- Added `tp_dealloc` to `ModuleLoaderType` (was `nullptr` — undefined behavior on dealloc).
- Hardened `SSTModuleFinder`: reuse single `ModuleLoader` instance, early `startswith` check, `try/except` guard.
- Added `python3.14-config` to `AC_PATH_PROGS` search list in `sst_check_python.m4`.

## Testing

All 10 Merlin tests pass with Python 3.14.2 on macOS (Apple Silicon):

| Test | Before | After |
|------|--------|-------|
| torus_64, torus_128, dragon_72, torus_5_trafficgen | SIGSEGV | ✅ PASS |
| dragon_128, dragon_128_fl, dragon_128_deferred | SIGSEGV | ✅ PASS |
| fattree_128, fattree_256, hyperx_128 | SIGSEGV | ✅ PASS |

## Files Changed

- `src/sst/core/model/python/pymodel.cc` — All code fixes
- `config/sst_check_python.m4` — Added `python3.14-config` to search list

### DIFF
--- a/config/sst_check_python.m4
+++ b/config/sst_check_python.m4
@@ -16,8 +16,8 @@ dnl check if user provided a specific python-config
 dnl search python3-config
   AS_IF([test $PYTHON_CONFIG_EXE = "no"],
     [AS_IF([test -n "$with_python"],
-        [AC_PATH_PROGS([PYTHON_CONFIG_EXE], ["python3-config" "python3.13-config" "python3.12-config" "python3.11-config" "python3.10-config" "python3.9-config"], ["no"], ["$with_python/bin"])],
-        [AC_PATH_PROGS([PYTHON_CONFIG_EXE], ["python3-config" "python3.13-config" "python3.12-config" "python3.11-config" "python3.10-config" "python3.9-config"], ["no"])])])
+        [AC_PATH_PROGS([PYTHON_CONFIG_EXE], ["python3-config" "python3.14-config" "python3.13-config" "python3.12-config" "python3.11-config" "python3.10-config" "python3.9-config"], ["no"], ["$with_python/bin"])],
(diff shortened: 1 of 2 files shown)


feldergast (dismissed):

gvoskuilen:
Run `pre-commit` on this to fix the format issue.

feldergast (approved):

PR #1587, opened 2026-04-14
Centralize profiling output from multiple sources

gvoskuilen:
* Enable JSON format output for all profiling data
* Adds a centralized "PerfReporter" class that combines profile output from verbose, timing-info, and profile points.
* Migrates some sync profiling information that was previously only available with a configure option to be available always at runtime using the sync profile point. This only applies to low-overhead profiling data, timing serialization within syncs still requires the configure option.
* Adds three "pretty-print" options for performance data: plain, as a tree, or as a comma-separated array

This PR changes run-time options:
`--print-timing-info` : Changed to just `--timing-info`, enables timing output
`--enable-profiling`: Unchanged, enables the requested profile points
`--timing-info-json`: Removed, use `--profiling-output` instead
`--profiling-output`: Controls the output for all sources of profiling data. Can accept 'stdout' or a filename or a list of 'stdout,filename'. Enable JSON by using a filename with a '.json' extension or pretty-print by using a filename with a '.txt' extension.

### DIFF
--- a/src/sst/core/profile/syncProfileTool.h
+++ b/src/sst/core/profile/syncProfileTool.h
@@ -14,6 +14,7 @@
 
 #include "sst/core/eli/elementinfo.h"
 #include "sst/core/profile/profiletool.h"
+#include "sst/core/rankInfo.h"
 #include "sst/core/sst_types.h"
 #include "sst/core/ssthandler.h"
 #include "sst/core/warnmacros.h"
@@ -22,6 +23,10 @@
 #include <cstdint>
 #include <map>
 #include <string>
+
+namespace SST::Util {
+class DataRecord;
+}
 
 namespace SST::Profile {
 
@@ -39,8 +44,10 @@ public:
 
     SyncProfileTool(const std::string& name, Params& params);
 
-    virtual void syncManagerStart() {}
+    virtual void syncManagerStart(bool UNUSED(ranksync)) {}
     virtual void syncManagerEnd() {}
+
+    virtual void updateSyncSize(size_t UNUSED(bytes), size_t UNUSED(events)) {}
 };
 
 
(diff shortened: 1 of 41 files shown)


src/sst/core/profile/syncProfileTool.h
```diff
@@ -65,12 +72,17 @@ class SyncProfileToolCount : public SyncProfileTool
...
+    void syncManagerStart(bool ranksync) override;
 
-    void outputData(FILE* fp) override;
+    void updateSyncSize(size_t bytes, size_t events) override;
+
+    void outputData(SST::Util::DataRecord* record, RankInfo rank) override;
 
 private:
     uint64_t syncmanager_count = 0;
+    uint64_t threadsync_count_ = 0;
```
feldergast, line 83: We may want to revisit how the rank versus thread sync timings are being tracked.   The ThreadSync will fire every time SyncManager does, RankSync may be left often.  This happens because RankSync requires a ThreadSync before/after it executes to makes sure all the interthread events are in the right place for a RankSync. We'll need to decide if we want to track all ThreadSync activity, or just track ThreadSyncs that happen with RankSyncs as a single sync; essentially just tracking rank versus thread-only syncs (which is what it is doing now).

This does not need to be addressed with this PR.

src/sst/core/sync/rankSyncParallelSkip.cc
```diff
@@ -95,6 +96,7 @@ RankSyncParallelSkip::registerLink(const RankInfo& to_rank, const RankInfo& from
         comm_send_map[to_rank].to_rank = to_rank;
         queue = comm_send_map[to_rank].squeue = new RankSyncQueue(to_rank);
         comm_send_map[to_rank].remote_size    = 4096;
+        comm_send_map[to_rank].squeue->setProfileTools(profile_tools_);
```
feldergast, line 99: For the future: May want to revisit the interface for the ProfilePoint.  Do we want to have a separate register call for the data collected in the send queue.

feldergast (approved):

PR #1590, opened 2026-04-16
Profile metadata

gvoskuilen:
Adds a metadata section to profile output that contains some information about the run (ranks, threads, simulated time, input file).

Also bypasses the TimingOutput object in main as that is currently only used as a pass-through to have BasicPerf output its data.


feldergast (dismissed):

kpgriesser:
I've updated our parameter sweep scripts to use the new JSON format for this branch. The scripts are now generating data. I do not see anything missing from the JSON.

feldergast (approved):

PR #1591, opened 2026-04-17
Interactive Console for Parallel Debug

kpgriesser:
Major Changes
- Single-rank, multithreaded interactive debug console support. In multithreaded simulations,
  interactive console is invoked from the syncManager, starting in thread 0.
- The user can navigate the object map, set watchpoints, etc. within a thread.
- The “info” command provides the rank, thread ID, and process ID for threads and the “thread”
  command switches control between threads.
- Ensure signal command line flags are processed before the checkpoint directory is created and
  have it fail immediately if the --checkpoint-prefix PREFIX is not valid.
- Add support for user defined commands in the interactive console, consisting of a named sequence
  of commands and supporting nested definitions
- API now provides return codes
- Remote functions return a string rather than print directly
- New command line auto-complete tab functionality. On the 1st tab, fills in the longest matching
  string prefix. On the 2nd tab: Lists all matching completion strings
- Provides EOF functionality for CTRL-D on empty line

Code Management
- clang formatting
- SimpleDebug class renamed to DebugConsole
- fixed header check violations for modified files

Command line changes
- sst command line replace sst.interactive.simpledebug with more generic sst.interactive.debugger
- removed restriction requiring serial mode only debug


gvoskuilen:
@kpgriesser Can you run `pre-commit run --all` on this branch and update?

src/sst/core/impl/interactive/debugCommands.h
```diff
@@ -0,0 +1,275 @@
...
+    std::string         str_help_;
+    ConsoleCommandGroup group_;
+    ExecutionType       exec_type_;
+
+    std::function<bool(std::string& cmd_str)>                     func_;
+    std::function<bool(std::string& cmd_str)>                     func_serial_;
+    std::function<bool(std::string& cmd_str)>                     func_thread_;
+    std::function<bool(std::string& cmd_str)>                     func_rank_serial_;
+    std::function<bool(std::string& cmd_str)>                     func_rank_parallel_;
+    std::function<bool(std::vector<std::string>& UNUSED(tokens))> func_remote_;
```
feldergast: UNUSED() macro shouldn't be needed here since this is only a declaration and not a definition of a function.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -0,0 +1,4818 @@
...
+    // Populate the command registry
+    cmdRegistry = CommandRegistry({
+
+        // Navigation
+        {
+            "help",
+            "?",
+            "<[CMD]>: show this help or detailed command help",
+            ConsoleCommandGroup::GENERAL,
+            [this](std::string& UNUSED(cmd_str)) { return cmd_help(cmd_str); },
```
feldergast: cmd_str is being used in the function so shouldn't be decorated with UNUSED().  Many of the following on lambdas have this same issue.

src/sst/core/impl/interactive/debugCommands.h
```diff
@@ -0,0 +1,275 @@
...
+    const std::string&         str_help() const { return str_help_; }
+    void                       setUserHelp(std::string& help) { str_help_ = help; }
+    const ConsoleCommandGroup& group() const { return group_; }
+    // Command Execution
+    bool                       exec(std::string& cmd_str) { return func_(cmd_str); }
+    bool                       exec_serial(std::string& cmd_str) { return func_serial_(cmd_str); }
+    bool                       exec_thread(std::string& cmd_str) { return func_thread_(cmd_str); }
+    bool                       exec_rank_serial(std::string& cmd_str) { return func_rank_serial_(cmd_str); }
+    bool                       exec_rank_parallel(std::string& cmd_str) { return func_rank_parallel_(cmd_str); }
+    bool                       exec_remote(std::vector<std::string>& UNUSED(tokens)) { return func_remote_(tokens); }
```
feldergast: tokens is used in the function.  Don't use UNUSED().

src/sst/core/impl/interactive/debugConsole.h
```diff
@@ -0,0 +1,305 @@
...
+    ExecState             eState = {};
+    std::stack<ExecState> eStack = {};
+
+    // Keep a pointer to the ObjectMap for the top level Component
+    SST::Core::Serialization::ObjectMapDeferred<BaseComponent>* base_comp_ = nullptr;
+
+    // Keep track of all the WatchPoints
+    std::vector<std::pair<WatchPoint*, BaseComponent*>> watch_points_;
+    bool                                                query_clear_watchlist();
+    bool                                                clear_watchlist(std::vector<std::string>& UNUSED(tokens));
```
feldergast: UNUSED() macro shouldn't be used in function declarations that don't also include the function definition.  There are several examples of this in this file.

src/sst/core/serialization/objectMap.h
```diff
@@ -544,16 +547,52 @@ class ObjectMap
...
        @param name Name of variable to find
+       @param confirm Prompt user to resolve duplicate match of name. Select first found if false.
 
        @return ObjectMap representing the requested variable if it is
        found, nullptr otherwise
      */
-    ObjectMap* findVariable(const std::string& name) const
+    ObjectMap* findVariable(const std::string& name, bool confirm = false) const
     {
+// TODO move body into objectMap.cc
```
feldergast: Should probably go ahead and move the code to objectMap.cc

src/sst/core/serialization/objectMap.h
```diff
@@ -544,16 +547,52 @@ class ObjectMap
...
+            if ( (ss >> n) && ss.eof() ) {
+                if ( n < 0 ) return nullptr;
+                if ( n < (int)selections.size() ) return selections[n];
+            }
+            std::cout << "Invalid entry" << std::endl;
+        }
+        std::cout << "Too many attempts\n" << std::endl;
+        return nullptr;
+#else
         auto& variables = getVariables();
```
feldergast: If this is dead code, go ahead and remove it and the #if/#else/#endif directives

src/sst/core/sync/rankSyncParallelSkip.cc
```diff
@@ -160,6 +160,76 @@ RankSyncParallelSkip::getSignals(int& end, int& usr, int& alrm)
     return sig_end_ || sig_usr_ || sig_alrm_;
 }
 
+void
+RankSyncParallelSkip::setShutdownFlags(bool enter_shutdown, Simulation_impl::ShutdownMode_t shutdown_mode)
+{
+    // This must be atomic because it can be set from any thread
+    // printf("Enter rankSync setFlags: \n input: enter_interactive %d, enter_shutdown %d, shutdown_mode %d \n",
```
feldergast: Remove commented out debug printf's in function

src/sst/core/sync/rankSyncSerialSkip.cc
```diff
@@ -139,6 +206,44 @@ RankSyncSerialSkip::execute(int thread)
...
+#ifdef SST_CONFIG_HAVE_MPI
+    int32_t local_flags[2]  = { static_cast<int32_t>(enter_shutdown_), static_cast<int32_t>(shutdown_mode_) };
+    int32_t global_flags[2] = { 0, 0 };
+    MPI_Allreduce(&local_flags, &global_flags, 2, MPI_INT32_T, MPI_MAX, MPI_COMM_WORLD);
+
+    enter_shutdown_ = global_flags[0];
+    shutdown_mode_  = global_flags[1];
+#endif
+}
+#if 0
```
feldergast: Remove dead code

src/sst/core/sync/syncManager.cc
```diff
@@ -471,15 +516,54 @@ SyncManager::findThreadSyncInterval()
...
+    // Get sim flags to exchange in threadSync
+    enter_interactive = sim_->enter_interactive_;
+    getSimShutdownFlags(enter_shutdown, shutdown_mode);
+    generate_ckpt = checkpoint_->getCheckpoint();
+}
+
 void
 SyncManager::execute()
 {
+#if 0  // SKK
```
feldergast: Remove dead code

src/sst/core/sync/syncManager.cc
```diff
@@ -549,31 +665,73 @@ SyncManager::execute()
...
+            // Each thread atomically sets shared flags in threadSync
+            threadSync_->setFlags(enter_interactive, enter_shutdown, shutdown_mode);
+        }
+
+        threadSync_->execute(); // exchange event queues, includes barrier
 
         // Handle signals for multi-threaded runs/no MPI
         if ( num_ranks_.rank == 1 ) {
             signals_received = threadSync_->getSignals(sig_end, sig_usr, sig_alrm);
+#if 0
```
feldergast: Remove dead code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -0,0 +1,4818 @@
...
+    else if ( tokens[0] == "summary" ) {
+        if ( current_thread == rank_.thread ) {
+            summary();
+        }
+    }
+    // If not DONE, process command
+    else if ( !done ) {
+        // If I am target thread, handle the incoming command
+        if ( current_thread == rank_.thread ) {
+#if 0
```
feldergast: Remove dead code

src/sst/core/sync/syncManager.h
```diff
@@ -65,6 +68,21 @@ class RankSync
...
+    virtual void setFlags(
+        bool enter_interactive, bool enter_shutdown, Simulation_impl::ShutdownMode_t shutdown_mode)     = 0;
+    /** Return exchanged interactive flags after sync */
+    virtual void getShutdownFlags(bool& enter_shutdown, Simulation_impl::ShutdownMode_t& shutdown_mode) = 0;
+    virtual void getCkptFlag(bool& generate_ckpt)                                                       = 0;
+    virtual void getFlags(
+        bool& enter_interactive, bool& enter_shutdown, Simulation_impl::ShutdownMode_t& shutdown_mode) = 0;
+    /** Clear interactive flags before next run */
+    virtual void clearFlags()                                                                          = 0;
+    virtual void interactiveExchange()                                                                 = 0;
```
feldergast: I find several implementations of this function, but am not finding where the function is ever called.  Is this a necessary function, or has the functionality been superseded by other code?

src/sst/core/simulation.cc
```diff
@@ -1747,6 +1777,10 @@ Simulation_impl::writeCheckpointConfigGraph(ConfigGraph* graph)
 void
 Simulation_impl::scheduleCheckpoint()
 {
+#if 0
```
feldergast: Remove dead code

feldergast (changes requested):
Generally looks good.  Just a few items of cleanup

feldergast (dismissed):

feldergast:
The testing platforms are having issues right now.  Setting this to WIP until that is resolved.

gvoskuilen:
The 2 rank (MR-2) job failed with the following failure modes.
You can run this locally using `$ sst-test-core -r 2`

1. There's a new warning line that gets printed in some tests now and should get filtered/ignored: "WARNING: EmptyRankSync: Checkpoint and interactive debug disabled"
  * Tests: test_Module, test_ParamComponent, test_UnitAlgebra
2. All of the signal/realtime tests fail with a time out

gvoskuilen:
Marking WIP to prevent testing until failures are addressed

tests/Makefile.inc
```diff
@@ -3,9 +3,23 @@
...
+    tests/cmd_DebugConsole_r4_rankserial1.cmd \
+    tests/cmd_DebugConsole_serial0.cmd \
+    tests/cmd_DebugConsole_serial1.cmd \
+    tests/cmd_DebugConsole_thread0.cmd \
+    tests/cmd_DebugConsole_thread1.cmd \
     tests/testsuite_default_Checkpoint.py \
     tests/testsuite_default_Component.py \
     tests/testsuite_default_ComponentExtension.py \
+    testsuite_default_DebugConsole_Checkpoint.py \
+    testsuite_default_DebugConsole.py \
```
feldergast: These need to be prefixed with tests/ since they are in the tests directory

src/sst/core/sync/syncManager.cc
```diff
@@ -400,9 +400,12 @@ SyncManager::setupSyncObjects()
         }
         else {
             rankSync_ = new EmptyRankSync(num_ranks_);
-            if ( num_ranks_.rank > 1 )
-                sim_->getSimulationOutput().output(
-                    "WARNING: EmptyRankSync: Checkpoint and interactive debug disabled\n");
+            if ( num_ranks_.rank > 1 &&
+                 (real_time_->canInitiateCheckpoint() || sim_->config.canInitiateInteractive()) ) {
```
feldergast: This doesn't catch the case where a checkpoint interval specified on the command line enables checkpointing.  You would need to check sim_->config.canInitiateCheckpoint().

src/sst/core/config.cc
```diff
@@ -664,6 +664,15 @@ Config::canInitiateCheckpoint()
     return false;
 }
 
+bool
+Config::canInitiateInteractive()
```
feldergast, line 668: This function should only check interactive_start_time_.  As with checkpointing, RealTimeAction should have a virtual canInitiateInteractive() added to the API and sst.rt.interactive should return true for that function.  

RealTimeManager would then also need a catInitializeInteractive() function that would be checked where the warning message is printed.  The realtime actions need to behave properly without having to add specific ones to Config since the realtime API is considered a third-party library API and libraries are allowed to create their own real time actions that could conceivably initiate interactive mode.

feldergast (changes requested):

src/sst/core/sync/syncManager.cc
skuntz: If I understand correctly, I would need to check both config.canInitiateCheckpoint AND real_time_->canInitiateCheckpoint. The realtime version checks all the signal related flags and the config checks the others. Then do the same with the canInitiateInteractive.

skuntz:
```
          if ( num_ranks_.rank > 1 &&
                ( real_time_->canInitiateCheckpoint() || sim_->config.canInitiateCheckpoint() ||
                real_time_->canInitiateInteractive() || sim_->config.canInitiateInteractive() ) ) {
                if ( rank_.rank == 0 )
                    sim_->getSimulationOutput().output(
                        "WARNING: EmptyRankSync: Checkpoint and interactive debug disabled\n");
            }
```

feldergast: Yes, that should work.  Or you could tailor the message to which of them is turned on, but I don't think that's necessary right now.  I suspect there could be a similar issue with thread sync for simulations only using one rank, but we can delve deeper into that after release.

feldergast (approved):

PR #1599, opened 2026-04-23
Header fixes

kpgriesser:
Branch after running `./scripts/test-includes.pl --fix` on mac and linux systems.
I'll rerun and force another update after the final sst-core code is ready it it has passed on all our systems.
I'm not running this on elements.


feldergast:
At this point to minimize the number of PRs before we branch for release, let's revisit this as the last thing before branching. Closing this PR.

PR #1624, opened 2026-05-25
Fix issue with set variable action in trace command

skuntz:
Issue: When using the 'set variable' action in a trace command, the variable was erroneously being set during trace creation instead of waiting until a trigger. 

The checkValue function (called when setting up a trace command with a 'set variable' action) was meant to catch errors and avoid a seg fault for invalid variables. It was erroneously updating the trace variable instead of just checking it. This changes the function to use a temporary value to avoid the assignment.

### DIFF
--- a/src/sst/core/serialization/objectMap.h
+++ b/src/sst/core/serialization/objectMap.h
@@ -1254,8 +1254,7 @@ public:
     {
         try {
             [[maybe_unused]]
-            decltype(SST::Core::from_string<T>(value)) v;
-            v = SST::Core::from_string<T>(value);
+            T v = SST::Core::from_string<T>(value);
             return true;
         }
         catch ( const std::exception& e ) {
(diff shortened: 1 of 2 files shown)


src/sst/core/serialization/objectMap.h
```diff
@@ -1253,7 +1253,9 @@ class ObjectMapFundamental : public ObjectMap
     bool checkValue(const std::string& value) const override
     {
         try {
-            *addr_ = SST::Core::from_string<T>(value);
+            [[maybe_unused]]
+            decltype(SST::Core::from_string<T>(value)) v;
```
feldergast: Is there a reason you can't simply use:

T v;
or
auto v = SST::Core::from_string<T>(value);

Using decltype seems like it should be a last resort.

feldergast (approved):

PR #1625, opened 2026-05-27
Transition from object map to object tree

skuntz:
Object Tree is an extension of the object map for use with the debug console. It provides an easily traversable structure to allow for storage and inspection of arbitrary types and improved usability, particularly for containers.

The object tree is organized as a tree, implemented as a vector of vectors with an enforced hierarchy and strong typing. It is dynamically built -- as the user traverses the simulation structure, new objects are added (same principle as ObjectMapDeferred). It uses a combination of CRTP, Visitor, Variant, and Polymorphism. 

Console command improvements with object tree include:

- `cd`: Components and objects named the same are explicitly disambiguated

- `ls`: Support for `ls -l `and `-ll` to control the level of detail

- `set`: Indices are now supported e.g. `set container[i][j][k]`

- watchpoints: Leverage the type information in the object tree to perform native comparison between generic types and constants; Watchpoints on a container size now possible with a small extension

- `print`: Indices, explicit verbosity, and formatting now supported and names are slightly demangled. 'print -v N -f [hex | dec | oct] container[i][j][k]`

General UI improvements:

- Ensures variable names are uniquely qualified by the object hierarchy

- Provides confirmation prompt when overwriting user-defined commands

- Adds 'show' command to display user-defined commands

- ctrl-d shuts down simulation fixing a runaway prompt on some systems

- Updates trigger output to avoid splitting text in parallel execution

- Adds paging to `print`, `info all`, `watchlist`, and `printtrace` commands (previously only enabled for `ls`)

- Resolves an issue where `print -r N`  does not print top level components


skuntz:
Replacing with PR from tactcomplabls/obj_tree_devel

PR #1626, opened 2026-05-27
Transition from object map to object tree and UI improvements

skuntz:
Object Tree is an extension of the object map for use with the debug console. It provides an easily traversable structure to allow for storage and inspection of arbitrary types and improved usability, particularly for containers.   

The object tree is organized as a tree, implemented as a vector of vectors with an enforced hierarchy and strong typing. It is dynamically built -- as the user traverses the simulation structure, new objects are added (same principle as ObjectMapDeferred). It uses a combination of CRTP, Visitor, Variant, and Polymorphism. It adds support for generic fundamentals and the numeric handler enables direct comparison against constants and other tree nodes.

Console command improvements with object tree include:

- `cd`: Components and objects named the same are explicitly disambiguated

- `ls`: Support for` ls -l` and `-ll `to control the level of detail

- `set`: Indices are now supported e.g. `set container[i][j][k]`

- watchpoints: Leverage the type information in the object tree to perform native comparison between generic types and constants; Watchpoints on a container size now possible with a small extension

- `print`: Indices, explicit verbosity, and formatting now supported and names are slightly demangled. `print -v N -f [hex | dec | oct] container[i][j][k]`

General UI improvements:

- Ensures variable names are uniquely qualified by the object hierarchy
- Supports `cd` into any type, including containers, and clearer printing of types
- Provides confirmation prompt when overwriting user-defined commands
- Adds `show `command to display user-defined commands
- ctrl-d shuts down simulation fixing a runaway prompt on some systems
- Updates trigger output to avoid splitting text in parallel execution
- Adds paging to `print`, `info all`, `watchlist`, and `printtrace` commands (previously only enabled for ls)
- Resolves an issue where `print -r N` does not print top level components
- Provides warnings for read-only variables


src/sst/core/from_string.h
```diff
@@ -20,13 +20,30 @@
...
+
+// Helper trait to detect tuple types
+template <typename T>
+struct is_tuple : std::false_type
+{};
+
+template <typename... Args>
+struct is_tuple<std::tuple<Args...>> : std::true_type
+{};
+
```
donofrio: This code can be removed, print formatting for debugConsole is now handled locally to that class.  I neglected to remove this code.

donofrio: (this goes for all changes in this file)

kpgriesser:
Testing update: 
- All `sst-test-core` and `sst-ext-tests` are passing on all TCL platforms.
- Completed `valgrind` and `address sanitizer` testing with no issues

To run `sst-ext-tests` with these changes use:
```
git checkout `obj_tree_merge`
cd build
cmake .. -DENABLE_ALL_TESTS=ON -DENABLE_NEW_TESTS=ON 
ctest
```
Once this PR is merged into `devel`, we will update `sst-ext-tests/devel` to match.

At this point, there is no additional work planned for this branch other than writing new tests as time allows.

kpgriesser:
In addition to the 'planned improvements' listed in the sst-docs web page, 

https://sst-simulator.org/sst-docs/docs/guides/features/debug#planned-improvements

TCL has issues and enhancements tracked in GitHub here:

https://github.com/tactcomplabs/sst-core/issues

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -458,8 +487,20 @@ DebugConsole::consoleExecute(const std::string& msg)
...
     std::cout << msg << std::endl;
 
-    // Create a new ObjectMap
-    obj_ = getComponentObjectMap();
+    // Try and serialize the entire map
+    // auto totalMap = SST::Core::Serialization::ObjectMapToTree::convertAndSerialize("root", obj_);
+    // printf("---- total map:\n ----");
+    // totalMap->Dump(2);
+
+    // if(curObj_ == nullptr || objTree_->isEmpty()){
```
feldergast: Remove commented out code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -758,6 +798,45 @@ DebugConsole::cmd_help(std::string& UNUSED(cmd_str))
...
+DebugConsole::cmd_show(std::string& UNUSED(cmd_str))
+{
+    // First check for specific command help
+    if ( tokens.size() == 1 ) {
+        if ( cmdRegistry.getUserRegistryVector().size() > 0 ) {
+            std::cout << "--- User-Defined Commands ---" << std::endl;
+            for ( const auto& cmd : cmdRegistry.getUserRegistryVector() ) {
+                // if ( g.first == c.group() ) {
+                //     std::cout << c << std::endl;
+                // }
```
feldergast: Remove commented out code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2666,8 +2863,13 @@ DebugConsole::cmd_logging(std::string& UNUSED(cmd_str))
         return false;
     }
     if ( tokens.size() > 1 ) {
+        if ( tokens[1] == replayFilePath ) {
+            std::cout << "Logging file path and replay file path cannot be the same" << std::endl;
```
feldergast, line 2871: Is there a specific reason to use std::cout instead of dout here?

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2687,6 +2889,10 @@ DebugConsole::cmd_replay(std::string& UNUSED(cmd_str))
         return false;
     }
     if ( tokens.size() > 1 ) {
+        if ( tokens[1] == loggingFilePath ) {
+            std::cout << "Replay file path and logging file path cannot be the same" << std::endl;
```
feldergast, line 2897: dout?

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+            constNode = std::make_unique<Core::Serialization::IntegerObj>(v, nullptr);
+        }
+        catch ( ... ) {
+        }
+        try {
+            double v  = std::stod(var);
+            constNode = std::make_unique<Core::Serialization::FloatObj>(v, nullptr);
+        }
+        catch ( ... ) {
+        }
```
feldergast, line 3170: I understand what's happening here, but this is an unusual use of try/catch (at least to me), so I would add a comment on why it is being used.

It also looks like you will always end up with a FloatObj, since strings representing integers are also valid floating point formats (i.e., the decimal point is optional).  You likely need to do the FloatObj first and then do the IntegerObj.  It does mean integer constants will end up creating both a FloatObj and an IntegerObj, but the FloatObj should get cleaned up by the unique_ptr.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+            path.push_front(parent->getObjName());
+            parent = parent->getParent();
+        }
+        op->objectsToCompare.emplace_back(
+            std::move(path), Core::Serialization::ObjTreeComparison::valType::OBJ, opObj2->clone());
+    }
+    else { // V2 is value string
+        std::unique_ptr<Core::Serialization::ObjTreeCont> constNode = nullptr;
+        // Convert the string value to an Integer or FloatObj
+        try {
```
feldergast, line 3204: See comment for same try/catch structure above.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2988,32 +3332,41 @@ parseAction(std::vector<std::string>& tokens, size_t& index, Core::Serialization
...
             std::cout << "Unknown variable: " << tvar << std::endl;
             return nullptr;
         }
 
         // Is variable fundamental
-        if ( !map->isFundamental() ) {
-            std::cout << "Can only set fundamental variable, " << tvar << " is not fundamental" << std::endl;
-            return nullptr;
-        }
+        //  if ( !map->isFundamental() ) {
```
feldergast: Remove commented out code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2988,32 +3332,41 @@ parseAction(std::vector<std::string>& tokens, size_t& index, Core::Serialization
...
+            name.insert(0, parent->getObjName());
+            parent = parent->getParent();
         }
 
-        std::string name = obj->getFullName() + "/" + tvar;
-
-        return new WatchPoint::SetVarWPAction(name, map, tval);
+        return new WatchPoint::SetVarWPAction(name, map->clone(), tval);
     }
 #if 0 // Do users want a heartbeat action?
```
feldergast: Probably okay to leave this in, but I would move the comment to the next line

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -3128,18 +3482,22 @@ DebugConsole::cmd_watch_rank_parallel(std::string& cmd_str)
...
             result << "Invalid comparison argument passed to watch command" << std::endl;
             return false;
         }
         size_t wpIndex = watch_points_.size();
-        auto*  pt      = new WatchPoint(wpIndex, name, c);
+        c->print(ss, 0, c->operators.size() * 2);
+        name     = ss.str();
+        auto* pt = new WatchPoint(wpIndex, name, c);
 
 #if 0 // watch variables currently don't trace, but they could automatically
```
feldergast: Move comment to next line and indent as if #if 0 wasn't there

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -4229,12 +4633,15 @@ DebugConsole::handleCommand()
     else if ( !done ) {
         // If I am target thread, handle the incoming command
         if ( current_thread == rank_.thread ) {
-            if ( obj_ == nullptr ) {
-                // Create a new ObjectMap
-                obj_ = getComponentObjectMap();
-                // Descend into the name_stack
+            // if(curObj_== nullptr || objTree_->isEmpty()){
```
feldergast: Delete commend out code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -4396,12 +4804,15 @@ DebugConsole::receiveCommandRankSerial()
         auto consoleCommand = cmdRegistry.seek(tokens[0], CommandRegistry::SEARCH_TYPE::BUILTIN);
         if ( consoleCommand.second ) {
             // Execute in target thread
-            if ( obj_ == nullptr ) {
-                // Create a new ObjectMap
-                obj_ = getComponentObjectMap();
-                // Descend into the name_stack
+            // if(curObj_ == nullptr || objTree_->isEmpty()){
```
feldergast: Delete commented out code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -4655,13 +5067,16 @@ DebugConsole::executeThread(const std::string& msg)
         handleCommand();
     }
     else {
-        // Init object map
-        if ( obj_ == nullptr ) {
-            // Create a new ObjectMap
-            obj_ = getComponentObjectMap();
-            // Descend into the name_stack
+        // Init object tree
+        // if( curObj_ == nullptr || objTree_->isEmpty() ){
```
feldergast, line 5041: Delete commented out code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -4730,9 +5148,13 @@ DebugConsole::executeRankParallel(const std::string& UNUSED_WO_MPI(msg))
...
     if ( rank_.rank == 0 && rank_.thread == 0 ) {
+        // Set autoCompleteEnable = false for MPI runs
+        autoCompleteEnable = false;
+
         // Clear R0 result string
         result.str("");
         result.clear();
+
         // Print Summary
         std::cout << "\nINTERACTIVE CONSOLE" << std::endl;
```
feldergast, line 5129: dout??

src/sst/core/impl/interactive/Makefile.inc
```diff
@@ -10,4 +10,5 @@ sst_core_sources += \
 	impl/interactive/cmdLineEditor.h \
 	impl/interactive/debugCommands.h \
 	impl/interactive/debugStream.h  \
-	impl/interactive/debugStream.cc
\ No newline at end of file
+	impl/interactive/debugStream.cc \
+	impl/interactive/ObjTreeHelpers.cc
```
feldergast: Add newline at end of file

src/sst/core/impl/interactive/ObjTree.h
```diff
@@ -0,0 +1,985 @@
...
+
+namespace SST::Core::Serialization {
+
+class ObjTreeCont
+{
+public:
+    enum class NodeKind { Generic, Integer, Float, String, Bool, Component, Container, GenericVal };
+
+    ObjTreeCont() :
+        parent_(nullptr),
```
feldergast: Move these initializations to the point of declaration for the class variables.  In general, we are moving to initializing all eligible class members this way to better avoid accidentally uninitialized variables.

src/sst/core/impl/interactive/ObjTree.h
```diff
@@ -0,0 +1,985 @@
...
+#include <type_traits>
+#include <typeinfo>
+#include <unordered_set>
+#include <utility>
+#include <variant>
+#include <vector>
+
+namespace SST::Core::Serialization {
+
+class ObjTreeCont
```
feldergast: It would be really good to have the ObjectTree API documented in this class and any of its children that introduce new functions. For example, what does apply() do?  The name seems to imply that it should take some action on the ObjectTree, but in at least one instance, it just prints a message.  Have a comment stating its purpose would be good.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -458,8 +487,20 @@ DebugConsole::consoleExecute(const std::string& msg)
...
+    // auto totalMap = SST::Core::Serialization::ObjectMapToTree::convertAndSerialize("root", obj_);
+    // printf("---- total map:\n ----");
+    // totalMap->Dump(2);
+
+    // if(curObj_ == nullptr || objTree_->isEmpty()){
+    if ( objTree_->isEmpty() ) {
+        objTree_->BuildTree(getComponentInfoMap());
+        curObj_ = objTree_;
+    }
+    if ( curObj_ == nullptr ) {
```
gvoskuilen: `else if` 

If you went through the `if` statement and `curObj_` is nullptr then `objTree` is also nullptr.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -534,39 +575,32 @@ DebugConsole::consoleExecute(const std::string& msg)
...
-    // Save the position on the name_stack, and clear obj_
+    // Save the position on the name_stack, and clear objTree_ and curObj_
     save_name_stack();
+
     done = true;
     return retState;
 }
 
-// Save the name stack of the current position, and clear obj_
+// Save the name stack of the current position, and clear objTree and curObj_
```
gvoskuilen: objTree_

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -1606,48 +1693,63 @@ DebugConsole::cmd_cd_remote(std::vector<std::string>& tokens)
...
     if ( selection == ".." ) {
-        auto* parent = obj_->selectParent();
-        if ( parent == nullptr ) {
-            result << "Already at top of object hierarchy" << std::endl;
+        auto* parent = curObj_->getParent();
+        if ( parent ) {
+            curObj_ = parent;
+        }
+        else {
+            printf("Already at top of object hierarchy\n");
```
gvoskuilen: std::cout for consistency? Or sst::output?

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -1606,48 +1693,63 @@ DebugConsole::cmd_cd_remote(std::vector<std::string>& tokens)
...
     }
 
-    bool                                 loop_detected = false;
-    SST::Core::Serialization::ObjectMap* new_obj       = obj_->selectVariable(selection, loop_detected, confirm_);
-    if ( !new_obj || (new_obj == obj_) ) {
-        result << "Unknown object in cd command: " << selection << std::endl;
+    // Search children by name
+    SST::Core::Serialization::ObjTreeCont* target = curObj_->findByName(selection);
+    if ( !target ) {
+        printf("Unknown object in cd command: %s\n", selection.c_str());
```
gvoskuilen: Mixing printf & cout in this file.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -1779,32 +1894,126 @@ DebugConsole::cmd_print_remote(std::vector<std::string>& tokens)
...
-        // Print current object
-        obj_->list(recurse);
-        return true;
+    // See if have a -v or not
+    int print_verbose = 0;
+    pos               = containsArg(tokens, "-v");
+    if ( std::string::npos != pos ) {
+        // Got a -v
+        if ( tokens.size() < (pos + 2) ) {
+            printf("Invalid format for print command (print [-r N] [-v N] [-f <base>] [<obj>][[idx][idx]])\n");
```
gvoskuilen: Do not mix printf/cout

src/sst/core/impl/interactive/ObjMapToTree.h
```diff
@@ -0,0 +1,381 @@
...
+    /**
+     * Takes a ComponentObj, serializes its BaseComponent via
+     * ObjectMapDeferred, and populates the ComponentObj's children
+     * with the serialized variables.
+     *
+     * @param compNode  The ComponentObj node to expand
+     * @param recursive If true, recursively convert all children
+     * @return true if serialization succeeded
+     */
+    static bool serializeComponent(ComponentObj* compNode, bool recursive = false)
```
feldergast, line 239: Not sure I like this function name since it's overloading "serialize".  This is really only happening in mapping mode and is not a general function for serializing a component.  I don't have a better name for now, but we probably want to revisit this at some point and use something different than "serialize" in the function and class names.

src/sst/core/impl/interactive/ObjMapToTree.h
```diff
@@ -0,0 +1,381 @@
...
+        for ( auto it = subComps.begin(); it != subComps.end(); ++it ) {
+
+            BaseComponent* sub = it->second.getComponent();
+            if ( sub ) {
+                addrs.push_back(static_cast<void*>(sub));
+            }
+        }
+    }
+
+    static void serializeSubComponents(ObjTreeCont* parent, ComponentInfo* compInfo, bool recursive)
```
feldergast, line 331: Same issue here with the word "serialize" as noted above.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -1606,48 +1693,63 @@ DebugConsole::cmd_cd_remote(std::vector<std::string>& tokens)
...
-            dynamic_cast<Core::Serialization::ObjectMapDeferred<BaseComponent>*>(obj_);
-        if ( base_comp ) base_comp_ = base_comp;
-    }
+    curObj_ = target;
+
     return true;
 }
 
+std::vector<size_t>
+DebugConsole::parseBracketIndices(std::string& token)
```
gvoskuilen, line 1725: If this will be parsing user-supplied tokens then it needs some error checking.
1. pushing ints into a vector of size_t will cause any negative integers to wrap
2. Handle `[` with no matching `]` (or nested, or otherwise bad syntax)
3. Handle bad inputs between `[` and `]` such as `[]`, `[cat]`. `atoi()` will still convert these/return 0 so the `indices` will contain `0` rather than be empty.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -1752,12 +1857,22 @@ DebugConsole::cmd_print_remote(std::vector<std::string>& tokens)
     // Index in tokens array where we may find the variable name
     size_t var_index = 1;
 
+    if ( tokens.size() < 2 ) {
+        printf("Invalid format for print command (print [-r N] [-v N] [-f <base>] [<obj>][[idx][idx]])\n");
```
gvoskuilen: Mixing cout/printf

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -1752,12 +1857,22 @@ DebugConsole::cmd_print_remote(std::vector<std::string>& tokens)
...
+    size_t      pos     = containsArg(tokens, "-r");
     // See if have a -r or not
-    int         recurse = 4; // default -r depth
+    int         recurse = 0; // default -r depth
     std::string tok     = tokens[1];
     if ( (tok.size() >= 2) && (tok[0] == '-') && (tok[1] == 'r') ) {
         // Got a -r
-        std::string num = tok.substr(2);
+        if ( tokens.size() < (pos + 2) ) {
+            printf("Invalid format for print command (print [-r N] [-v N] [-f <base>] [<obj>][[idx][idx]])\n");
```
gvoskuilen: mixing cout/printf

src/sst/core/impl/interactive/ObjTree.h
```diff
@@ -0,0 +1,985 @@
...
+    bool     isLeaf() const
+    {
+        return kind_ == NodeKind::Integer || kind_ == NodeKind::Float || kind_ == NodeKind::String ||
+               kind_ == NodeKind::Bool || kind_ == NodeKind::GenericVal;
+    }
+
+    virtual void clear() { children_.clear(); }
+
+protected:
+    ObjTreeCont*                              parent_;
```
feldergast: Initialize these to default values here.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -1906,63 +2115,52 @@ DebugConsole::cmd_set_remote(std::vector<std::string>& tokens)
...
+                    return true;
+                }
+                else {
+                    result << "Failed to set " << tokens[1] << " to " << tokens[2] << " (type: " << target->getType()
+                           << ")" << std::endl;
+                    return false;
+                }
+            }
+            else {
+                result << "ERROR: Unable to set value - object not indexable" << std::endl;
```
gvoskuilen: Consistently use (or not) the "ERROR:" prefix. E.g., on lines 2141, 1987, 2153, etc.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
 
+// Create a helper object that contains the paths to the objects, the types being compared
+//  as well as the operators used. Creates and returns a new ObjTreeComparison that the caller owns.
+//  nullptr returned as an error condition.
+Core::Serialization::ObjTreeComparison*
+parseComparison(std::vector<std::string>& tokens, size_t& index, Core::Serialization::ObjTreeCont* curObj,
+    Core::Serialization::ObjTreeComparison* op = nullptr)
+{
+
+    const std::string& var = tokens[index++];
```
gvoskuilen, line 3087: I assume it's guaranteed by the call that index < tokens.size()?

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+        op = new Core::Serialization::ObjTreeComparison();
+    }
+    op->operators.push_back(Core::Serialization::ObjTreeComparison::getOperationFromString(opstr));
+
+    std::string v2;
+    if ( Core::Serialization::ObjTreeComparison::Op::CHANGED != op->operators.back() ) {
+        if ( index >= tokens.size() ) {
+            std::cout << "Invalid format for trigger test. Valid formats are <var> changed and <var> <op> <val>"
+                      << std::endl;
+            return nullptr;
```
gvoskuilen, line 3104: if `op` was new'd on line 3090, it leaks here.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+        }
+        v2 = tokens[index++];
+    }
+
+    // Is operator valid
+    if ( op->operators.back() == Core::Serialization::ObjTreeComparison::Op::INVALID ) {
+        std::cout << "Unknown comparison operation specified in trigger test" << std::endl;
+        return nullptr;
```
gvoskuilen, line 3113: if `op` was new'd on line 3090, it leaks here.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+        }
+        try {
+            double v  = std::stod(v2);
+            constNode = std::make_unique<Core::Serialization::FloatObj>(v, nullptr);
+        }
+        catch ( ... ) {
+        }
+        if ( constNode == nullptr ) {
+            std::cout << "Unknown Constant in expression" << std::endl;
+            return nullptr;
```
gvoskuilen, line 3233: if `op` was new'd on line 3090, it leaks here.

src/sst/core/impl/interactive/ObjTree.h
```diff
@@ -0,0 +1,985 @@
...
+
+    bool setElementFromString(std::vector<size_t> indices, std::string val)
+    {
+        ObjTreeCont* elem    = getElementAt(indices);
+        bool         success = elem != nullptr;
+        if ( elem ) {
+            success = elem->setFromString(val);
+        }
+        if ( !success ) {
+            std::cout << "Element not found at path {";
```
feldergast: Do you want to be able to use dout here?

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+    else {
+        std::unique_ptr<Core::Serialization::ObjTreeCont> constNode = nullptr;
+        // Convert the string value to an Integer or FloatObj
+        try {
+            int64_t v = std::stoll(var);
+            constNode = std::make_unique<Core::Serialization::IntegerObj>(v, nullptr);
+        }
+        catch ( ... ) {
+        }
+        try {
```
gvoskuilen: This will * always * overwrite the int parse because `stod` parses an int just fine.

src/sst/core/impl/interactive/debugConsole.cc
feldergast: Should standardize everything on either printf or cout or dout rather than have a mix of all three.

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+            parent = parent->getParent();
+        }
+        op->objectsToCompare.emplace_back(
+            std::move(path), Core::Serialization::ObjTreeComparison::valType::OBJ, opObj->clone());
+
+        // In changed mode, we do not use v2, but add a "placeholder" object to the compairson list. This is a touch
+        // wasteful, but makes our life a lot easier later
+        if ( op->operators.back() == Core::Serialization::ObjTreeComparison::Op::CHANGED ) {
+            std::deque<std::string> tmpPath;
+            tmpPath.push_front("unnit");
```
gvoskuilen: `uninit`?

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -2862,6 +3072,140 @@ DebugConsole::cmd_document(std::string& UNUSED(cmd_str))
...
+        parent = opObj->getParent();
+        path.push_front(opObj->getObjName());
+        while ( parent && !parent->isRoot() ) {
+            path.push_front(parent->getObjName());
+            parent = parent->getParent();
+        }
+        op->objectsToCompare.emplace_back(
+            std::move(path), Core::Serialization::ObjTreeComparison::valType::OBJ, opObj->clone());
+
+        // In changed mode, we do not use v2, but add a "placeholder" object to the compairson list. This is a touch
```
gvoskuilen: "comparison"

src/sst/core/impl/interactive/ObjTreeHelpers.h
```diff
@@ -0,0 +1,393 @@
...
+            if ( start_state == CLEAR ) { // Not previously triggered
+                state_ = TRIGGER;         // State becomes trigger record
+            }
+            // printf("    Sample: trigger\n");
+
+        } // if trigger
+
+        if ( start_state == TRIGGER || start_state == POSTTRIGGER ) { // trigger record or post trigger
+            state_ = POSTTRIGGER;                                     // State becomes post trigger
+            // printf("    Sample: post trigger\n");
```
feldergast: Removed commented out code

src/sst/core/impl/interactive/ObjTreeHelpers.h
```diff
@@ -0,0 +1,393 @@
...
+                triggerCycle = cycle;
+            }
+
+            // Sample all the trace object buffers
+            ObjTreeCont* varBuffer_;
+
+            for ( size_t obj = 0; obj < objBuffers_.size(); obj++ ) {
+                auto& [bufVec, storedTriggerIdx] = objBuffers_[obj];
+                varBuffer_                       = bufVec.back().get();
+                // varBuffer_->sample(cur_, trigger);
```
feldergast: Remove commented out code

src/sst/core/impl/interactive/ObjTreeHelpers.h
```diff
@@ -0,0 +1,393 @@
...
+            if ( numRecs_ < bufSize_ ) {
+                tagBuffer_[cur_] = state_;
+                numRecs_++;
+                cur_ = (cur_ + 1) % bufSize_;
+                if ( cur_ == 0 ) first_ = 0; // 1;
+            }
+            else { // Buffer full
+                // Check to see if we are overwriting trigger
+                if ( tagBuffer_[cur_] == TRIGGER ) {
+                    // printf("    Sample Overrun\n");
```
feldergast: Remove commented out code

src/sst/core/impl/interactive/ObjTreeHelpers.h
```diff
@@ -0,0 +1,393 @@
...
+            os << "buf[" << i << "] " << handlerBuffer_.at(i) << " @" << cycleBuffer_.at(i) << " ("
+               << state2char.at(tagBuffer_.at(i)) << ") ";
+
+            std::ostringstream tmpBuf;
+            std::string        objNames;
+            for ( size_t obj = 0; obj < objBuffers_.size(); obj++ ) {
+                ObjTreeCont* varBuffer_ =
+                    std::get<std::vector<std::unique_ptr<ObjTreeCont>>>(objBuffers_[obj])[buf].get(); // DDD: Maybe?
+                varBuffer_->Dump(1, std::ios_base::dec, tmpBuf);
+                // std::cout << varBuffer_->getObjName() << "=" << varBuffer_->get() << " ";
```
feldergast: Remove commented out code

src/sst/core/impl/interactive/debugConsole.cc
```diff
@@ -4229,12 +4633,15 @@ DebugConsole::handleCommand()
...
-                // Create a new ObjectMap
-                obj_ = getComponentObjectMap();
-                // Descend into the name_stack
+            // if(curObj_== nullptr || objTree_->isEmpty()){
+            if ( objTree_->isEmpty() ) {
+                objTree_->BuildTree(getComponentInfoMap());
+                curObj_ = objTree_;
                 cd_name_stack();
             }
+            if ( curObj_ == nullptr ) {
```
gvoskuilen: `else if` since this does nothing if the first `if` leaves 'curObj_` as `nullptr`

src/sst/core/impl/interactive/ObjTreeHelpers.h
```diff
@@ -0,0 +1,393 @@
...
+            os << ss.str(); // Print everything at once to avoid splitting
+        }
+    }
+
+    void printVars(std::stringstream& ss)
+    {
+        for ( size_t obj = 0; obj < objBuffers_.size(); obj++ ) {
+            ObjTreeCont* varBuffer_ =
+                std::get<std::vector<std::unique_ptr<ObjTreeCont>>>(objBuffers_[obj]).back().get();
+            // ss << SST::Core::to_string(varBuffer_->getName()) << " ";
```
feldergast: Remove commented out code

src/sst/core/serialization/objectMap.h
```diff
@@ -1336,67 +1354,69 @@ class ObjectMapFundamental : public ObjectMap
...
-                return new ObjectMapComparison_var<REF, double>(
-                    name, addr_, op, name2, static_cast<double*>(var2->getAddr()));
-            }
-            else if ( type == "long double" ) {
-                return new ObjectMapComparison_var<REF, long double>(
-                    name, addr_, op, name2, static_cast<long double*>(var2->getAddr()));
-            }
-        } // end if first var is arithmetic
+            return new ObjectMapComparison_var<REF, T>(name, addr_, op, name2, static_cast<T*>(var2->getAddr()));
+        } /*
```
feldergast, line 1358: Remove commented out code

src/sst/core/serialization/objectMapDeferred.h
```diff
@@ -122,6 +128,25 @@ class ObjectMapDeferred : public ObjectMap
     std::string type_ = "";
 };
 
+class ComponentSerializer : public ObjectMapDeferred<BaseComponent>
```
feldergast, line 131: Probably want to change this name since it is for mapping mode only and not for general Component serialization

src/sst/core/serialization/ObjTreeLeaves.h
```diff
@@ -0,0 +1,996 @@
...
+class ComponentInfoMap;
+}
+
+namespace SST::Core::Serialization {
+class ObjTreeCont
+{
+public:
+    enum class NodeKind { Generic, Integer, Float, String, Bool, Component, Container, FRef, GenericVal };
+
+    ObjTreeCont() :
```
feldergast: Put the default values in at point of variable definition instead of in constructor initialization list

src/sst/core/impl/interactive/ObjMapToTree.h
```diff
@@ -0,0 +1,381 @@
...
+                    co->makeReadOnly();
+                }
+                return co;
+            }
+        }
+
+        auto* node = new ObjTreeCont(name, type);
+        if ( recursive ) {
+            for ( const auto& [n, m] : objMap->getVariables() )
+                if ( auto* child = convert(n, m, false) ) node->addChildObj(child);
```
gvoskuilen, line 193: Is this supposed to keep passing the `recursive=true` flag or no?

src/sst/core/watchPoint.h
```diff
@@ -196,18 +196,18 @@ class WatchPoint : public Clock::HandlerBase::AttachPoint, public Event::Handler
...
-    std::string                                            name_;
-    Core::Serialization::TraceBuffer*                      tb_ = nullptr;
-    size_t                                                 wpIndex;
-    HANDLER                                                handler        = ALL;
-    bool                                                   trigger        = false;
-    HANDLER                                                triggerHandler = HANDLER::NONE;
-    size_t                                                 triggerCount   = 0;
-    bool                                                   reset_         = false;
-    WPAction*                                              wpAction;
+    // size_t                                                 numCmpObj_ = 0;
```
feldergast, line 199: Remove commented out code

feldergast (changes requested):
Overall looks good. There are several comments on cleanup that can be done.  Also, there are some duplicate files in the serialization and interactive directories.

It would also be nice to standardize on dout versus cout (and there are several instances of using printf).  I tried to note them where I saw them, but probably missed some.

gvoskuilen (changes requested):

feldergast:
There's at least one new file missing from Makefile.am that caused the make dist test to fail:

tests/refFiles/test_DebugConsole_rankserial0.out

There may be others that weren't reported.

feldergast:
Now failing with this compile error on make dist:

In file included from baseComponent.cc:33:
09:01:33 ../../../src/sst/core/watchPoint.h:17:10: fatal error: sst/core/impl/interactive/ObjTreeHelpers.h: No such file or directory
09:01:33  #include "sst/core/impl/interactive/ObjTreeHelpers.h"
09:01:33           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Probably means ObjTreeHelpers.h is not in the impl/interactive/Makefile.inc file

src/sst/core/impl/interactive/debugConsole.cc
donofrio: Agree we should have a consistent stream story.  I changed everything to be:
- removal of all printf
- Single line error / warnings / status all go to std::cout
- multi-line output goes to dout
- Some functions print into stringstream, `result` but those also go out via std::cout

src/sst/core/impl/interactive/debugConsole.cc
donofrio: Good catch - we likely always got a FloatObj. I changed so we still check for integer first, but it is more aggressive in making sure it was *actually* an integer by seeing if the number of characters processed is the same as the number in the input (so if there is a trailing decimal, or any fraction it skips to attempt to process it as a float). Also added comments on why we use the catch(...) - a bit overkill, but catch if out of range or invalid then print an error and exit gracefully.

src/sst/core/impl/interactive/debugConsole.cc
donofrio: excellent point. fixed

src/sst/core/impl/interactive/debugConsole.cc
donofrio: Fixed (here and in the other spots)

src/sst/core/impl/interactive/debugConsole.cc
donofrio: Yes - that was some faulty logic - thanks for catching that. Coded up the function a bit more defensively, (see other comment on this function)

src/sst/core/impl/interactive/debugConsole.cc
donofrio: 🤦

src/sst/core/impl/interactive/debugConsole.cc
donofrio: Fixed - thanks for catching that.

src/sst/core/impl/interactive/debugConsole.cc
donofrio: good catch - I added some additional error checking to avoid negative values, empty brackets and non-numeric inputs

feldergast:
Still missing a file for make dist

```
In file included from baseComponent.cc:33:
08:01:47 ../../../src/sst/core/watchPoint.h:17:10: fatal error: sst/core/impl/interactive/ObjTreeHelpers.h: No such file or directory
08:01:47  #include "sst/core/impl/interactive/ObjTreeHelpers.h"
08:01:47           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

skuntz:
@feldergast Do all .h and .cc files in impl/interactive need to be included in Makefile.inc?

feldergast:
Yes.  Make dist creates a tarball of all the files in the repo.  There are different entries in Makefile.am for things that get installed, versus things that just get built or added to the tarball.

feldergast (approved):

gvoskuilen (approved):

PR #1628, opened 2026-06-04
Remove check_param_type()

jmlapre:
It isn't pythonic at all.  Python uses duck typing or fails with a pretty clear message.  Plus inspect.stack() is fairly heavyweight.


berquist (approved):
Justin and I have talked about this a few times.  The type annotations are complete and correct enough that all of this runtime checking can be removed in favor of mypy.

PR #1629, opened 2026-06-10
Additions to the element Clock API and other small fixes:

feldergast:
- Clock::Handler now tracks the clock it is associated with and whether or not it is currently active.  This allows more efficient checkpointing of clock handlers.
- Added activate() and deactivate() to Clock::HandlerBase which can be used in place of unregisterClock() and reregisterClock().
- Added a templated registerClock() function that creates and returns the Clock::Handler that can be used to stop and start the handler.  This will become the new API for clocks and will ensure that a clock handler is ever only registered with a single clock.
- Added a new test to test the different ways to register, stop and restart clocks
- Added a templated BasicEvent class that can be used to easily create simple events through use of a using statement.
- Made it so that simulations with no cross-rank links, but with a checkpoint-sim-period set will set the minimum partition latency to the checkpoint period so that a sync object will be created. Otherwise, the checkpoints don't actually happen becuase they are controlled by the sync object.


gvoskuilen (approved):
LGTM

PR #1632, opened 2026-07-23
Added registerOrderedClock() to BaseComponent API

feldergast:
The new function will register clock handlers that are guaranteed to be called in the order they were registered (when registed at the same frequency).  This guarantee if for the entire ComponentTree. The handlers are registered with a Clock::Group that will maintain the order, even when the clocks are activated and deactivated.


jpkenny (dismissed):
LGTM

gvoskuilen (approved):

PR #1635, opened 2026-07-23
Update GitHub Actions versions

berquist:
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/ requires two things:
- Updating GitHub Actions to versions that use Node 24 instead of 20
- `action/checkout` specifically guards against `pull_request_target` workflows unless opting in to a specific unsafe flag, which I've determined is safe for us to do (https://docs.github.com/en/actions/reference/security/securely-using-pull_request_target).

This updates most, but not all, of the actions to use the latest major version.  I think there are a few others that are using Node 20, specifically pre-commit, but the Node 20 removal date is not immediate, and updating pre-commit actually requires switching to https://github.com/j178/prek-action.  This PR is most important for the `actions/checkout` change.

I'm also unpinning a few of the actions from refs and switching to moving tags, at least until we move to full Autotester 2 usage.

### DIFF
--- a/.github/workflows/PR_clang-format_test.yml
+++ b/.github/workflows/PR_clang-format_test.yml
@@ -60,10 +60,13 @@ jobs:
 
 #   Checkout this users SST-core repo/branch
     - name: Checkout Users SST-Core source
-      uses: actions/checkout@v4
+      uses: actions/checkout@v7
       with:
         path: ./sst-core_source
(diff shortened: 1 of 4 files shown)


feldergast (approved):

feldergast:
Since this PR is fixing the clang-format issues, I manually removed clang-format-fail and added clang-format-test.

Issue #1636, opened 2026-07-24
Core: subtle bug in testing_compare_filtered_diff()


jmlapre:
https://github.com/sstsimulator/sst-core/blob/837f3495d0368d905191294edad6272547e5c6f2/src/sst/core/testingframework/sst_unittest_support.py#L1129

`filters` is initialized to `list()`.  It should be set to `None` and then handled at the top of the function.  See the link below for more details.  This is one of the quirks of Python!

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

PR #1637, opened 2026-07-24
Fix default mutable argument: filters

jmlapre:
Otherwise filters is initialized when the function is defined and *not* when it is called as one would expect.

See https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments for more information.


berquist (dismissed):

feldergast (approved):
LGTM

jmlapre:
Closes #1636

PR #1638, opened 2026-07-27
Use unsecure Node version for now

jmlapre:
Use checkout@v4

Also add ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION env. variable

We *have* to circle back and fix this.  This is only so we can get PRs moving again.


berquist:
This is a duplicate from https://github.com/sstsimulator/sst-core/pull/1635/.

PR #1641, opened 2026-08-31
Add floating point tolerance for floating point statistics in tests

gvoskuilen:
Adds a comparison function for tests that have floating point values to use a tolerance. This fixes floating point rounding differences between gcc versions on recent Rocky/RHEL platforms. The statistics and checkpoint/statistics tests use the new comparison.

### DIFF
--- a/src/sst/core/testingframework/sst_unittest_support.py
+++ b/src/sst/core/testingframework/sst_unittest_support.py
@@ -1097,6 +1097,7 @@ class CheckpointRefFileFilter(IgnoreAllBeforeFilter):
 
 
 
+
 def _read_and_filter(fileloc: str, filters: Sequence[LineFilter], sort: bool, is_ref: bool) -> List[str]:
     lines = list()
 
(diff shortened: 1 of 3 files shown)


jmlapre (approved):

src/sst/core/testingframework/sst_unittest_support.py
```diff
@@ -1181,6 +1180,99 @@ def testing_compare_filtered_diff(
...
+            if abs(ov - rv) > threshold:
+                diffs.append("< {0}> {1}  [float field {2}: ref={3}, out={4}, diff={5}, threshold={6}]\n".format(
+                    ref_line, out_line, j, rv, ov, abs(ov - rv), threshold))
+                break
+
+    for line in out_lines[len(ref_lines):]:
+        diffs.append("> {0}".format(line))
+    for line in ref_lines[len(out_lines):]:
+        diffs.append("< {0}".format(line))
+
```
berquist, line 1268: This could use a comment that this captures when one of the files is longer than the other.

src/sst/core/testingframework/sst_unittest_support.py
```diff
@@ -1181,6 +1180,99 @@ def testing_compare_filtered_diff(
...
+        out_key, out_vals = _extract_float_key_and_values(out_line)
+        ref_key, ref_vals = _extract_float_key_and_values(ref_line)
+
+        if out_key != ref_key or len(out_vals) != len(ref_vals):
+            diffs.append("< {0}> {1}".format(ref_line, out_line))
+            continue
+
+        if not out_vals:
+            diffs.append("< {0}> {1}".format(ref_line, out_line))
+            continue
```
berquist, line 1255: I'm not sure what this block is, or how it wouldn't be captured already by the above block.

berquist (changes requested):
I can confirm that this passes Rocky 10 on AnyBranch and building a UBI 10 container image.

Since this is allowing previously skipped tests to pass, and we don't know exactly what failing output will look like yet, this could use some tests.  I started some previously in `tests/testsuite_default_testingframework.py`.  As long as the generated tests cover each of the edge cases of the function, it will be a lot faster to have Claude generate the tests and go over them carefully than do anything by hand.

